Issue #45, opened 2014-12-02
Convert to the layout API


waddlesplash:
This will fix a lot of the sizing issues.

dsizzle:
I think the main two items left are:

- Layers window
- Colors window

dsizzle:
I think this ticket can be closed, no?  The only item I'm aware of that is not-layout-savvy is the Save file panel, but that's not a LayoutAPI thing (rather unfortunately).

humdingerb:
Yay! \o/

dsizzle:
I appreciate the cheer emoji \o/ but can we also close this issue?  I like seeing the ticket numbers go down. :laughing: 

\o/ \o/ \o/ 

¯\_(ツ)_/¯

humdingerb:
It's of course all Web+ fault. :) Merged PRs appear to be still open but arent't, and closing an issue either doesn't work at all, or you're not sure if it's actually closed. I apparently forgot to double-check that with the phone. Doing that now...

PR #52, opened 2017-02-18
convert Empty Paint Window to the layout API

owenca:
font 8
font 24

### DIFF
--- a/artpaint/paintwindow/PaintWindow.cpp
+++ b/artpaint/paintwindow/PaintWindow.cpp
@@ -48,6 +48,7 @@
 #include <Button.h>
 #include <Clipboard.h>
 #include <Entry.h>
+#include <LayoutBuilder.h>
 #include <MenuBar.h>
 #include <NodeInfo.h>
 #include <Path.h>
@@ -130,57 +131,31 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
 		fAdditionalHeight += fMenubar->Bounds().Height() + 1.0;
 	}
 
-	BRect bounds = Bounds().InsetByCopy(-1.0, -1.0);
-	bounds.top = fMenubar ? fMenubar->Bounds().Height() : bounds.top;
-
-	fHorizontalScrollbar = new BScrollBar(BRect(bounds.left,
-		bounds.bottom - B_H_SCROLL_BAR_HEIGHT,
-		bounds.right - B_V_SCROLL_BAR_WIDTH, bounds.bottom),
-		"horizontal", NULL, 0, 0, B_HORIZONTAL);
-	AddChild(fHorizontalScrollbar);
+	fHorizontalScrollbar = new BScrollBar("horizontal", NULL, 0, 0, B_HORIZONTAL);
 	fHorizontalScrollbar->SetRange(0.0, 0.0);
 	fHorizontalScrollbar->SetSteps(8.0, 32.0);
 
-	fVerticalScrollbar = new BScrollBar(BRect(bounds.right - B_V_SCROLL_BAR_WIDTH,
-		bounds.top, bounds.right, bounds.bottom - B_H_SCROLL_BAR_HEIGHT),
-		"vertical", NULL, 0, 0, B_VERTICAL);
-	AddChild(fVerticalScrollbar);
+	fVerticalScrollbar = new BScrollBar("vertical", NULL, 0, 0, B_VERTICAL);
 	fVerticalScrollbar->SetRange(0.0, 0.0);
 	fVerticalScrollbar->SetSteps(8.0, 32.0);
 
 	fAdditionalWidth += B_V_SCROLL_BAR_WIDTH - 1.0;
 	fAdditionalHeight += B_H_SCROLL_BAR_HEIGHT - 1.0;
 
-	bounds.top += 1.0;
-	bounds.right -= B_V_SCROLL_BAR_WIDTH + 1.0;
-	bounds.bottom -= B_H_SCROLL_BAR_HEIGHT + 1.0;
-
 	// The status-view is not optional. It contains sometimes also buttons that
 	// can cause some actions to be taken.
 	if ((views & HS_STATUS_VIEW) != 0) {
 		// Create the status-view and make it display nothing
 		// Note: this sucks, since the status view resizes himself...
-		fStatusView = new StatusView(BRect(0.0, bounds.bottom, bounds.right,
-			bounds.bottom));
+		fStatusView = new StatusView();
 		fStatusView->DisplayNothing();
-
-		// place the statusbar along bottom of window on top of scrollbar
-		// and also record size that it occupies
-
-		// here add the statusview to window's hierarchy
-		AddChild(fStatusView);
-
-		// update the bottom value to be 1 pixel above status view
-		bounds.bottom -= fStatusView->Bounds().Height() + 1.0;
 
 		// update the fAdditionalHeight variable
 		fAdditionalHeight += fStatusView->Bounds().Height() + 1.0;
 	}
 
 	// make the background view (the backround for image)
-	fBackground = new BackgroundView(BRect(0.0, bounds.top, bounds.right,
-		bounds.bottom));
-	AddChild(fBackground);
+	fBackground = new BackgroundView(BRect(0, 0, 0, 0));
 
 	if ((views & HS_SIZING_VIEW) != 0x0000)  {
 		// we need to show the creata canvas area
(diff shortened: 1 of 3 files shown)


owenca:
This is the first step for addressing Issue #45.

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -130,57 +131,31 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
...
 
-	BRect bounds = Bounds().InsetByCopy(-1.0, -1.0);
-	bounds.top = fMenubar ? fMenubar->Bounds().Height() : bounds.top;
-
-	fHorizontalScrollbar = new BScrollBar(BRect(bounds.left,
-		bounds.bottom - B_H_SCROLL_BAR_HEIGHT,
-		bounds.right - B_V_SCROLL_BAR_WIDTH, bounds.bottom),
-		"horizontal", NULL, 0, 0, B_HORIZONTAL);
-	AddChild(fHorizontalScrollbar);
+	fHorizontalScrollbar = new BScrollBar(BRect(0, 0, 0, 0), "horizontal", NULL, 0, 0, B_HORIZONTAL);
```
humdingerb: Something to look out for: if there's a BRect in some control's c'tor, consult the HaikuBook if there's a layout aware version. As is the case for BScrollBar.

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -130,57 +131,31 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
...
 	fHorizontalScrollbar->SetRange(0.0, 0.0);
 	fHorizontalScrollbar->SetSteps(8.0, 32.0);
 
-	fVerticalScrollbar = new BScrollBar(BRect(bounds.right - B_V_SCROLL_BAR_WIDTH,
-		bounds.top, bounds.right, bounds.bottom - B_H_SCROLL_BAR_HEIGHT),
-		"vertical", NULL, 0, 0, B_VERTICAL);
-	AddChild(fVerticalScrollbar);
+	fVerticalScrollbar = new BScrollBar(BRect(0, 0, 0, 0), "vertical", NULL, 0, 0, B_VERTICAL);
```
humdingerb: Same here.

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -130,57 +131,31 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
...
-	bounds.bottom -= B_H_SCROLL_BAR_HEIGHT + 1.0;
-
 	// The status-view is not optional. It contains sometimes also buttons that
 	// can cause some actions to be taken.
 	if ((views & HS_STATUS_VIEW) != 0) {
 		// Create the status-view and make it display nothing
 		// Note: this sucks, since the status view resizes himself...
-		fStatusView = new StatusView(BRect(0.0, bounds.bottom, bounds.right,
-			bounds.bottom));
+		fStatusView = new StatusView(BRect(0, 0, 0, 0));
```
humdingerb: BView, too, has a layout aware c'tor. Not sure how deep you wanna dig for your layouting conversion...

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -130,57 +131,31 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
...
-
 		// update the fAdditionalHeight variable
 		fAdditionalHeight += fStatusView->Bounds().Height() + 1.0;
 	}
 
 	// make the background view (the backround for image)
-	fBackground = new BackgroundView(BRect(0.0, bounds.top, bounds.right,
-		bounds.bottom));
-	AddChild(fBackground);
+	fBackground = new BackgroundView(BRect(0, 0, 0, 0));
```
humdingerb: Same here.

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -988,7 +971,7 @@ PaintWindow::SetHelpString(const char *string,int32 type)
 bool
 PaintWindow::openMenuBar()
 {
-	fMenubar = new BMenuBar(BRect(), "menu_bar");
+	fMenubar = new BMenuBar(BRect(), "menu_bar", false);
```
humdingerb: Also available with layout aware c'tor. BMenuBar("menu_bar") should do it.

artpaint/paintwindow/PaintWindow.cpp
owenca: Tried but couldn't get it to work. Do I have to set the last BLayout* argument (defaults to NULL) in the layout-aware constructor?

artpaint/paintwindow/PaintWindow.cpp
owenca: Ditto.

artpaint/paintwindow/PaintWindow.cpp
humdingerb: I don't think so. Something like BView("status view", B_WILL_DRAW | B_SUPPORTS_LAYOUT) should work. Of course that'll entail converting BackgroundView.cpp and StatusView.cpp to use layouting, too. That's what I meant with how deep you want to go...

waddlesplash: Yep.

owenca: backgroundview

The above is what I got after switching to the layout c'tor without converting BackgroundView to the layout API. It would probably require converting the container box and the image view to the layout API. Too deep for me to dig now.

By the way, B_SUPPORTS_LAYOUT is automatically set in layout c'tors.

owenca: statusview

The above is for StatusView. The black foreground color disappears when the square box moves to the right as the result of increasing the window size. I'd have to dig even deeper.

humdingerb: Exactly. I don't think it's worth it, as it seems to work OK as it is. I'd say we merge.

artpaint/paintwindow/PaintWindow.cpp
owenca: I got StatusView to work but had no luck with BackgroundView.

humdingerb: OK. Let's keep tis PR open for a day should @waddlesplash wishes to comment. Otherwise, @waddlesplash, feel free to merge.

owenca:
Always create the menu bar and the status view (or else the layout builder might crash) but hide them before the layout builder if they were not supposed to be created in the first place.

diversys:
How about aligning text controls (and button)? Also what about that dropdown menu? Maybe there is a native control for that...

owenca:
The container box is a child of the background view and will be dealt with in the next step. I wish I could do something like this:

`if (views & HS_MENU_BAR) .add(fMenubar)`

instead of always creating the menu bar.

owenca:
I found a way to conditionally add fMenubar.

owenca:
Now fStatusView is added to the group layout only if the view has been created.

humdingerb:
Looks complicated... That doesn't mean it's wrong, of course, but my C++ amateur brain creases... :) 
Do give a shout when you think you're ready to merge. And remove the commented out lines first.

owenca:
It's ready unless there is a simpler way to do it with the same efficiency. (The "always create the view first and then hide it if necessary" is wasteful, I think.) Can [PulkoMandy](https://github.com/pulkomandy) or [waddlesplash](https://github.com/waddlesplash) please chime in?

I left the commented-out `.Add()` lines there on purpose. They tell us where the views are supposed to be in the hierarchy.

artpaint/paintwindow/StatusView.cpp
```diff
@@ -410,7 +409,7 @@ SelectedColorsView::Draw(BRect area)
 	SetHighColor(255, 255, 255, 255);
 	StrokeRect(foreground_rect);
 	foreground_rect.InsetBy(1, 1);
-	foreground_rect = foreground_rect & area;
+//	foreground_rect = foreground_rect & area;
```
humdingerb: The comments in the layout builder are alright then what about this one?

owenca: I commented it to have a quick workaround for this:

https://cloud.githubusercontent.com/assets/6434360/23097501/3c7c63ca-f5ea-11e6-8a3a-6bd7c6d28287.png

I guess it was there for a more efficient filling of the square for the foreground color (black). We might be able to restore the line after the status view is converted to the layout API.

artpaint/paintwindow/PaintWindow.cpp
```diff
@@ -316,15 +291,27 @@ PaintWindow::PaintWindow(BRect frame, const char* name, uint32 views,
...
+//				.Add(fStatusView, 0)
+				.End()
+			.Add(fVerticalScrollbar)
+			.End()
+		.Add(fHorizontalScrollbar);
+
+	if (fMenubar) layout->AddView(0, fMenubar);
+
+	if (fStatusView)
+		((BGroupLayout*) ((BGroupLayout*) layout->ItemAt(fMenubar != NULL))->ItemAt(0))->AddView(fStatusView, 0);
```
pulkomandy: This looks a bit ugly. But it can be improved.

- Handle the status view before the menu bar. It will save you the "fMenuBar != NULL" trick.
- It may be easier to do it in a different way:

```
BGroupLayout * layout = new BGroupLayout(B_VERTICAL, 0);
layout->Add(fBackground);
if (fStatusView) layout->Add(fStatusView, 0);
```

And later on in your builder:

```
.Add(layout) // replaces lines 297-300
```

owenca: Thanks! I should have thought of the first bullet.

I prefer your 2nd bullet to my way and tried the follwing, which replaces lines 294-308:
`
BGroupLayout* inner = new BGroupLayout(B_VERTICAL, 0);
inner->AddView(fBackground);
if (fStatusView) inner->AddView(fStatusView, 0);

BGroupLayout* outer = BLayoutBuilder::Group<>(this, B_VERTICAL, 0)
//	.Add(fMenubar)
	.AddGroup(B_HORIZONTAL, 0)
		.Add(inner)
		.Add(fVerticalScrollbar)
		.End()
	.Add(fHorizontalScrollbar);

if (fMenubar) outer->AddView(0, fMenubar);
`
but I got this:
add

When I changed `.Add(inner)` to `.AddGroup(inner)`, I got this:
addgroup

owenca:
I figured it out. Both ways in #if ... #else ... #endif works, but I like the 2nd one better.

pulkomandy:
Yes, 2nd one is good. You can remove the #if block now :)

owenca:
Done!

PR #53, opened 2017-03-19
Fix for image disappearing while manipulated

PeteCA:
This fixes the problem of the image disappearing when translated or rotated, caused by the removal of cpu_clock_speed from system_info.  (Waddlesplash replaced this with a fixed value, but it was way too low.  A fixed value is not very satisfactory, anyway.)

### DIFF
--- a/artpaint/viewmanipulators/RotationManipulator.cpp
+++ b/artpaint/viewmanipulators/RotationManipulator.cpp
@@ -24,6 +24,7 @@
 #include "Selection.h"
 #include "HSPolygon.h"
 #include "StringServer.h"
+#include "SysInfoBeOS.h"
 
 
 #define PI M_PI
@@ -105,9 +106,10 @@ void RotationManipulator::SetPreviewBitmap(BBitmap *bitmap)
 	}
 
 	if (preview_bitmap != NULL) {
-		system_info info;
-		get_system_info(&info);
-		double speed = info.cpu_count * 2000; // TODO: used to be info.cpu_clock_speed but was removed
+		// Use a custom header to get the legacy system_info with cpu speed
+		BeOS_system_info info;
+		get_BeOS_system_info(&info);
+		double speed = info.cpu_count * info.cpu_clock_speed;
 		speed = speed / 15000;
 
 		BRect bounds = preview_bitmap->Bounds();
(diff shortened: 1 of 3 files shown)


waddlesplash:
Do we really care about building this on BeOS? Can't we just make it Haiku-only?

PeteCA:
On Sun, Mar 19, 2017 at 04:15:51PM -0700, waddlesplash wrote:
> Do we really care about building this on BeOS? Can't we just make it Haiku-only?
> 
Sorry -- some misunderstanding there...  This isn't to make it run
in BeOS (though I suspect it might).  It's to handle the way that
ArtPaint decides the resolution it can use when an image is
being manipulated.  It needs the clock speed for that.  (You 
replaced it with a fixed value, but that was way out of the 
expected range so it just blanked the image completely.  I
inserted a more suitable GHz value, which worked on my machine,
but doesn't allow for any adjustment.)  I experimented with a large
canvas, and saw that the original approach seemed to be
necessary.  As was pointed out in the mailing list, an even more
complete solution may be needed sometime, but I think this is
good for some time.

	-- Pete --

waddlesplash:
Honestly I'd say at this point to use the private function, or set it to a constant. (Whoops, I see now what I did, lol.)

PeteCA:
On Sun, Mar 19, 2017 at 04:59:55PM -0700, waddlesplash wrote:
> Honestly I'd say at this point to use the private function, or set it to a constant. (Whoops, I see now what I did, lol.)
(:-))  Yeah -- I finally figured that out!

Which private function?  That's essentially what I did, just wrapping
it in a convenient header.

waddlesplash:
get_rounded_cpu_speed, of course. It's a C function and despite being in a private header is unlikely to change for some time.

PeteCA:
On Sun, Mar 19, 2017 at 05:10:21PM -0700, waddlesplash wrote:
> get_rounded_cpu_speed, of course. It's a C function and despite being in a private header is unlikely to change for some time.
> 
Hmm.  I've sort of avoided private headers, but I see you've used
them in a few places.  (I'm chasing a Layers Window crash that
seems to have been there for some time.)

waddlesplash:
Since we have the source code and can rebuild whenever we want, there's no real reason to not use the private headers honestly.

PR #54, opened 2017-03-21
ArtPaint: New Screenshot

digib0y:
*Removed old screenshot
*Added new screenshot


diversys:
Would be nicer with window tabs

digib0y:
Okay, will update it soon.

scottmc:
Haiku's screenshot app has a feature that allows you to capture the active window including the tab.  When you have the image you want to grab press the printscreen button and it should pop up with the option to save the image.  In that window it should also have a checkbox to grab the window with the tab.

PR #55, opened 2017-03-26
Restored transparency of saved and loaded image files and other fixes

PeteCA:
I merged my previous branch that fixed dynamic display to master, and corrected RGB32 bitmaps to RGBA32, so that transparency in an image is preserved (for PNG etc) when it is saved.  Also now preserves the transparency of a loaded image.  COrrected a badly-placed semaphore that allowed a race-condition crash when the Layer Window was closed.

### DIFF
--- a/artpaint/layers/Layer.cpp
+++ b/artpaint/layers/Layer.cpp
@@ -50,7 +50,7 @@ Layer::Layer(BRect frame, int32 id, ImageView* imageView, layer_type type,
 
 		fLayerData = new BBitmap(BRect(frame.LeftTop(),
 			BPoint(max_c(frame.right, sourceRect.right),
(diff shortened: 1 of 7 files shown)


scottmc:
@PeteCA, Please close any open issues that this addressed.

PeteCA:
On Mon, Mar 27, 2017 at 01:24:43PM -0700, Scott McCreary wrote:
> Merged #55.
> 
Thanks, Scott!

PeteCA:
Urk.  Thanks for pointing that out.  It would not be nice to leave that feature of preview_bitmap unimplemented, though.  It is pretty much essential feedback when you're doing any transforming of the image; otherwise you have no idea what your mouse-move just did!  Guess I'll have to use that private 'get_rounded_cpu_speed' then.  If that's to be a permanent part of the API, shouldn't it be made non-private quickly?

PR #56, opened 2017-03-27
Updated version to 2.1.1 (with various adjustments)

PeteCA:
Sorry -- forgot to do this before the last push.  The versioning of ArtPaint was a bit of a mess by now.  The resource file said "2.1" but other references were all still to "2.0".  I've updated everything to "2.1.1" to account for my fixes.  In updating the About window, I commented out the references to osdrawer;  these should be updated to correct pointers, but I'm not sure what these are!  I also completely removed the unused VersionConstants.h, which still was at "1.2.1"!


scottmc:
Perhaps change the osdrawer lines to point to this github page.
https://github.com/HaikuArchives/ArtPaint

PeteCA:
To be specific, the entries I removed were:

"For new versions of the program see:
http://dev.osdrawer.net/projects/list_files/artpaint"

"Any comments or questions about the program can be sent to:"
http://dev.osdrawer.net/projects/artpaint/boards"

In the 'era of HaikuDepot', I'm not sure what the equivalents are.
 Maybe the first should just be "HaikuDepot" or perhaps "HaikuPorts"?

I guess we don't have "boards"; the "issues" here may be equivalent, or maybe just the Haiku Community pages?

PeteCA:
If it's OK, my next PR will include whatever is a suitable update for the above, plus the fixed add-ons, and the docs folder as I have it now.

scottmc:
Maybe something like:
"For latest released version check Haiku Depot"

and

"To report any issues, please check the issues page on github, and if not already reported, please open a new one."
 https://github.com/HaikuArchives/ArtPaint/issues

PeteCA:
Sounds good.  I'll go with that...

PeteCA:
On second thoughts, I think I'll leave those bits out for now.  Your suggestions are fine in English, but I didn't point out that my quotes are a combination of localized strings, and the URLs.  Without a multilingual expert or two it'd be difficult to be general.

humdingerb:
ArtPaint needs to be adapted to using catkeys. That said, with currently only German and French besides English, we should be able to update the few things needed.

PeteCA:
Agreed.  Maybe this could be a GSoC suggestion?  I'm going to add some issues -- this can be one of them.

Meanwhile can some one tell me why I can't add another PR?  Maybe one can't have two outstanding on the same branch?  I had two open before, but the first was for a side branch.  If so, I can close this one and add the one in waiting.

humdingerb:
Thanks Pete. Maybe more a GCI task than GSoC, which is already running with the student working on their proposals at the moment.

PeteCA:
Thanks!
 I said: "Meanwhile can some one tell me why I can't add another PR", but I had no trouble just now.  I *suspect* that I was actually no longer logged in when I thought I  was! (:-/)

PR #64, opened 2017-04-01
Fixed all add-ons and restored documentation

PeteCA:
All the add-ons are now working.  The problem was the same as for the dynamic display -- the erroneous 'speed' fix 4 years ago.  All the sources are now in the "Working" folder, except for Color Reducer, which I guess works but has a terrible control panel and is probably rarely needed so it is in the "QUESTIONABLE" folder.  None of the add-ons that were listed as crashing ArtPaint seem to do so now, so they are also in the working set.  The "Working" folder is still a subfolder of AddOns, which is totally redundant, but moving things up a level would have entailed editing 27 Makefiles, so I've left it as-is for now.

I updated Heikki's original manual slightly (removing the ShareWare and Downloads sections, and changing the Tools Window image to match), and added it to the tree.  I think it's good enough for now, but could probably stand some more attention.


waddlesplash:
Thanks!

PeteCA:
And thanks from this end...!  I expect to leave it there for a while.  Perhaps @digib0y would like to look at some of the new issues.

PeteCA:
Scott has suggested that I should tag this as release v2.1.1.  I've done this at PeteCA/master, but I can't seem to issue a PR for it here.  Maybe it just has to be done directly?

waddlesplash:
It does. Done: https://github.com/HaikuArchives/ArtPaint/releases/tag/2.1.1

PeteCA:
Great! Thanks a lot.

PR #66, opened 2017-04-23
Fix to NumberSliderControl to cure brush and text problems

PeteCA:
Version 2.1.2.  Made the TextTool control sliders active again, and fixed a glitch that sent a brush inactive if any of its sliders were changed.  (Both problems due to NumberSliderControl.)

### DIFF
--- a/artpaint/controls/NumberSliderControl.cpp
+++ b/artpaint/controls/NumberSliderControl.cpp
@@ -4,7 +4,7 @@
  *
  * Authors:
  *		Karsten Heimrich <host.haiku@gmx.de>
- *
+ *		(fixes by Pete Goodeve 2017)
  */
 
 #include "NumberSliderControl.h"
@@ -187,11 +187,12 @@ NumberSliderControl::_InitMessage()
 		fMessage = new BMessage;
 
 	if (fMessage) {
-		if (fMessage->HasInt32("value"))
+		if (!fMessage->HasInt32("value"))	// may have been set by creator
 			fMessage->AddInt32("value", 0);
 
-		if (fMessage->HasBool("final"))
-			fMessage->AddBool("final", true);
+//		if (fMessage->HasBool("final"))
+	// Should always have this -- often ignored:
+		fMessage->AddBool("final", true);
 	}
 }
 
--- a/artpaint/controls/NumberSliderControl.cpp
+++ b/artpaint/controls/NumberSliderControl.cpp
@@ -183,6 +183,12 @@ NumberSliderControl::TextViewLayoutItem() const
 void
 NumberSliderControl::_InitMessage()
 {
+	// The message returned by the control will have at least
+	// the two items: 'value' and 'final'.
+	// 'value' is of course the current value of the control;
+	// It may have been given an initial value at construction.
+	// 'final' will be false while the slider is being moved,
+	// and set true at mouse-up.
 	if (!fMessage)
 		fMessage = new BMessage;
 
@@ -190,8 +196,6 @@ NumberSliderControl::_InitMessage()
 		if (!fMessage->HasInt32("value"))	// may have been set by creator
 			fMessage->AddInt32("value", 0);
 
-//		if (fMessage->HasBool("final"))
-	// Should always have this -- often ignored:
 		fMessage->AddBool("final", true);
 	}
 }
--- a/COMPILING
+++ b/COMPILING
@@ -1,4 +1,10 @@
-For easiest way to build, open up Terminal to this directory and run "./build.sh"
-Afterwards ArtPaint and it's add-ons will appear built in the "dist" directory
+For easiest way to build, open up Terminal to this directory and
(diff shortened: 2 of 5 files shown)


humdingerb:
Small change, big effect! :)
One question though: why do you comment out stuff instead of removing it? We can always roll back with git should it turn out your fix wasn't quite right...

PeteCA:
> One question though: why do you comment out stuff instead of removing it? We can always roll back with git should it turn out your fix wasn't quite right...
> 
Sorry... (:-/)  Just sloppiness, really.  I typically comment out when I'm changing something I'm not quite sure of, as it makes recovering easy.  Then I forget to clean up...  [I don't work in GitSpace... I use a separate tree where I can play, then copy the changed essentials to  Git.]

humdingerb:
I, too, usually comment things out I'm not sure of... :)
It's always a good idea to do a "git diff" before "git commit". You'll also catch all that whitespace pollution that I always manage to add.

PeteCA:
"Whitespace Pollution"...?!  Whitespace is the soul of clarity1 (:-))

Actually, looking at those comments, I think most of them I left in deliberately, as a guide to the changes in logic.  There's one place where I only added info to the first instance and just commented out the rest.  Maybe I should just remove those, but if you think they're all superfluous I'll dump them all.

PeteCA:
Just one more thing... [Columbo mode... (:-)]  I'll hold off on my next push until I hear your comments on my comments, but I've made one more important change!

<rant>I'm getting a bit fed up with this guy Julun/Karsten!  He seems to have done all these rewrites without checking their consequences.  The one major bug I was still facing was that the Magic Wand didn't work.  Well it did!  You just had to select "Subtract Area" when you wanted to add a selection!!  The test was reversed!</rant>  Flipped two lines and everything is fine...

PeteCA:
Or not...  I was short on testing myself!  The  selection now works, but using it seems not to!  I'll dig.

PeteCA:
Update:  I found the *actual* bug.  You gotta have a Sync() before you try to manipulate bits directly!  So the Magic Wand works now, but I discovered another glitch in Selection in general, that selections don't get merged properly.  I'm  still struggling with this -- nothing much has changed  in that code since the earliest source record -- so it's not worth committing the Magic Wand until I locate that.  I'll remove the comments if you think I should, but otherwise maybe go ahead with this as is.

artpaint/controls/NumberSliderControl.cpp
```diff
@@ -187,11 +187,12 @@ NumberSliderControl::_InitMessage()
...
 
 	if (fMessage) {
-		if (fMessage->HasInt32("value"))
+		if (!fMessage->HasInt32("value"))	// may have been set by creator
 			fMessage->AddInt32("value", 0);
 
-		if (fMessage->HasBool("final"))
-			fMessage->AddBool("final", true);
+//		if (fMessage->HasBool("final"))
+	// Should always have this -- often ignored:
```
humdingerb: Maybe remove the commented line above and rather explain why "final" should always be true and what's it for.

humdingerb:
I commented in-line...

> "Whitespace Pollution"...?! Whitespace is the soul of clarity1 (:-))
With that I meant stray tabs and spaces at the end or otherwise empty lines. I'm not saying your PR has those, but in general. I know it happens to me a lot...

Also, I'm just commenting from my limited POV. You're a better coder than me and if you disagree with something I wrote, don't hesitate to just go with what you think is right... :)

PeteCA:
> 
> > "Whitespace Pollution"...?! Whitespace is the soul of clarity1 (:-))
> With that I meant stray tabs and spaces at the end or otherwise empty lines. I'm not saying your PR has those, but in general. I know it happens to me a lot...

Ah, OK... the whitespace you *can't* see.   Yes that probably happens to me too.

I'll accept your thoughts on comments.  Give me a little while and I'll push a revision.

PeteCA:
OK -- I've cleaned up the comments -- and in the process erased some large commented-out sections that it seems to do fine without!  Please commit if it looks OK.  Thanks!

humdingerb:
I'm not completely sure what the former code in TextManipulator::Draw() was supposed to do, but the Text Tool still works and I trust you know what you're doing...
Tell me, if you'd want to add an explanation for that in you last commit message.
(Just in case, you can do a "git commit -a --amend", edit your last commit message, and then do a "git push -f" to force it into your last commit.)

PeteCA:
I didn't think there was any need for explanation there.  That code was commented out since the very beginning of the archive.  I assume Heikke probably did it himself, and as you say, things work. so I'm sure it's just detritus.

PeteCA:
Oh...Kay...  I've fixed Selection [by hacking around a Haiku bug (:-/)] but I'd rather leave it for a following commit.

humdingerb:
Sorry, I missed the /* in line 115 and thought that part wasn't commented before... oops.
Merging.

PeteCA:
Thanks!  Selection fixes coming over...

PR #69, opened 2017-05-04
Add-on updates, ColorReducer brought back to 'Working' set, Documenta…

PeteCA:
…tion updated

I tagged this as 'v2.1.2', but I don't think it came through.  Anyway I think it's suitable as the next HaikuPorts release.

A few of the add-ons have been cleaned up a bit, so their control panels are more explanatory, and 'ColorReducer' is back in the set.  Documentation has been updated where needed to match the current state.


addons/AddOns/Working/ColorSeparator/ColorSeparator.cpp
```diff
@@ -224,28 +225,30 @@ ColorSeparatorManipulatorView::ColorSeparatorManipulatorView(ColorSeparatorManip
...
 
 	message = new BMessage(MENU_ENTRY_CHANGED);
 	message->AddInt32("value",SHOW_YELLOW);
-	cmyk_menu->AddItem(new BMenuItem("Y",message));
+	cmyk_menu->AddItem(new BMenuItem("Yellow",message));
 
 	message = new BMessage(MENU_ENTRY_CHANGED);
 	message->AddInt32("value",SHOW_BLACK);
-	cmyk_menu->AddItem(new BMenuItem("K",message));
+	cmyk_menu->AddItem(new BMenuItem("blacK",message));
```
humdingerb: Let's capitalize as the other menu items "Black".

addons/AddOns/Working/ColorSeparator/ColorSeparator.cpp
```diff
@@ -224,28 +225,30 @@ ColorSeparatorManipulatorView::ColorSeparatorManipulatorView(ColorSeparatorManip
 	target = t;
 	manipulator = manip;
 
-	BMenu *cmyk_menu = new BMenu("cmyk_menu");
+	BMenu *cmyk_menu = new BPopUpMenu("cmyk_menu");
```
humdingerb: For the ColorSeparator add-on, radio buttons may be better suited than a popup menu. Also, the menu always shows "cmyk_menu" instead the selected colour.

PeteCA:
Whaaat??!  On my system, the menu behaves exactly as intended.  The selected colour shows in the menu button.  ("cmyk_menu" used to be what was always there, but it never should actually appear now.)  Somehow you must still have an older version.

Yes, radio buttons might be nice, but I don't think the difference is major, and I had quite a lot of changes to make, so I was looking to keep things minimal.

And "blacK" was very deliberate!  According to history, the "K" either stands for "Key" or "blacK", and "Key" doesn't make much sense here.  Capitalizing "blacK" that way keeps things consistent.

At some future date, I think the add-on should separate on RGB as well, but I'd rather not do any more rewrites at this point.  I'm satisfied.

humdingerb:
Indeed, I was wrongly in your "master" branch instead of "work". I still think "blacK" is weird, but OK... :)

I better merge this PR then. Or do you have any regrets?

PeteCA:
On Thu, May 04, 2017 at 10:50:31AM -0700, humdinger wrote:
> Indeed, I was wrongly in your "master" branch instead of "work". I still think "blacK" is weird, but OK... :)
Huh?  How does that happen?  I thought the idea of a PR was to direct you to the right place. (And after all my trouble to get my branching right... (:-))
> 
> I better merge this PR then. Or do you have any regrets?
> 
No regrets at the moment.... Maybe later...(:-))  Thanks.

humdingerb:
> Huh?  How does that happen?  I thought the idea of a PR was to direct you to the right place. (And after all my trouble to get my branching right... (:-))

I had checked out your ArtPaint fork, but didn't change into the "work" branch you used for this PR. So I was in your outdated "master" branch went building.

PR #72, opened 2017-07-03
Reverted to old bitmap import in Layer.cpp

PeteCA:
Every image loaded into ArtPaint lost its rightmost and bottom-most pixels.  Turns out this is a bug in the BBitmap::ImportBits used in Layer.cpp.  Don't want to mess trying to compensate for the bug, so I just restored the (only slightly longer) code used in the original.

### DIFF
--- a/artpaint/layers/Layer.cpp
+++ b/artpaint/layers/Layer.cpp
@@ -50,7 +50,7 @@ Layer::Layer(BRect frame, int32 id, ImageView* imageView, layer_type type,
 
 		fLayerData = new BBitmap(BRect(frame.LeftTop(),
 			BPoint(max_c(frame.right, sourceRect.right),
-				max_c(frame.bottom, sourceRect.bottom))), B_RGBA32);
+				max_c(frame.bottom, sourceRect.bottom))), B_RGB_32_BIT);
 
 		if (!fLayerData->IsValid())
 			throw std::bad_alloc();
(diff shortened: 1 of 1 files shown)


waddlesplash:
Uh. The correct thing to do here is to report the bug?

PeteCA:
Hmm.  I located where Julun made the change (commit 91b34c7e09f378c3c6a728e6dff247e525af535a) -- apparently to fix a crash.  It looks like I _may_ have re-introduced the crash (with mismatched bitmap frames), though I also got a crash in the version before my fix.  I'll investigate.

PeteCA:
On Sun, Jul 02, 2017 at 07:15:57PM -0700, waddlesplash wrote:
> Uh. The correct thing to do here is to report the bug?
> 
Well, yes, I intend to do that.  But meanwhile... (:-/)

PeteCA:
It looks as if the crash I noticed has nothing to do with this patch.  It's hard to reproduce anyway, and with multiple runs I seem to see it at about the same frequency both with and without ImportBits.  (It happens at Quit time only -- apparently in the LayerWindow destructor, on a call to RemoveFloater.)

So as far as I'm concerned, the patch stands.

PeteCA:
Slightly red-faced...  It seems that the "width" and "height" values needed by ImportBits are _not_ the values returned by BRect::Width()/Height(), but pixel counts!  Not too ashamed, though, because I think the parameter names in the header are misleading -- they fooled both me and Julun the same way!  (Adrien pointed out the error in a reply to my bug report.)

So I've restored ImportBits to Layer.cpp, with the appropriate adjustment (as that method will obviously not be modified).  Should be properly mergeable now.

humdingerb:
Thanks for the fix!

PeteCA:
Thanks!

PR #73, opened 2017-07-18
Fixes for out-of-range Bitmap references

PeteCA:
Nobody seems to get Bitmap dimensions right!  I happened to be working with a power-of-2 sized image, and hit _two_ separate exceptions!  GaussianBlur and Sharpness both crashed (they use the same code), and trying to scale up x 2 did the same.  Both seem to be in Heikke's original code (but don't crash BeOS).  They are both only off-by-one accesses, and don't even trigger a Haiku exception every time.  They were a pain to track down, but I think the fixes are correct.

(In ScaleManipulator.cpp, I also changed "..._bpr" variables to "..._pxpr", because they are  not bytes-per-row, but 32-bit pixel counts.)

### DIFF
--- a/addons/UtilityClasses/ImageProcessingLibrary.cpp
+++ b/addons/UtilityClasses/ImageProcessingLibrary.cpp
@@ -4,7 +4,6 @@
  *
  * Authors:
  * 		Heikki Suhonen <heikki.suhonen@gmail.com>
- *	Array overflow fix 2017/7/18  Pete G.
  */
 #include <OS.h>
 #include <stdio.h>
--- a/addons/UtilityClasses/ImageProcessingLibrary.cpp
+++ b/addons/UtilityClasses/ImageProcessingLibrary.cpp
@@ -4,7 +4,7 @@
  *
  * Authors:
  * 		Heikki Suhonen <heikki.suhonen@gmail.com>
- *
+ *	Array overflow fix 2017/7/18  Pete G.
  */
 #include <OS.h>
 #include <stdio.h>
@@ -281,7 +281,7 @@ void ImageProcessingLibrary::filter_1d_and_rotate_clockwise(int32 *s_bits,int32 
 
 		uint32 *target_array_position = target_array;
 		for (int32 x = left;x<=right;x++) {
-			*(d_bits + (d_bpr-y) + x*d_bpr) = *target_array_position++;
+			*(d_bits + (d_bpr-1-y) + x*d_bpr) = *target_array_position++;
 		}
 	}
 
@@ -297,7 +297,6 @@ void ImageProcessingLibrary::filter_1d_and_rotate_counterclockwise(int32 *s_bits
 	uint32 *target_array = new uint32[width];
 
 
-	s_bits++;	// TODO: Why this?
 	for (int32 y=top;y<=bottom;++y) {
 		uint32 *source_array_position = source_array;
 		for (int32 dx = 0;dx<kernel_radius;++dx) {
--- a/artpaint/viewmanipulators/ScaleManipulator.cpp
+++ b/artpaint/viewmanipulators/ScaleManipulator.cpp
@@ -96,9 +96,9 @@ BBitmap* ScaleManipulator::ManipulateBitmap(ManipulatorSettings *set,
 		if (intermediate_bitmap->IsValid() == FALSE)
 			throw std::bad_alloc();
 		uint32 *target_bits = (uint32*)intermediate_bitmap->Bits();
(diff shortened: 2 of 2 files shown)


humdingerb:
FWIW, I tried scaling with the un-fixed version a few times and it didn't crash.
Anyway, if @waddlesplash agrees with your changes, who am I to stand in your way? :)

Just one comment in the review...

addons/UtilityClasses/ImageProcessingLibrary.cpp
```diff
@@ -4,7 +4,7 @@
  *
  * Authors:
  * 		Heikki Suhonen <heikki.suhonen@gmail.com>
- *
+ *	Array overflow fix 2017/7/18  Pete G.
```
humdingerb: You wanna change that to be roughly in the same format as Haikki's entry? One more indent, name, email (if you want) and "Array overflow fix (18-07.2017)" in brackets, maybe.

PeteCA:
On Tue, Jul 18, 2017 at 11:08:48PM -0700, humdinger wrote:
> FWIW, I tried scaling with the un-fixed version a few times and it didn't crash.
> Anyway, if @waddlesplash agrees with your changes, who am I to stand in your way? :)
> 
Yes, it doesn't crash every time, and it only happens when the width is a power of 3  (I guess something to do with the precise way memory is allocated), but it was consistent enough to be really frustrating!

PeteCA:
On Tue, Jul 18, 2017 at 11:10:29PM -0700, humdinger wrote:
> 
> You wanna change that to be roughly in the same format as Haikki's entry? One more indent, name, email (if you want) and "Array overflow fix (18-07.2017)" in brackets, maybe.
> 
Hmm, no -- I think I'd just rather remove it...  Not really necessary anyway.  I think I would always want to standardize on that date format I used (never seen your variant before!); it's internationally understandable and becoming common.
(I didn't indent it because I don't consider myself an "author" there. It was more an annotation.)

humdingerb:
I typo'ed my date, wanted to write 18.07.2017. But you're more correct anyway, according to ISO 8601 it's YYYY-MM-DD, so 2017-07-18.
Removing is of course the cleanest solution... :)

PeteCA:
I'm now glad that nobody has merged this yet... (:-/)  My fix works, but I've just realized a much cleaner one that I'll replace it with when I get a chance (later today when I'm on my machine).  I'll explain then.

PeteCA:
Mfff... My "cleaner solution" is stymied by floating-point precision!

Probably good to explain the original problem.   When an image is to be scaled up, as it steps through the destination row -- say 512 pixels, therefore 0..511 -- it needs to step through the source -- 256 wide, say -- and pick suitable corresponding pixels.  It computes a suitable float step size by dividing the source width by the destination width, giving in this case a  step of 0.5.  However it doesn't just grab a single pixel; it gets the integer floor and ceiling indexes of the current floating offset, and averages the retrieved pixels.  The problem is that if it steps from 0 to 511, the final offset is 255.5, the ceil of which is 256!  Oops!  Not fatal until the very last row, and the memory allocation must be exact, but the above values crash consistently.

I thought I could do better by dividing the right-hand pixel indexes, rather than the widths (255.0/511 in the above), which works fine in theory, but it turns out the final offset ends up as something like 255.0049, which still has a ceil of 256!  Damn...

So I'll leave the current fix alone, as it seems solid.  There are still 'edge effects' if right or bottom pixels differ greatly from the  adjacent ones (because it's averaging them) but I can't see any consistent way to improve that.  The fix doesn't make anything worse.

I've removed my annotation at the head of ImageProcessingLibrary.cpp

PeteCA:
Sorry, folks.  My scaling fix wasn't good enough!  I was too specific to scaling exactly by 2.  It would still crash if I scaled by a bit less than that (so the step was larger than 0.5).  The check is now strict on the final offset value.  I've tried a lot of possibly dangerous values and can't make it crash now.

humdingerb:
Don't sweat it, Pete. :)
You think it's now good to merge, or should be let it simmer for a few more days? How long is the mean incubation period for bugs...? :)

PeteCA:
Infuriating, though... (:-/)  I'm pretty certain it's correct, now; the logic seems right,  and as I say I gave it a bit of a torture test.  Any future crashes are likely to be from something else.  I'd say merge it, and then I can try to resync my own master.  Thanks.

humdingerb:
Thanks for your perseverance. :)

PeteCA:
I get this horrible itch when things don't work.... (:-))

Thanks for merging!

Issue #35, opened 2013-11-22
Smallest window size is too small


corihudson:
When the window is shrunk to the minimum width, the colour palette covers the zoom buttons. 
![squeeze](https://f.cloud.github.com/assets/3019167/1602729/4e06793e-539d-11e3-8183-ecfb6bf30bca.png)

dsizzle:
This is fixed by #104

PR #75, opened 2018-11-12
Fix issue #35: Smallest window size is too small

Vrondir:


scottmc:
Ultimately we'd like to see this dixed my moving over to layout management, but will accept this pull request as in interim solution.
https://github.com/HaikuArchives/ArtPaint/issues/45

PR #79, opened 2019-12-12
Remove country flags in Window->Settings->Language tab.

### LINKED ISSUE #78: Remove flags in settings window
Languages don't have flags. We should remove them like it's planned for Haiku's user docs.
Also, the flags aren't correctly vertically aligned... Let's get rid of them.

emily0618:
Fixes #78.


humdingerb:
Thanks!

PR #82, opened 2022-03-02
ArtPaint: use newer get_system_info/get_cpu_info instead of legacy BeOS call

dsizzle:
Removed reliance on "get_BeOS_system_info" and use newer "get_system_info"
call plus "get_cpu_info" to get the CPU clock speeds.

Also removed "round" function that conflicts with std::round, and
updated call to "ImportBits" to use the public API instead of a
private one.

### DIFF
--- a/addons/AddOns/Working/BlurAddOn/Blur.cpp
+++ b/addons/AddOns/Working/BlurAddOn/Blur.cpp
@@ -136,9 +136,7 @@ int32 BlurManipulator::PreviewBitmap(Selection *sel, bool full_quality,BRegion *
 	if ((settings == previous_settings) == FALSE) {
 		// The blur will be done separably. First in the vertical direction and then in horizontal
 		// direction.
-		system_info info;
-		get_system_info(&info);
-		thread_count = info.cpu_count;
+		thread_count = GetSystemCpuCount();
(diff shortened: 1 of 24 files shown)


korli:
it would make sense to avoid repeating the same calls to calculate speed.

dsizzle:
I moved the get_system_info code into the base class for the Manipulators so that it only lives in one place. 

Ideally I suppose it ought to be a singleton or part of the app itself so that it's only called exactly once, but that's more work.

korli:
Looks good. Please keep in mind that current_frequency is dynamic and can be anything between 0 and the reference clock speed.

dsizzle:
Is there a better way to get clock speed?  That was all I could find

korli:
get_cpu_topology_info() returns a default_frequency. Did you try this?

dsizzle:
Didn't try that. I wasn't aware of that method.

dsizzle:
I updated to use get_cpu_topology_info and fall back to get_cpu_info.  I tried to squash the commits into one but it doesn't seem to have updated this PR.  The squashed commit is 2541ed8dff04b332b48bc82be564671278121d66; not sure why it shows all of the previous commits still. :(

PR #83, opened 2022-03-15
Color selection fix

dsizzle:
Use sizeof(typecode) instead of sizeof(value) because value is a void* and thus could be anything.  

On my haiku this issue prevents color selection from working due to a mismatch in the actual size and expected size.


korli:
Could you point to where this is used? Doesn't look correct. Type code has nothing to do with the value here.

dsizzle:
You're totally right - it wasn't correct.  I was getting lucky that the sizeof(type_code) happened to be the correct size.  I added a parameter to define the size, since I still thing doing sizeof() on a void piece of data is dangerous.  

Only appears to be used in PaintApplication::SetColor()

korli:
Thanks, please check the issues which could get updated with your changes.

PR #84, opened 2022-03-15
ArtPaint: change POWERPC/INTEL defines to use BYTE_ORDER == BIG_ENDIAN

### LINKED ISSUE #81: x86_64 seems to be broken
I've tried the 64bit version, and the image/canvas is just black, I cannot pick any colors in the lower right corner... I can't really do anything.

dsizzle:
Change __INTEL__ and __POWERPC__ preprocessor checks to use BYTE_ORDER
== BIG_ENDIAN. This allows more generic support for architectures, and
also fixes alpha compositing issue on x86_64 (and perhaps selection
issue on x86).

Didn't change a couple of places where inline asm is used because that might still be specific to the PowerPC arch, but I changed the elif to just else so everything else besides POWERPC will just fall through.

Fixes #81

### DIFF
--- a/addons/AddOns/Working/Interference/Interference.h
+++ b/addons/AddOns/Working/Interference/Interference.h
@@ -120,7 +120,7 @@ inline 	asm	float reciprocal_of_square_root(register float number)
 	frsqrte		fp1,number;	// Estimates reciprocal of square-root
 	blr
 }
-#elif __INTEL__
+#else
 float reciprocal_of_square_root(register float number)
 {
 	return 1.0 / sqrt(number);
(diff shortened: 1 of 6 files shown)


korli:
Actually most of POWERPC using reciprocal_of_square_root should stay as is.

dsizzle:
I'm not sure what you mean - I did leave the POWERPC reciprocal_of_square_root part as-is - I didn't change it to the endianness check, but I did make the "#elif __INTEL__" into just "#else" so non-POWERPC archs would still work (as they also use the 1/sqrt version of reciprocal_of_square_root)

PR #87, opened 2022-03-25
ArtPaint, add Dutch translation

Begasus:
Quite a big PR, but seems to be OK here (if I missed anything please tell me).

ArtPaint in Dutch on 32bit(nightly) (build with gcc2):

Art-Paint-32bit


Begasus:
Thanks!

PR #95, opened 2022-03-27
Anti-Dither: Use a BSpinner

### LINKED ISSUE #94: AntiDither addon: Use a BSpinner control
...instead of the BTextControl:

anti-dither

humdingerb:
Set minimal block size to "1". A block size of "0" makes no sense...
Fixes #94

### DIFF
--- a/addons/AddOns/Working/AntiDitherer/AntiDitherer.cpp
+++ b/addons/AddOns/Working/AntiDitherer/AntiDitherer.cpp
@@ -10,17 +10,15 @@
 #include <CheckBox.h>
 #include <ClassInfo.h>
 #include <LayoutBuilder.h>
+#include <Spinner.h>
 #include <StatusBar.h>
 #include <string.h>
 #include <Window.h>
 
 #include "AddOns.h"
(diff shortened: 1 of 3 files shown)


korli:
Looks good to me.

dsizzle:
Looks fine, just seems like an overly large field. 

Also apparently BSpinner is an undocumented control?

humdingerb:
> Looks fine, just seems like an overly large field.

That just appears like that , because that size includes the label and +/- widgets. I deemed it necessary, because without setting the size, it would fill the width of the window. But, after a second thought, and I didn't like setting the max size explicitely in the first place, I now went to put the BSpinner in a horizontal group and add some Glue(). There, better. :)

> Also apparently BSpinner is an undocumented control?

Yes, it's in a private header that I had to add int he addon's makefile.

humdingerb:
@korli , those new github-action seems to fail... builds OK locally.

korli:
it looks like your branch isn't rebased.

korli:
no, now  pkgman says it wants a reboot. this is annoying.

Issue #97, opened 2022-03-27
Accept drag'n'dropped images/project files


humdingerb:
A dropped image/project file on the main window should open it in a new window. Unless it's a 'grey' new window without canvas, then the dropped file should open in this window, not a new one.

Extra points:
- when an image/project file is dropped on the "Layers" window, it's added as a new layer(s).

dsizzle:
The mechanics of accepting a dropped image as a new layer don't seem hard, but this is related to #428 and #380

humdingerb:
More extra points:
Dropping a file on the an open image should add the file as new layer, dropping it on the grey background around the canvas should open as new project.

PR #103, opened 2022-03-31
Open dropped images/project files

humdingerb:
Simply forwards the dropped refs message to the main app.
Addresses #97, though the there mentioned ideas are not completely
trivial to implement. A first step at least.

### DIFF
--- a/artpaint/application/PaintApplication.cpp
+++ b/artpaint/application/PaintApplication.cpp
@@ -210,6 +210,11 @@ PaintApplication::MessageReceived(BMessage* message)
 			}
 		}	break;
 
+		case B_SIMPLE_DATA:
+		case B_REFS_RECEIVED: {
+			RefsReceived(message);
+		}	break;
+
 		default:
 			BApplication::MessageReceived(message);
 			break;
--- a/artpaint/paintwindow/PaintWindow.cpp
+++ b/artpaint/paintwindow/PaintWindow.cpp
@@ -510,6 +510,11 @@ void
 PaintWindow::MessageReceived(BMessage *message)
 {
 	switch ( message->what ) {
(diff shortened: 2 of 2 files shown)


humdingerb:
@dsizzle : you think this little change is OK or good enough for now?

dsizzle:
@humdingerb sorry, I didn't comment because I don't know how drag/drop works in Haiku, but I will test this tonight.

dsizzle:
It seems to work great, except I got a crash when I tried to drag the same image in a second time. 

(where do crash reports save to?  I tried to generate one but no idea where it went)

humdingerb:
You're right! Not just with the same file, but apparently any file...

This is what I get:
[ArtPaint-4253-debug-05-04-2022-05-36-19.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8415146/ArtPaint-4253-debug-05-04-2022-05-36-19.report.txt)

When an app crashes, you choose "Save report" from the alert that pops up. You'll find the debug report on the Desktop. Choosing "Debug" will call the Debugger. If you build with "DEBUGGER = TRUE" in the Makefile, you'll see the sourcecode in the Debugger.

The above crash seems to be fixed by initializing `fImageSizeWindow(NULL) `in PaintWindow's constructor and change in `PaintWindow::FrameMoved()`:

`if ((fImageSizeWindow != NULL) && (!fImageSizeWindow->IsHidden()))`

Not 100% sure that's the solution. You're more familiar with the code, and an actual programmer to boot... :)

dsizzle:
hey, don't give me too much credit - I missed that I hadn't initialized fImageSizeWindow in the constructor!  I will amend that patch for #116 

but - this patch is good to merge!

humdingerb:
Great! :)

PR #104, opened 2022-04-01
Fix bad resizing of PaintWindow and convert to Layout API

### LINKED ISSUE #38: Window should scale to image size if possible
When opening an image, the window should, if possible, scale to the image size.

### LINKED ISSUE #39: 'Crop to selection' feature
I use this feature in GIMP fairly often, and it's not in ArtPaint currently.

### LINKED ISSUE #59: Expanded main Paint Window can't be shrunk again
If the Paint Window is initially small -- say with the default canvas -- and is dragged to fill most of the screen, it can't be dragged back to its original, or other smaller, width.  (Height doesn't seem to have this fault.)  It has something to do with Layout management.

dsizzle:
This is a big change, but the overall feel is better imo. I think there are still some custom layouts but this helps a lot with the Layout API conversion.  Open to suggestions but I think it's much nicer than before.  Could probably nitpick the layouts more.

From the commit message:
Converted PaintWindow and StatusView (and children) to use Layout API.
This fixes the resizing problem where the main window could never be
shrunk after being expanded.

- Removed a ton of custom layouting code in favor of Layout API.
StatusView now uses a nice CardLayout instead of adding and removing
controls, for example, and the widgets placed in the StatusView all
use Layout API.
- Changed BackgroundView to a ScrollView. This removes the extra
layout of placing the scrollbars, but also makes more sense in my
opinion because the scrollbars only affect the ImageView, not the
StatusView - so they shouldn't be placed outside the StatusView
- Help bar is now properly visible in different status layouts
- Rotate and Translate manipulators display properly now
- Removed custom "create canvas" box and place it in the StatusView
- On canvas creation, resize the PaintWindow to fit the new canvas
- Resize To Fit works again

Fixes #59, Fixes #38, Addresses #45


dsizzle:
> You need to rebase against master to include korli's recent changes to solve the conflicts.

I am not seeing any conflicts?  I also see this in github:
conflict

what am I missing?

> I get a crash when trying to open an image:

darn. I spent so much time worrying about the 'create canvas' that I didn't test that. I see the same, sorry!

> Are the scroll bars supposed to float with a 1-pixel gap in the canvas?

Do you mean between the scrollbar and the canvas?  I did add a little space because it looked like it was overlapping.  But I also see there is a gap on the right side of the vertical bar so I will see if I can close that gap.

> The new status bar appears to be quite high with lot's of unused grey space.

It's definitely taller because the old one wasn't actually tall enough - the "help line" gets cut off in manipulator views and the controls overlap:
manip01
manip04

so - one option is to remove the manipulators from that view.  The only manipulators that show up in the status bar are Translate and Rotate, and everything else has a separate dialog. Translate and Rotate could also be pulled into separate dialogs.  I was also thinking of pulling the "Create Canvas" controls into a separate dialog anyway, and adding a "File->New Image" menu option, because that's more 'standard' than current (aka 'master' branch) Create Canvas box (which was all non-standard layout code, which is why I moved it to the status bar)

> The checkmarks and radio buttons in the setup windows are centered again, not left-aligned.
> A few BBoxes in the tool setup windows have multiple "Mode" labels.

These don't make sense.  I didn't touch any of that code and it looks fine on my branch - are you sure your branch is up to date?

humdingerb:
Weird. I see this:
gaps

> Do you mean between the scrollbar and the canvas? I did add a little space because it looked like it was overlapping. But I also see there is a gap on the right side of the vertical bar so I will see if I can close that gap.

Here's a shot:
gaps
The outer scrollbar borders should merge with the window borders. May need negative SetInsets().

> so - one option is to remove the manipulators from that view. 

I think that would be a good idea.
And having the coordinates left-aligned, help text center and magnification right-aligned in a one row status bar. That would probably mean setting a slightly greater minima window l width, but I think most users drive a high enough resolution for that.

>> The checkmarks and radio buttons in the setup windows are centered again, not left-aligned.
> These don't make sense. I didn't touch any of that code and it looks fine on my branch - are you sure your branch is up to date?

Yes. I re-built cleanly after removing the "dist" and all "object" folders...

humdingerb:
I updated to your latest commits and the centered checkboxes etc. are gone and now correct. Dunno if I messed something up, or I built before that was committed. Anyway, one down. :)

humdingerb:
Also: you should take credit for your hard work by adding yourself to the copyright headers of the touched files. :)

BTW, I'm also of the opinion to drop the 90s scrolling credits and have a regular BAboutWindow.

dsizzle:
- fixed the crash when opening images
- removed the manipulators from the status view and shrank it to its original size 
- got rid of those 1 pixel borders

I left the color selection in the statusview for now; may revisit based on your comment to make a single-line status bar with just the help but I want to think about where to put it so I will do that as a separate pull request.  

For now I think this addresses the issues you found but please let me know if you notice any other issues!

dsizzle:
> Also: you should take credit for your hard work by adding yourself to the copyright headers of the touched files. :)

Hmm... maybe I will do this when I do the Locale issue - I'll have to touch almost every source file anyway!

> 
> BTW, I'm also of the opinion to drop the 90s scrolling credits and have a regular BAboutWindow.

Agreed - but in a separate PR

dsizzle:
The CI build failed:

Run vmshell 'cd src; ./build.sh'
  vmshell 'cd src; ./build.sh'
  shell: sh -e {0}
user@127.0.0.1: Permission denied (publickey,keyboard-interactive).
Error: Process completed with exit code 255.

what does this mean @korli ?

humdingerb:
> I left the color selection in the statusview for now; may revisit based on your comment to make a single-line status bar with just the help but I want to think about where to put it so I will do that as a separate pull request.

Right, I forgot about that.. :)
Maybe, since everything else has its own window, the colour palette will get its own, too? Further down the line, there could be some optional 'intelligent' auto-S&T that glues all those windows together. Or at least always remember the user's S&T arrangement.

Crash is fixed, thanks!
If we get those (phantom?) conflicts fixed, I'd say it's time to merge.

dsizzle:
> Right, I forgot about that.. :) Maybe, since everything else has its own window, the colour palette will get its own, too? Further down the line, there could be some optional 'intelligent' auto-S&T that glues all those windows together. Or at least always remember the user's S&T arrangement.

In photoshop, the equivalent control sits at the bottom of the tools pane - but I think it would be odd to glue that control at the bottom of the ArtPaint tool pane.  I'd rather put it in the Color Palette window if anywhere?  

but also yes, some kind of 'docking' would be nice? 

There is also the 'progress bar' that goes in the status view, but I couldn't figure out how to make it trigger.  It seems like if a manipulator takes too long to complete it would show but mine are instant.   

> 
> Crash is fixed, thanks! If we get those (phantom?) conflicts fixed, I'd say it's time to merge.

Do you still see conflicts?  This is my forked branch in github:

github1

and on this PR I see:
github2

...and my CLI git says my branch is up to date.  I'm not sure what else to check/do.

or are you referring to the failed CI build?

humdingerb:
> In photoshop, the equivalent control sits at the bottom of the tools pane - but I think it would be odd to glue that control at the bottom of the ArtPaint tool pane. I'd rather put it in the Color Palette window if anywhere?

The Color Palette window is quite big, what with all them sliders.
Just putting the forground/background panes + current palett in a window would be OK, I think. It'd be nice, if those widgets could be arranged by resizing the window. Just like the Tools window can be resized to have e.g. 2 columns. Then you could arrange the new Color window to have the fore/background panes with the palette below it and tileit under the Tools window. Kinda like Photoshop, just more flexible.

> There is also the 'progress bar' that goes in the status view, but I couldn't figure out how to make it trigger. It seems like if a manipulator takes too long to complete it would show but mine are instant.

Todays machines are just too powerful compared to BeOS days..You'll have to come up with more complex add-ons first. :)

humdingerb:
> Do you still see conflicts?

Turns out, I'm the doofus again... :)
Github says it cannot "rebase" due to conflicts. It let's me "squash & merge" though. You got any last wishes, or shall I merge?

humdingerb:
Oh, BTW, the scrollbars still don't share their borders with the window borders. You can check with Magnify. Wanna try to fix that in this PR, or do a separate PR to finally get this one off the table?

dsizzle:
I'm OK with merging this as-is. The philosophy of the patch was to fix the paint window resizing, and it seemed like I couldn't get it to behave without ripping out all the manual layouts. If that's working, I'm fine if the user forums complain about a flagrant waste of pixels. :D especially if it's only visible in Magnify.

dsizzle:
Augh. I realized that I mis-tagged the issues. This does NOT fix #39, but rather #38. Is that easy to change?

humdingerb:
> Augh. I realized that I mis-tagged the issues. This does NOT fix #39, but rather #38. Is that easy to change?

Sure. Just do a 'git commit -a --amend', change the message and push your change.

dsizzle:
ok, I amended that commit, and since I was there I fixed the scrollbars. :)

humdingerb:
Great, let's merge!

PR #113, opened 2022-04-05
ArtPaint: subtract scrollbars from background width

### LINKED ISSUE #112: Canvas overlays scrollbars on window resize
The canvas seems to lay on top of the scroll bars when resizing the window.
This only seems to happen especially when the window becomes larger than the canvas and the dark grey background becomes visible, and if you're resizing slowly.

Here's a BeScreenCapture of it:

Hmm, the upload doesn't seem to work. Try this: http://0x0.st/ob8q.mp4

dsizzle:
when resizing the window, subtract the width/height of the scrollbars
so that the image doesn't overlap them.

Fixes #112

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -1084,13 +1084,16 @@ void ImageView::adjustPosition()
 {
 	if (LockLooper() == TRUE) {
 		BPoint top_left;
-		BRect pbounds = Parent()->Bounds();
-		top_left.x = (pbounds.Width() - getMagScale()*the_image->Width()) / 2;
-		top_left.y = (pbounds.Height() - getMagScale()*the_image->Height()) / 2;
+		BRect bg_bounds = Parent()->Bounds();
+		BRect vertBounds = ScrollBar(B_VERTICAL)->Bounds();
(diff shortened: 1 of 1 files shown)


humdingerb:
For some reason, I cannot make the window much smaller that the initial canvas size. Not on current master either. I thought I remembered that fixed.
Anyway, the tiny bit I can resize the window smaller doesn't clip the scrollbars anymore. Merge?

dsizzle:
how big is your initial canvas size?  The window does have minimum size limits.

dsizzle:
yeah, min size of the window is set to 500x400 by default.

humdingerb:
The initial canvas size doesn't seem to matter. Happens with 900x600 as well. But only when creating an empty canvas. Dropped files can be resized no problem.

dsizzle:
Interesting. It seems to work fine for me, but I can play with it more. 

I suspect it's the `B_AUTO_UPDATE_SIZE_LIMITS` flag on the PaintWindow causing the issue.  Initially I took that out but ended up putting it back.  Might need to experiment. 

 That shouldn't be affected by this patch though.  This should be ok to merge if you're not seeing the overlap.

humdingerb:
Alright!

PR #116, opened 2022-04-05
ArtPaint: lock fImageSizeWindow before checking if it's hidden

### LINKED ISSUE #115: Crash when loading Project file
Loading an ArtPaint project crashes trying to resize the frame.

dsizzle:
Added lock before checking if fImageSizeWindow IsHidden in both
FrameResized and FrameMoved methods of PaintWindow.  Also added
checks for fImageSizeWindow before all other calls to Lock in
case it's NULL.

Fixes #115


humdingerb:
Looks good!

PR #123, opened 2022-04-08
Layer Window: activate on click and don't eat first click

### LINKED ISSUE #122: Layers window: can't grab scrollbar knob
Create a few layers to, so the scroll bar is needed to scroll up/down. Grab the scrollbar knob and try to move it. Ya can't.

Scrolling via mouse wheel works, albeit very slowly. Needs a speed-up.

dsizzle:
Added message filter so window activates on click. Removed "accept
first click" flag so that the grabbing of scrollbar doesn't need a
second click.

Also added step values to speed up the scrolling for the scroll wheel.

Fixes #122

### DIFF
--- a/artpaint/layers/LayerWindow.cpp
+++ b/artpaint/layers/LayerWindow.cpp
@@ -10,6 +10,7 @@
 #include "FloaterManager.h"
 #include "LayerWindow.h"
 #include "MessageConstants.h"
+#include "MessageFilters.h"
 #include "Layer.h"
 #include "UtilityClasses.h"
 #include "LayerView.h"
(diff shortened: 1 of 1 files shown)


humdingerb:
Works as advertised.
As I can't judge if the solution of a message filter is OK, I'll lave the PR open for a bit for others to comment.

One little issue that didn't change with this PR: if your mouse rests above a pop-up menu, the mouse-wheel doesn't work.
Also not related: for some reason, if you choose to create a new layer with such a pop-up menu, it doesn't work every time... unrelated bug I suppose.

humdingerb:
Thanks!

PR #125, opened 2022-04-10
ArtPaint: Fixed copy/paste with selection

### LINKED ISSUE #121: Copy&Paste doesn't work
* Select a part on an image
* ALT+C to "Copy active layer"
* ALT+V to "Paste as a new layer"

This creates a new, but empty layer.

dsizzle:
When importing data into a bitmap, it MUST be the same size as the
source bitmap otherwise it doesn't work, which is why pasting a
selection into a layer didn't do anything. There's an alternate form
of ImportBits that allows a data size and offset, so I used that and
passed the offset of the selection through.

Fixes #121

### DIFF
--- a/artpaint/layers/Layer.cpp
+++ b/artpaint/layers/Layer.cpp
@@ -27,7 +27,7 @@
 
 
 Layer::Layer(BRect frame, int32 id, ImageView* imageView, layer_type type,
-		BBitmap* bitmap)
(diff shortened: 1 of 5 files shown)


humdingerb:
Works! Will merge later, if there are no objections.

humdingerb:
Thank you!

PR #136, opened 2022-04-22
Localization: Fix broken Finnish catkeys

dsizzle:
I guess I hadn't tested the Finnish catkeys, and it turns out they were broken.  Here's a fixed version.


humdingerb:
Thanks!

PR #145, opened 2022-04-26
Manipulator: kill current manipulator when deleting a layer

### LINKED ISSUE #144: Deleting layers during a manipulator operation causes bad behavior
In one of the UI changes I moved the Translation and Rotation manipulators out of the Status view and into separate windows like the other manipulators.  However, that means that they can get hidden behind the paint window. 

Since the manipulator dialogs aren't modal, you can do other operations while a manipulator is active - for example, you can merge, re-order, and even delete(!) layers while the Translation manipulator is active. Yikes.  Needless to say it doesn't end well. 

The manipulators should either prevent other things from occurring until they are completed (via Ok or Cancel), or they should be effectively Cancelled if another tool is chosen.

dsizzle:
Use "PostponeMessageAndFinishManipulator" function when deleting a
layer - otherwise you can delete a layer that's being manipulated
and naturally bad things happen.

Fixes #144

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -392,19 +392,19 @@ void ImageView::MessageReceived(BMessage *message)
 
 		// This comes from layer's view and tells us to delete the layer.
 		case HS_DELETE_LAYER:
-			{
+			if (!PostponeMessageAndFinishManipulator()) {
 				Layer *removed_layer;
 				int32 removed_layer_id;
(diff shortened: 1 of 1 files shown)


humdingerb:
Why did comment out the line instead of removing it?

dsizzle:
that commented-out line was something I forgot to undo.  I shouldn't push changes right before bed. ;D

humdingerb:
4 bleary eyes see more than 2... :)

PR #146, opened 2022-04-26
Localization of all the add-ons

humdingerb:
The first step.

Has anyone any idea if we can avoid creating separate "locales" folders and catkeys in every add-on-folder?
Or how to actually invoke the 'bindcatalogs" when building the addons? It doesn't work with:
`make -f Makefile_addons bindcatalogs`


humdingerb:
@dsizzle : Hit a little snag: The add-on names in the PaintWindow's menu and the help texts aren't translated. I have to admit, that is above my coding pay grade. :)

I'll make more strings that I find translatable in this PR, but I fear a real programmer is needed to make it actually work everywhere.

humdingerb:
Yeah, it's not a big issue though to have separate catkeys. I just updated the PR with it.
Can we change the build.sh as suggested above?

The other issue, that the menu labels and help texts aren't picked up, is more pressing. I'm out on that one... :)

dsizzle:
The makefile change is fine, but not ideal - it re-links the catkeys every single time, even if they haven't changed. It's the same for the main app, though.

dsizzle:
I posted PR #147 that seems to do the trick - give it a try.  I pulled your branch and it seemed to work for Blur. 

It's not the best solution because it has to instantiate and kill each add-on to get the localized strings, but it works and doesn't have any performance concerns.  This could probably be refactored at some point to be more elegant.

humdingerb:
Seems to work fine!
I updated this PR to move the addons out of the "Working" folder and changed the build.sh, as commented above.

I'll merge this and #147 then...

PR #150, opened 2022-05-01
Interference controls

### LINKED ISSUE #90: Interference addon setup window has no controls
Interference addon setup window has no controls:

interference

dsizzle:
added controls for interference wavelengths and center points. Give it a try!

Fixes #90


humdingerb:
Nice!
You could change the help text to "Left-click to move center point A, right-click for point B." Or something...
Maybe even call those just "Center A" and "Center B" in panel and help text?

Also, I never really used this effect yet, but would it make sense to have a checkbox "Use fore- and background colors" to optionally use those instead of grayscale?

humdingerb:
That is, use the foreground colour for values below 50% and the background for the rest. Or something.

dsizzle:
Yeah, Center A and Center B is better.

Regarding the colors, I agree but I think it should just always be fg and bg colors. You can always set it to black and white if you want grayscale.

humdingerb:
Then maybe a checkbox for "Grayscale"? That way, you wouldn't lose your fb and bg colours if you wanted to create an interference pattern to be used for an alpha-channel, for example. May be interesting if layers get different blending modes one day.
But I leave it up to you, if you you deem it useful...

dsizzle:
well, I was thinking the same thing - the grayscale is only really useful for fun compositing, which ArtPaint doesn't have yet. 

So, in the meantime I did add the Grayscale checkbox.  I also updated the help text and labels.

humdingerb:
Very good!

PR #161, opened 2022-05-06
Localization fixes/improvments

humdingerb:


dsizzle:
This is great. While you're doing this can you change the instances of "air-brush" to "airbrush"?

humdingerb:
>  While you're doing this can you change the instances of "air-brush" to "airbrush"?

Already did that, but apparently missed some... I'll amend the commit in this PR.
While we're at it, there's a bit of a mix of colors <-> palettes.
I'm for sticking with the more generic "colors" as your not manipulating whole palettes (or "color sets"), but individual colours.

Also, I think we should rename the "Show x window" to just say "Colors","Tools", "Layers" etc. since we're already in the _Window_ menu. Whatcha say?

dsizzle:
Sounds good. Honestly I haven't looked through the whole app for consistency but while doing the locale conversion I remember there were some non ideal words and phrases so feel free to fix anything that seems awkward.

humdingerb:
Alright then. Merge.

PR #162, opened 2022-05-06
Don't resize ToolSelectionWindow to default size

### LINKED ISSUE #107: Save "Tools" window size
When resizing the window, the tools icons re-arrange, so you can have more than one column of tools, or arrange them horizontally in rows.
To save that configuration the window needs to remember its size, not just its position.

humdingerb:
The stored window size was ignored and always resized to one
tool icon column.
While true that real horizontal arrangement of the tool icons
still isn't implemented, there's no reason to keep reverting
to one-column tool icons.

Fixes #107


dsizzle:
I won't have time to test it until later, but I would say my only concern with a change like this is understanding if there's a way to get into a bad state that hard to recover from. I can't think of any way that's possible off the top of my head though.

dsizzle:
Tested and it works fine for me! ship it.

humdingerb:
Thanks!

PR #168, opened 2022-05-11
Brush tool enh

### LINKED ISSUE #167: Brush Tool enhancements
- Add Angle slider so user can set it more precisely vs. just dragging
- Add 'Lock Dimensions' checkbox to make square/circular brushes easier
- Make 'Reset Brush' button reset brush to defaults

dsizzle:
Added some enhancments to Brush Tool - angle slider, lock dimensions button, and enabled reset brush button. 

Fixes #167


dsizzle:
oh I also added "SetEnabled" and "Value" to NumberSliderControl.

artpaint/tools/BrushEditor.cpp
```diff
@@ -28,7 +29,7 @@
 
 
 #include <math.h>
-
+#include <stdio.h>
```
humdingerb, line 32: Debug leftover?

humdingerb (changes requested):
There are two things that that I'm not sure are easy to implement:
1. If you "Lock dimensions", the "Height" silder is disabled and keeps showing the old value, even when you change the "Width" slider. Actually, it should mirror the width value. Only if you deactivate the "Lock", it jumps to the current width value.     
If it's not possible to keep updating a disabled slider, maybe the slider shouldn't be disabled at all, but both sliders could set each others value. So the user could grab any of them and both get updated accordingly.

2. Can a stored brush remember its "Lock" setting?. Currently, if you restore some asymmetrical brush (width != height) while having your "Lock" on, the panel 'lies': It shows the asymmetrical brush with a locked "Height". Impossible. :)     
Also, if you disable the "Lock", the brush snaps to be symmetrical, which probably isn't what you want.

artpaint/tools/BrushEditor.cpp
```diff
@@ -94,6 +96,19 @@ BrushEditor::BrushEditor(Brush* brush)
 		new NumberSliderControl(B_TRANSLATE("Height:"), "0",
 		message, 0, 100, false);
 
+	message = new BMessage(kLockDimensionsChanged);
+
+	fLockDimensions =
+		new BCheckBox(B_TRANSLATE("Lock Dimensions"),
```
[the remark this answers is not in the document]
dsizzle: fixed

artpaint/tools/BrushEditor.cpp
dsizzle: fixed!

dsizzle:
1. I decided to keep the slider enabled, so the user can drag either width or height, but when you click lock it always locks height = width.  Not much can be done there.  I think it feels better this way.
2. it didn't make sense to store the lock setting imo because it's more of a UI convenience than a brush property.  I just made sure to uncheck that box when you select a brush from the brush window. idk if you have stronger feelings about it.  this eliminates the weirdness you described, however.

humdingerb (approved):

humdingerb:
Your reasoning for 1) and 2) is sound. Let's merge!

dsizzle:
Thanks! While trying to resolve the alpha drawing, it's challenging to figure out the math for brush fading. I really wanted to be able to reset the angle to 0 and figured might as well add the other stuff.

PR #170, opened 2022-05-18
Brush editor v2

### LINKED ISSUE #169: Brush tool enhancements v2
After consideration:

- convert width and height sliders to 'size' and 'ratio' instead.  This allows a user to simply resize a brush without messing with the proportions
- Fade slider doesn't work well. With a smaller brush, the fade quickly overtakes the brush and makes it transparent.  It should work like a percentage and make the edges soften towards the center. At 100% there should always be at least 1 pixel of the original color, like this:
brush-rect

dsizzle:
Ok - I think this is good to go:

- changed brush to size/ratio instead of width/height
- added FloatSliderControl so that ratio could be a decimal
- changed "Fade" to "Hardness" and reversed so that 100 is the hardest edge and 0 is softest
- fixed elliptical brush so it actually fades to the center no matter the size or orientation by replacing the ellipse brush algorithm completely.

Fixes #169


humdingerb:
Nice!

A few thoughts:
* I wonder, though, if a ratio of 1:20 is really needed. There's hardly much difference between 1:10 and 1:20.
(I still maintain that having the slider knob at a 1:1 ratio in the center of the slider (maybe with an added hash mark and going -10 (left) and +10 (right) would be nicer to avoid having to fiddle with the Angle slider to get it to 90°,if you want your ellipse standing instead of lying on its side.)
* If you're not working with very hi-res images, the Size slider feels a bit fiddly. It's difficult to set small brush sizes, like <20. It'd be nicer if the size scale wouldn't be linear. You know what I mean?
* It'd be cool if a double-click on the Ratio and the Angle sliders would reset just those settings ("1" and "0°" respectively).

Oh, and do update the en.catkeys.

dsizzle:
A few thoughts in return:

- for the ratio I just picked something, and initially I had it stretching the brush instead of squashing. In that case the 1:10 vs 1:20 was very different. However it felt backwards so I switched it to how it works now and i agree visually it's not much different. 1:10 is probably enough.
- with regards to the -10 to +10, then the angle would probably only need to be like -45 to 45 because I think you could cover any possible orientation of the brush.
- but speaking of which, I think the angle slider probably only needs to be -90 to 90 anyway even as is. That would make it easier to get exactly 90. The dragging to rotate is a bit weird in that instance but how useful is that feature anyway? It's neat but does it help the ux? 
- I'm not sure what to do about the size slider. Photoshop has a similar slider from like 0 to 2000. Maybe the snooze function of the slider could make it feel better? I can try making it nonlinear but I don't know if that would make it harder to get larger brushes. 
- regarding the snapping, I personally just type in the exact numbers when I need that? I can look into adding a double click since the slider classes are already subclasses of BSlider

dsizzle:
Ok, so I updated the ratio to max out at 1:10.

I also updated the angle slider to -90 to 90 and fixed the dragging rotation to match.  Now it's super easy to rotate 90 degrees by slamming the slider all the way to the right.

double-clicking is non-trivial - it seems like various parts of the Haiku os implement it by calculating the timing between clicks, but there doesn't appear to be a nice user API for getting a double-click. 

I tried messing with making the slider proportional and it didn't really help much.  I might explore that again but for now I gave up.  I also tried lowering the SnoozeAmount() for the slider and it didn't do anything.  From what I can tell it's all about the granularity of the BSlider and there's not much that can be done aside from making the window wider (and therefore the slider), or decreasing the range of the slider.

I updated the en catkeys.

dsizzle:
You know, a simple fix for size would be to just make it a spin control instead of a slider. It doesn't need to be a slider.  What do you think?

humdingerb:
> I also updated the angle slider to -90 to 90 and fixed the dragging rotation to match.

Good. Only the "-" direction only goes up to -89. should be -90 well, I'd say.

Double-clicking to reset the Angle slider may not be super great anyway. It's neat, but not really discoverable...
However, bringing it easily to back to "0" without too mush fiddling would be nice. How anout a small button "0" to the right of the slider to reset?

> From what I can tell it's all about the granularity of the BSlider and there's not much that can be done aside from making the window wider (and therefore the slider),[...]

Yeah... A good first step would be to have the setup windows horizontally resizable. :)

> You know, a simple fix for size would be to just make it a spin control instead of a slider. It doesn't need to be a slider. What do you think?

First off, I think it would be nice to have every widget - tool or addon - that uses a slider to have a spinner + textcontrol. Most addon's just have a slider with "More" or "Less" labels, but if you need to make the same settings for different projects, you need a value.

The issue with a spinner _instead_ of a slider is that it'll probably take quite a while to go from 1 to 500 by holding the "+" widget...
The potential issue with BSpinner is that it takes quite some screen estate, being two fat +/- buttons side by side. OTOH, they are easy to click without much aiming. :)

dsizzle:
> Only the "-" direction only goes up to -89. should be -90 well, I'd say.

This is very much on purpose, because -90 is identical to 90. Having both is confusing to the code as well as it might end up jumping to +90 when you try to set it to -90, or vice versa. 

>However, bringing it easily to back to "0" without too mush fiddling would be nice. How anout a small button "0" to the right of the slider to reset?

As I mentioned in an earlier comment, you can always just type '0' in the box.  That's why I like having the number next to the slider.  Do we also need a reset-to-0 button? In other apps I type values when I need them to be exact.  

>First off, I think it would be nice to have every widget - tool or addon - that uses a slider to have a spinner + textcontrol. 

Agreed. The fact that I couldn't set the angle precisely is one thing that drove me to make this change!!  I'm not a fan of the textual labels in ArtPaint.  I'd rather have numeric values for everything instead of sliders that go from "Purplish" to "Smelly" or something abstract.

However, a spinner AND a slider seems weird.  A textcontrol+slider seems more typical.

>The issue with a spinner instead of a slider is that it'll probably take quite a while to go from 1 to 500 by holding the "+" widget...

But again, is that something in a typical workflow?  When I used to use Photoshop I mainly used the '[' and ']' keys to change the size of the brush.

dsizzle:
I made another attempt at making the slider proportional, and I think it works a lot better.  It's not perfect but it's better. Give it a try and let me know what you think.

humdingerb:
> I made another attempt at making the slider proportional, and I think it works a lot better. It's not perfect but it's better.

I agree, much better!

> This is very much on purpose, because -90 is identical to 90. Having both is confusing to the code as well as it might end up jumping to +90 when you try to set it to -90, or vice versa.

Technically correct, but the common users might wonder why they are unable to hit -90°, when they move the slider to slowly turn clockwise.

> As I mentioned in an earlier comment, you can always just type '0' in the box. That's why I like having the number next to the slider. Do we also need a reset-to-0 button? In other apps I type values when I need them to be exact.

I generally agree, though esp. with a graphics app, many users are very mouse focused and having to click in a textbox, possibly selecting first all characters and then typing a number, can be viewed as a bit of a hassle. Maybe more so when using a nice graphics tablet (I imagine, I only have a stone old Wacom from around 2000...)

Idea: Can we except the mouse-wheel, see over which slider we're hovering and in/decrease the value by "1"?
That way one could use the slider to get near where we want to go (e.g. "0" angle" and do the fine tuning with th wheel.

Sorry to keep you on this side show when you're keen to solve that alpha thing... :)

humdingerb:
One more thing:

> But again, is that something in a typical workflow? When I used to use Photoshop I mainly used the '[' and ']' keys to change the size of the brush.

Sliders can be moved with the left/right cursors, if the knob has focus. However, currently you cannot give focus by clicking the knob, you need to click into the textbox, for example, and cycle your way to the knob with the TAB key.
Can we have a knob have focus by clicking on it? Would be another way to finetune a value.

dsizzle:
Hmm. Not sure why focus doesn't work. I will try to fix that so at least the arrow keys help.

Not that this needs more "feature creep" 😀 but maybe a possible solution is to add a modifier support to the rotation window to snap at 0, +-45, and +-90. That would give a mouse-friendly way to get exact angles.

With regard to typing, I think most users have one hand on the keyboard anyway for modifier keys, but I guess numbers are farther away from the modifiers. And presumably you can map modifiers to the various graphics tablet buttons which you can't easily do for number keys.

humdingerb:
If it's not too much trouble to implement, the SHIFT-lock for 'straight' angles would be nice, though again pretty well hidden... OTOH, maybe ArtPaint will gain that same SHIFT modifier elsewhere, to lock angles for the Line tool, or keep aspect for rect/ellipse/selections.

dsizzle:
So far I can't figure out the focus.  Even changing the whole slider thing into a BControl it doesn't get the clicks.  It seems to ALWAYS focus on the text control no matter what. :(

The shift- lock was what I was thinking, because I know there are other tickets to do modifiers for drawing operations.

dsizzle:
Adding the shift modifier was super easy - give it a try!  I also reversed the angle of the ellipse because it was backwards from where the arrow pointed. 

Also tweaked the angle slider to be -90 to 90 and it works fine. I was afraid it would get confused.

humdingerb:
Nice! (Though I thought you meant a SHIFT-lock on the angle slider...)

Something's wrong with the ratio though. 1.00 isn't sqare anymore.

No idea about the focus on thumb thing. I thought a MakeFocus(true) in the DrawFocusMark() hook might do it. Either I'm doing something wrong (very possible) or it just doesn't. All sliders in Haiku behave this way apparently...

humdingerb:
Oh, is a tooltip "Hold shift key to snap 45° angles" for the preview easily possible?

dsizzle:
Yes, I fixed the ratio and added a tooltip (and I remembered to update the en catkey too).

dsizzle:
> No idea about the focus on thumb thing. I thought a MakeFocus(true) in the DrawFocusMark() hook might do it. Either I'm doing something wrong (very possible) or it just doesn't. All sliders in Haiku behave this way apparently...

I also tried capturing the MouseDown on the whole NumberSliderControl and doing a MakeFocus(true) on the BSlider, but it didn't work.  

I'm not sure if adding shift- locking for the slider would be doable since the slider is a regular BSlider and we're not handling the mouse events (and therefore the shift modifier wouldn't be attached to anything, if that makes sense).

If there aren't enhancement tickets for BSlider already maybe I should file some - like snapping to hash marks and increasing the resolution of the slider.  It also seems to have a max # of possible "positions" no matter what you set the values to (but you can decrease the # of positions).  it'd also be nice to have a native floating point slider instead of having to make a custom one!

dsizzle:
Actually i wonder I can detect the shift key by itself and adjust the numbers. For now I'm going to sleep tho.

humdingerb:
If I move the mouse very slowly over the preview box while holding SHIFT, I manage to have other that %45° angles. I'm sorry to drag this out further... :)

Also, if you TAB to the thumb, the cursor keys don't de/increment the value by one; it needs a few key presses. I suppose calling SetKeyIncrementValue(int(1)) would do the trick?

dsizzle:
I had purposely left some "space" between the snapped areas, but I removed that so now it should snap only to 45s. let me know if it feels better.

For the key increment, I have to calculate it based on the size of the slider.  It's still not perfect but it's way better.  Sometimes you need 2 keypresses because it happens to be a fractional value.  I tried rounding it up but then it made some numbers impossible to hit. Let me know!

Don't worry about "dragging it out" - I totally appreciate the QA!  And if I'm changing stuff I want it to be an improvement, not just a change!

humdingerb:
I can still set angles other than 45° while holding SHIFT. Though I'm not sure it's bad thing, I just noticed it, because I expected the angles to really snap only to those angles. It's easy enough to move the mouse to the approximate 45° angles by going for the preview corners or median lines. From there it snaps reliably to the 45°s.

> For the key increment, I have to calculate it based on the size of the slider.

Isn't that only necessary for the floating Ratio slider (BTW, totally go for a ticket to add such a beast to the API)? The other sliders wouldn't need the value-by-size-calculation and then SetKeyIncrementValue(int(1)) would work to to in/decrease by one every cursor key press. I imagine... I have no idea of course... :)

I suppose you cannot use a range of 100 to 1000 and set the textcontrol to Slider->Value() / 100 when modified to avoid the floating-slider-size-calc altogether.

dsizzle:
I can't get angles that are not multiples of 45. Maybe it's a combination of my screen resolution, VM, and mouse?  I tried a bunch and it always snaps. :shrug: 

Ok, while typing a long explanation for why I need to calculate the increment, I realized an easy fix.  To make a long story short, behind the scenes I was always using a range of 1-1000 for the BSlider no matter what range it actually way.  Instead I changed it to use a range that is an even multiple of the range provided, if that makes sense.  I.e. for a range of 180 degrees, the inc value for a 1000-tick slider is 5.5555555.  If I use a 900-tick slider instead then the inc value is exactly 5.  Try it, it works perfectly now.

I also noticed that at small sizes the Elliptical brush didn't work.  A brush of size 1 showed nothing at all instead of 1 pixel.  I fixed that also.

HOWEVER

I found that the brush doesn't really update properly if you use the arrow keys. :(  Need to figure out why that's happening.

Also I'm not sure the 'Ratio' is intuitive.  Basically if you have a Size of 1, and a ratio of 10, it stays 1 x 1 pixel.  If you want a 1 x 10 brush, you have to set the size to 10 and the ratio to 10.  That might be hard to figure out.  I feel like size 1 and ratio 10 should give a 1x10 brush, right?   I may have come around to your thinking finally with the -10 to 10 slider. :laughing: Then a -10 could give a 1 x 10.  But do we treat 0 as 1?  You can't have a ratio of 0.

dsizzle:
> I found that the brush doesn't really update properly if you use the arrow keys. :( Need to figure out why that's happening.

Ok, I figured it out. Seems like there's an event on the slider that sometimes doesn't fire, so the editor didn't know the slider wasn't being moved anymore so it didn't update the brush.

All right, at this point I think the bugs are out.  The intuitiveness of the Ratio slider is the only question, and I don't have a good answer yet. Anything I can think of has issues - two sliders might be the most intuitive approach, but it has the oddness of 1:1, 2:2, 3:3, etc. all being the same.  But it's easy to make a 1:10 or 10:1 or 3:2 etc. with two sliders from 1-10.  

my brain hurts.

dsizzle:
Wonderbrush only has perfectly circular brushes from what I can see.

dsizzle:
I changed the ratio slider to a -10 to 10 slider, with 0 being 1:1.  It feels pretty good, but there are 2 caveats:

1. -1, 0, and +1 are all 1:1.  Does that matter? Not sure
2. It's integers so there is a noticeable change between each value.  I tried making it a floating point slider, but the math inbetween -1 and +1 makes it behave weirdly.  

If you think it should be a floating slider for smoothness, I can try to fudge the range so that -1 and +1 aren't 1:1. 

Give it a try and let me know!

humdingerb:
One more thing: Can you add a hash mark at Angle = 0 (and Ratio = 0, if you keep the -10/+10 range). That gives a clue where the "neutral" setting is.

dsizzle:
> That was the nice thing about Ratio going from 1 to 10: you could just grab the thumb and slam it to the left. But I know you'll hate me if I suggested to go back to that... :) 

I feel like the -10 to 10 is more intuitive. I think the 0 to 10 could be confusing. It was easier to code though. :smile: 

> Generally, I wouldn't worry too much about the GUI being mathematically correct (like ratio "0" vs "1"). The common user isn't a a mathematician and will take that in stride.

yeah, the only two reasons I care about mathematical correctness are:
1. does it make sense to a user - like is it an intuitive concept even if I don't care about the math.  2 = 2x as big, for example
2. is it easy to code - a ratio of 1.0 to 10.0 is just a simple multiply or divide without anything else fancy behind the scenes.  
 
A ratio of 0 means you have to handle that differently since you can't divide by 0 and multiply by 0 gives 0.  :laughing:  but yeah, mathematical correctness and ease of coding aren't good arguments against ease-of-use.

dsizzle:
Ok, the Ratio slider is now a float slider

Also I added the hash marks but they are rather subtle.  

anything else?  or can we put this baby to bed?

humdingerb:
> anything else? or can we put this baby to bed?

I'd love to, the little bugger is getting cranky... :)
Just one more thing, if it's too much hassle we'll leave it as is. The Ratio slider... It's a bit weird to have ratios <1 and the aspect of the brush not "reverse" when you get >1, and a ratio =1 not being square. I suppose it's not easily done to have 1 being square, in the middle of the slider and going to -1.10 / 1.10 when moving the slider, skipping the <1 values?

dsizzle:
I don't think I can break the laws of numbers, no. :laughing:  

I think it would be incredibly odd to have 1 at the center and then -10 at the left and +10 at the right. where does -1 go?

The only way I can see it working is if the slider went from 0.1 to 10.0, where 1 is in the middle. But even then, it's odd because the left side is a range of 0.1-1.0, and the right side is a range of 1.0-10. I think I would have to make a special RatioNumberControl that behaved that way. Or a strange "SplitNumberControl" where it works differently on the left half vs. right?  

I am going to push back and say no.  

Maybe the word "Ratio" is implying too much.  Maybe it should be something like "Distortion" where 0 = no distortion.

humdingerb:
> I think it would be incredibly odd to have 1 at the center and then -10 at the left and +10 at the right. where does -1 go?

It's +/-1, could be a "normalized" -1.  :) 
Anyway, let's not dwell on it any longer. Renaming to "Distortion" would be OK, or if you want to be whimsical "Squeeze" :)
Question is, if we move away from "ratio" would it be better to abandon the floating point and just go from -100 to +100?

I swear this is the last you'll hear from me on that. Just shout, "MERGE", and I merge... :-D

dsizzle:
I went with "Rectangularity" because "Squeeze" and "Squash" seemed too whimsical. This is SERIOUS business. :laughing: 

We can always change it later. 

MERGE unless there are bugs. ;)

humdingerb:
I'll merge now, to get this thig off the table. :)

But. "Rectangularity"?!  That's not a word, and worse, it's very long...
How about instead of "Ratio" we say "Aspect". That get's rid of the mathematical part, but it'll still be in the back of your mind when reading "Aspect".

dsizzle:
It *is* a word!
https://www.google.com/search?q=rectangularity

I considered "Aspect" but I thought it might be too similar to Ratio.

PeteCA:
On Wed, May 25, 2022 at 08:16:41AM -0700, Dale Cieslak wrote:
> It *is* a word!
> https://www.google.com/search?q=rectangularity
Doesn't seem to mean exactly what we want, though...
("in the shape of a rectangle...")   A square is a 'rectangle', too,
but an ellipse isn't1
> 
> I considered "Aspect" but I thought it might be too similar to Ratio.
> 
I'm not sure that "Aspect" is clear, either, if you encounter it 'blind'
in the panel, but probably makes sense in context.

I'll side with Humdinger on this one.

humdingerb:
> It _is_ a word!
I just looked at dictionary.com and it wasn't listed there. :)

Anyway, I already wrote why I like "Aspect". Alternatives: "Form", "Shape"

dsizzle:
Aspect is perfectly fine. I wasn't attached to Rectangularity. I just found it as a synonym of squareness 😆

humdingerb:
OK, great, I'll update the label then. :)

BTW, I saw various merge commits and commit messages that were superseded by later commits. What are your thoughts on using  feature branches and "git commit --amend" and maybe sometimes "git -i rebase" to have just one commit (or really contextually separated commits) in a PR?

humdingerb:
You may have to "git pull --rebase origin master" then "git push -f dsizzle master". Then create a new feature branch with "git checkout -b mysuperfeature". Not sure though, I'm quickly in the woods with git myself...

dsizzle:
Yeah that's kind of my workload now, but the github layer adds something special. 

I wonder if creating a second feature branch from a feature branch would roll up all the commits?

humdingerb:
Dunno. Recursive branches all the way down... :)

PR #171, opened 2022-05-18
Float window fix

### LINKED ISSUE #158: Global settings: "Keep in front" options act weird
* Make sure the default "Keep in front" setting is all 'off'.
* Show all the windows from the "Windows" menu.
-> Now the canvas window obscures all those windows
* Turn all "Keep in front" settings "on".
-> Now the canvas window is behind all those windows
* Restart ArtPaint.
-> Now the Brushes, Layers and Toolsetup windows are non-floating again, so the canvas window obscures them. Only the Color palette and Tool selection windows are immune.

If you enter the Global settings again, the checkmarks are all still "on". Click "Apply" and all the windows have the correct floating feel again.

If you close e.g. the Brushes window while floating and re-open it, it's again a normal non-floating window.

Weird.

dsizzle:
The problem was that the window 'feel' (and look) were being set too early on window creation.  I moved them to after the window 'pointer' is initialized.

Fixes #158


humdingerb:
Well done! Merge.

PR #178, opened 2022-06-01
Brush tool preview: Use CTRL instead of SHIFT to snap angles

humdingerb:
Since I just found out that the straight line tool uses CTRL to snap
to 90°/45°/0° angles, and the rectangle and ellipses tools also use
CTRL to lock the aspect to 1:1, i.e. for squares and circles.


dsizzle:
I think this is ok to make ArtPaint consistent for now, but I'd rather change all of the angle-snapping to use SHIFT long-term, for better matching with other apps.

humdingerb:
> I think this is ok to make ArtPaint consistent for now, but I'd rather change all of the angle-snapping to use SHIFT long-term, for better matching with other apps.

Good point. I updated the PR to do just that. All snapping done with SHIFT now, dunno why the commit message hasn't updated on this PR...

dsizzle:
Nice. It looks like maybe the catkeys still say ctrl though?

humdingerb:
Missed that outdated catkeys before. Thanks for the heads-up!

PR #180, opened 2022-06-02
Alpha Drawing Fixes: fix drawing with alpha colors

### LINKED ISSUE #137: Drawing with a colour with alpha < 255 doesn't work as expected
Choose a foreground colour like pure blue 0,0,255 and reduce the alpha value to 50%.
Drawing anything should end up with the background being slightly visible through the half-transparent ink. But it's only a fully opaque lighter blue.

dsizzle:
Big change!
- Lots of code cleanup. consolidated many places where 'selection'
code and non-selection code were duplicated
- Reimplemented ellipse drawing with alpha-friendly algorithm
- Added some new compositing types to PixelOperations.h (src_over and
src_out)
- Made Airbrush, Brush, and Freehand line tools draw into a temp buffer
that gets composited into the current layer after drawing a stroke
- updated BitmapDrawer functions to allow use of different compositing
functions
- added some functions to BitmapUtilities for clearing a bitmap with a
given color, and compositing a bitmap onto another

Fixes #137


dsizzle:
I collapsed it all into one commit - it's pretty huge.  Please test it when you have a chance.

humdingerb:
Seems to work pretty good. Much better than before, definitely!
One thing remaining, I think you mentioned it before in #137, are if soft brush edges or airbrushes are drawn on top of each other. Instead of shine-thru they appear to get brighter, like blending with white instead.

Maybe related to the new temp buffer or compositing:
The freehand tool swallows the first click. Only the second click results in a painted dot. But since we may remove that tool anyway, because the brush tool does it all...

The brush tool, though, also has an issue: it only paints if you move the mouse. simple single clicks are not registered. Same with the hairy brush, though that one probably needs the motion to do its thing(?)

dsizzle:
Argh, I totally knew about the single click issue but I was so happy to fix the ellipses and the transparency tool that I forgot. 

I think the brighter edges might be better than before but definitely not perfect. I may beg to get the rest of this merged and file that as a separate ticket if it gives me too much headache to improve it. 😅

humdingerb:
Alright, then let's merge this thing!

dsizzle:
oh sorry, I miscommunicated.  I meant I would try to fix these issues first, and if it was a pain then I'd beg you to merge. 

In this case I'll just open bugs.  Thanks!!

PR #185, opened 2022-06-03
Alpha drawing: fix halo around soft-edged brushes

### LINKED ISSUE #183: Drawing with soft-edged tools displays as "shiny" instead of transparent
The alpha for soft-edged tools (Brush tool, Airbrush tool) isn't correct on the edges.  This shows as a "halo" instead of transparency:

alpha-edge

dsizzle:
- instead of compositing on a transparent white background,
composite brush strokes on a transparent background that is the
same color as the stroke being drawn. That way there's no unexpected
color; it's just a color composited on itself.

Fixes #183


humdingerb:
That looks much better! \o/

PR #192, opened 2022-06-04
Brush/Freehand tool: fix single-click drawing

### LINKED ISSUE #184: Brush tools don't allow painting a single point
Brush tools as of #180 now require drawing a stroke to make a mark.  Clicking a single point doesn't draw anything.  It used to work.

dsizzle:
Single-click to draw a dot with the brush and freehand tool now
makes sure the compositing of the stroke happens even if no stroke
is drawn.

Fixes #184


humdingerb:
Works, thanks!

PR #193, opened 2022-06-04
Crash fix: fix fImageOpenPanel that shoulda been fProjectOpenPanel

### LINKED ISSUE #191: Crash when trying to open a project after launch
* Launch ArtPaint
* Go directly to "File->Open project..."
* Crash

[ArtPaint-15633-debug-03-06-2022-15-00-53.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8833250/ArtPaint-15633-debug-03-06-2022-15-00-53.report.txt)

No crash if you first load an image.

dsizzle:
Looks like a copy/paste mistake from the localization; trying to set a label on the wrong dialog box that was NULL. 

Fixes #191

### DIFF
--- a/artpaint/application/PaintApplication.cpp
+++ b/artpaint/application/PaintApplication.cpp
@@ -186,7 +186,7 @@ PaintApplication::MessageReceived(BMessage* message)
 			}
 
 			fProjectOpenPanel->SetMessage(&filePanelMessage);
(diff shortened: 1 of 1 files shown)


humdingerb:
Thanks for the quick fix!

dsizzle:
Thanks for the quick merges!

PR #196, opened 2022-06-05
Text Alpha: set alpha blending modes so that text renders correctly

### LINKED ISSUE #88: Alpha on text is incorrect
The alpha masking of text is not correct; see image below:

alpha-text

dsizzle:
Fixes #88


humdingerb:
Nice and easy fix!

PR #201, opened 2022-06-06
Register ArtPaint project MIME type

humdingerb:
including:

* Short/long descriptions
* Sniffer rule
* Icon

Shamelessly lifted from WonderBrush. Thanks!


artpaint/application/PaintApplication.cpp
```diff
@@ -476,6 +479,64 @@ PaintApplication::SetColor(rgb_color color, bool foreground)
...
+
+	// set descriptions
+	if (mime.SetShortDescription("ArtPaint project") < B_OK)
+		fprintf(stderr, "Could not set short description of mime type!\n");
+	if (mime.SetLongDescription("ArtPaint project format containing layers etc.") != B_OK) {
+		fprintf(stderr, "Could not set long description of mime type!\n");
+	}
+
+	// set sniffer rule
+	snifferRule = "0.9 (0xffffffff010101010200000002020202001100110800000000000000)";
```
dsizzle, line 516: is this for sure correct?  Based on my experience with file/libmagic, most signatures weren't quite that long.

PeteCA:
On Mon, Jun 06, 2022 at 09:00:03PM -0700, Dale Cieslak wrote:
> @dsizzle commented on this pull request.
> +	// set sniffer rule
> +	snifferRule = "0.9 (0xffffffff010101010200000002020202001100110800000000000000)";
> 
> is this for sure correct?  Based on my experience with file/libmagic, most signatures weren't quite that long.  
> 
I don''t think the sniffer algorithm has any length limits -- as long as
it's constant and unique.  I'll try and check tomorrow how that sequence
is actually generated.

humdingerb:
@PeteCA : an actual investigation from the source would be much appreciated.

I just saved a dozen different projects, big, small, with one or a number of layers, with/without transparency, and the first 28 bytes didn't change when looking at it in DiskProbe...

PeteCA:
On Mon, Jun 06, 2022 at 09:00:03PM -0700, Dale Cieslak wrote:
 
> +	// set sniffer rule
> +	snifferRule = "0.9 (0xffffffff010101010200000002020202001100110800000000000000)";
> 
> is this for sure correct?  Based on my experience with file/libmagic, most signatures weren't quite that long.  

OK, I've done some digging...  The two most convenient relevant files seem to
be application/ProjectFileFunctions.cpp and application/ProjectFileFunctions.h.
It *doesn't* look as if the above hex is necessarily fixed, but we may be able
to extract fixed bits.

Breaking it down into its components (read them with a monospace font!):

0x ffffffff  01010101  02000000   02020202   00110011   08000000   00000000
   little    file ID   number of  section    dimension  section   data?
   endian              sections   start      section ID length

The above looks all wrong in StyledEdit -- use an editor!

The first surprise is that there are 'sections' -- no sign of that at the use level!
The endianness code is 0 for big endian, but I'm not sure we have to deal with that.
Presumably the 'file ID' is always the second int32, and probably the "Section Start"
and "Dimension Sector ID" will be in the same place.  Whether there will always
be 2 sections, and the dimension sector will always be 8 bytes, I don't know.
There should be a corresponding "section end" code -- 02020203 somewhere.
Note that the endianness, file ID and section start are conveniently endianness
independent; others are not.

PeteCA:
I think this might be an adequate sniffer rule (at least for little endian
machines):

	0.50 ([4] 0x01010101) ([12] 0x0202020200110011)

humdingerb:
Thanks, Pete, I changed the PR to use your suggested sniffer rule. After a few more tests, it seems to work well. We can change it if something like the sections etc.changes in the future.

dsizzle:
are you waiting for more thumbs-ups?  If so I think this looks good. thanks for doing the work!

humdingerb:
Thanks!

PR #204, opened 2022-06-09
Alpha bg two

dsizzle:
@humdingerb @PeteCA (or anyone else who wants to) 

Can you give this PR a try?  I think it seems to solve the alpha checkerboard background but I am wondering if there's something I missed. 

I was working on a more complicated solution a while back but then came back to this and it seems pretty simple - which makes me wonder why I was making it so hard before, or if I'm not covering all bases here - but I can't find anything so far that doesn't work.

One note is that rotating or translating the entire Canvas will still show a white background, but I'm ok with that because it's not really the image bg?   IDK

Also, this is NOT mergeable yet because I was trying to get the Settings window to work to save the alpha bg colors and size of the grid (but for some reason this settings junk is fighting me).

But if you could please try this and let me know if at least the bg works then I would greatly appreciate it.  Thanks!!!

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -166,3 +166,44 @@ BitmapUtilities::ClearBitmap(BBitmap* bitmap, uint32 color, BRect* area)
 		bits += bpr - width;
 	}
 }
+
+
+void
+BitmapUtilities::CheckerBitmap(BBitmap* bitmap,
+	uint32 color1, uint32 color2, BRect* area)
+{
+	uint32 width = bitmap->Bounds().IntegerWidth()+1;
+	uint32 height = bitmap->Bounds().IntegerHeight()+1;
+	uint32 bpr = bitmap->BytesPerRow() / 4;
+
+	int32 start_x = 0;
+	int32 start_y = 0;
+
+	if (area) {
+		width = area->IntegerWidth()+1;
+		height = area->IntegerHeight()+1;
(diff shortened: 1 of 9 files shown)


humdingerb:
Appears to work nicely! Well done!
Of course it'd be nicer if the alpha-checkers bg wouldn't rotate/move/resize with the canvas, but it's no biggie.

I think this PR is missing the Text tool transparency fix.

dsizzle:
Thanks for testing! Yes, this needs to be cleaned up and rebased on master before it’s ready.

> Of course it'd be nicer if the alpha-checkers bg wouldn't rotate/move/resize with the canvas, but it's no biggie.

Yep, agreed, but I couldn’t decide if it’s worth the effort since I’m not totally convinced how often people use those features vs. manipulating layers. :shrug:

dsizzle:
ack, I *knew* it was too easy - the checkered bg saves into the image when you save it :disappointed: 

...so close and yet so far...

dsizzle:
closing; I appreciate the testing and I will clean up and rebase and such and then reopen when it's ready.

...and fix saving. :smile:

dsizzle:
It’s complicated, but that doesn’t really work out. I believe Pete had gone down that road when he was trying to add this.

In any case I have a solution that works where I re tender with the white alpha bg right before saving, and it’s invisible to the user. 

However I think the dumb problem is actually simply that the checker bg has no alpha and so I’m gonna see if simply making it transparent is a simple elegant solution. 🤦 😆

PR #205, opened 2022-06-11
Alpha checker bg: Added checkerboard alpha background as seen in similar tools

### LINKED ISSUE #63: Transparency tool doesn't work
ArtPaint 2.1.1:
I'd like to make a transparent layer, so I choose the transparency tool: it said: "Press the mouse button to adjust layer's transparency"): the only thing that happen, when I attempt to use this tool, is that ArtPaint will use the whole CPU (as I can see from ProcessController in the Deskbar) and no trasparency is applied.

EDIT:
If i open a PNG file with a transparent background, in ArtPaint this transparent background is replaced with a white color. Seems that ArtPaint doesn't support the alpha channel/transparency?

dsizzle:
Cleaned up and finished patch for alpha checker background

- Added checkerboard alpha background to main window and layer thumbnails
- Added mode to render with the white bg (needed to properly save files)
- Added settings tab to Global settings to allow different grid sizes and colors
- Updated SettingServer to check for != B_ERROR instead of B_OK - this is because
if a setting doesn't exist the check will fail, so it's impossible to add new
settings.

Fixes #63

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -166,3 +166,44 @@ BitmapUtilities::ClearBitmap(BBitmap* bitmap, uint32 color, BRect* area)
 		bits += bpr - width;
 	}
 }
+
+
+void
+BitmapUtilities::CheckerBitmap(BBitmap* bitmap,
+	uint32 color1, uint32 color2, uint32 grid_size,
+	BRect* area)
+{
+	uint32 width = bitmap->Bounds().IntegerWidth()+1;
+	uint32 height = bitmap->Bounds().IntegerHeight()+1;
+	uint32 bpr = bitmap->BytesPerRow() / 4;
+
+	int32 start_x = 0;
+	int32 start_y = 0;
+
+	if (area) {
+		width = area->IntegerWidth()+1;
+		height = area->IntegerHeight()+1;
+		start_x = (int32)area->left;
+		start_y = (int32)area->top;
+	}
+
+	uint32 grid_color[2] = { color1, color2 };
+	uint32 cur_color = grid_color[0];
+
+	uint32* bits = (uint32*)bitmap->Bits();
+	bits += start_x + bpr * start_y;
+
+	for (int y = 0;y < height;++y) {
+		for (int x = 0;x < width;++x) {
+			int row = (x+start_x) / grid_size;
+			int col = (y+start_y) / grid_size;
+			if (row % 2 == col % 2)
+				cur_color = grid_color[1];
+			else
+				cur_color = grid_color[0];
+
+			*bits++ = cur_color;
+		}
+		bits += bpr - width;
+	}
+}
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -4,6 +4,7 @@
  *
  * Authors:
  * 		Heikki Suhonen <heikki.suhonen@gmail.com>
+ *		Dale Cieslak <dcieslak@yahoo.com>
  *
  */
 #include "BitmapUtilities.h"
@@ -181,8 +182,13 @@ BitmapUtilities::CheckerBitmap(BBitmap* bitmap,
 	int32 start_y = 0;
 
 	if (area) {
+		*area = *area & bitmap->Bounds();
 		width = area->IntegerWidth()+1;
 		height = area->IntegerHeight()+1;
+		if (width > bitmap->Bounds().IntegerWidth()+1)
+			return;
+		if (height > bitmap->Bounds().IntegerHeight()+1)
+			return;
 		start_x = (int32)area->left;
 		start_y = (int32)area->top;
 	}
@@ -192,18 +198,19 @@ BitmapUtilities::CheckerBitmap(BBitmap* bitmap,
 
 	uint32* bits = (uint32*)bitmap->Bits();
 	bits += start_x + bpr * start_y;
+	uint32 row_size = bpr - width;
 
-	for (int y = 0;y < height;++y) {
-		for (int x = 0;x < width;++x) {
-			int row = (x+start_x) / grid_size;
-			int col = (y+start_y) / grid_size;
-			if (row % 2 == col % 2)
+	for (int y = start_y;y < height+start_y;++y) {
+		int rowMod2 = (y / grid_size) % 2;
+		for (int x = start_x;x < width+start_x;++x) {
+			int col = x / grid_size;
+			if (rowMod2 == col % 2)
 				cur_color = grid_color[1];
 			else
 				cur_color = grid_color[0];
 
 			*bits++ = cur_color;
 		}
-		bits += bpr - width;
+		bits += row_size;
 	}
 }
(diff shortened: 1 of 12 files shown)


dsizzle:
Can we do a dot-release after this?  There are a decent number of fixes big and small, and last time I looked on Polyglot there were some completed translations (Turkish and Catalan at least, i believe).

humdingerb:
Unfortunately, I get a crash on startup:

[ArtPaint-2174-debug-11-06-2022-06-56-33.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8883416/ArtPaint-2174-debug-11-06-2022-06-56-33.report.txt)

That aside, I'm all for a new release!
Since we addded/changed a few strings, we should give the translators a week or so to update their translations. I'll try to raise attention in the forum thread set aside for these things.

Also, if you can fix the crash, don't forget to generate new en.catkeys that I'll upload to Polyglot. :)

dsizzle:
The crash should be fixed, and I updated the catkeys. 

I tried to amend so there wouldn't be multiple commits but it caused conflicts so the commits are a little messy, sorry.  i guess in the future the only way around it (aside from a new clean PR) is pushing with -f to force it?

humdingerb:
> i guess in the future the only way around it (aside from a new clean PR) is pushing with -f to force it?

At least that's what I do, I don't know if that's OK for PRs... (?)
In case of f-pushing, it's a good idea to drop a comment what was changed.

So. I can launch ArtPaint without crashing, but it still does when trying to rotate or move a **layer**.
Strange. I replicated that crash 4 or five times. Then I looked at the settings and tried other colors. Then I tried rotating the **canvas**, which worked. Then I made a selection and tried to rotate that, and it worked. Now I cannot reproduce the original crash, i.e. launch ArtPaint, doodle something, rotate, crash...
In any case, this was the report from when it still crashed, just in case: 
[rotate_ArtPaint-2965-debug-12-06-2022-06-53-29.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8885367/rotate_ArtPaint-2965-debug-12-06-2022-06-53-29.report.txt)

Some remarks on the new settings:
* The settings window is too narrow, the new label "Background" is cut off.
* The tab label "Background" isn't quite true. You set the representation of transparent background, not the actual background that'll get saved with the image. Maybe just "Transparency"?
* The button "Revert" is actually a "Defaults". I think it'd look better left-aligned.
* "Grid Size" should be sentence-cased "Grid size:" with a colon at the end for consistency.
* The tool tips "Drag color from palette to set" could be misread to drag a color from the palette of the main window, which doesn't work. How about "Drop a color from the Colors window" and open the Colors window when you click on a colour well. Clicking the well is what most people will do first anyway, and that way the Colors window pops rigt up.
* As it's not relevant what's colour 1 and 2, how about just "Colors:" and put the colour wells on one row, side-by-side. Saves a bit of space and avoids a bit of empty gray space.
* After saving, there's a white background instead of the checkers where the file dialog was. It redraws when waving another window in front of it.
* "Clear layer" and "Clear canvas" fill the layer/canvas with the background colour. Should probably be transparent/checkers?

Sorry, it got such a long list...

dsizzle:
Excellent, thanks for the feedback!!

in terms of the crash, based on your description I bet if you delete your ArtPaint settings it will come back. Sounds like adjusting the settings fixed the issue.

> After saving, there's a white background instead of the checkers where the file dialog was. It redraws when waving another window in front of it.

that shouldn’t happen, thought I fixed that.

> "Clear layer" and "Clear canvas" fill the layer/canvas with the background colour. Should probably be transparent/checkers?

this is also the behavior if you “clear” a selection or do “cut” on a selection I believe. I think this is a separate issue to solve.  I noticed it and meant to file a ticket.

humdingerb:
> in terms of the crash, based on your description I bet if you delete your ArtPaint settings it will come back.

Indeed, it does.

dsizzle:
Found and fixed another crash if you draw off the edge of the window. Yay!

humdingerb:
OK. I just checked out this PR afresh, and I get a crash after launching and clicking "Create canvas"... :}

[Uploading ArtPaint-5164-debug-13-06-2022-12-27-33.report.txt…]()

humdingerb:
I forgot to add: Can we move the "Micscellaneous" tab to the far right. I think it's nicer to have the specific tabs come first and always keep the catch-all-else Misc at the back.

humdingerb:
A new issue: opening an image only shows it's bottom 20%. The rest is checkers.

dsizzle:
> A new issue: opening an image only shows it's bottom 20%. The rest is checkers

any chance the not-checker part is the part that was behind the open dialog?

I fixed some other redraw issues with stuff behind menus. I wonder if it’s fallout from forcing the white render on save, or forcing the redraw after the settings are changed. Otherwise I’m not sure how the checker bg itself could cause these… this was a simple change without the dang transparency settings!

dsizzle:
> OK. I just checked out this PR afresh, and I get a crash after launching and clicking "Create canvas"... :}

your crash report link is invalid, and I cannot reproduce this - can you repost the crash report?  It works fine for me, even deleting settings.

> A new issue: opening an image only shows it's bottom 20%. The rest is checkers.

I also can't repro this.  All the images I open are working fine.

dsizzle:
I moved the Miscellaneous tab to last - but can't repro the other issues.

humdingerb:
I removed all object folders and checked out PR 205 afresh. I also removed all ArtPaint settings. I still have those issues. Here's once more the "Create canvas" crash:
[ArtPaint-3682-debug-14-06-2022-09-23-37.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8898503/ArtPaint-3682-debug-14-06-2022-09-23-37.report.txt)

Turns out, it's not only the bottom 20% that are visible from an opened image, it's a randomly positioned bar, somethimes it's all checkers, no image. Looks like this:

screenshot1

https://user-images.githubusercontent.com/6107158/173547611-afe7cc12-3c4c-45c9-9749-9d2e2c2e133c.mp4

I see various transparency issues with other tools, like a filled rectangle, or just doodle a few separate brush strokes over each other.

humdingerb:
Oh, and changing zooom level or resizing the window also wreaks havock on the canvas...

dsizzle:
I pulled exactly this PR into a new branch and I still can't reproduce the image loading issue nor the Create Canvas crash. 

I *can* reproduce the weirdness from zooming.  It kind of never recovers and behaves oddly from there on out. :( 

I can force it to redraw and it fixes it, but I'm still confused as to why I need to...I didn't really change anything in the rendering pipeline except the background color is squares instead of solid white... :exploding_head:

humdingerb:
This is all very weird. I zipped up the binary I just compiled from scratch from this PR:
[ArtPaint.zip](https://github.com/HaikuArchives/ArtPaint/files/8907967/ArtPaint.zip)

This was compiled with an Intel i7, Haiku hrev56156, 64bit.

dsizzle:
I found the bug (I think)!  After much testing I discovered that the CheckerBitmap function was the source of trouble - and I discovered that I was not using the proper width and height of the drawn area!  I fixed that and I don't see any weird behavior.  Please give it a try; I'm hoping this was the problem as it seems to all work perfectly now on my end.

humdingerb:
Yay! Indeed fixed!
Let's merge! \o/

PR #206, opened 2022-06-13
Color Reducer: dynamically allocate input_colors to avoid stack overflow

### LINKED ISSUE #91: Reducer addon crashes when changing palette mode
Changing the "Palette mode" from "BeOS palette" to "GLE palette" crashes ArtPaint:

[ArtPaint-5834-debug-27-03-2022-08-53-17.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8357233/ArtPaint-5834-debug-27-03-2022-08-53-17.report.txt)

dsizzle:
Changed stack allocation of input_colors[] to heap allocation to avoid
stack overflow.  This allows GLA palette mode to function instead of
crashing.

Fixes #91


humdingerb:
Very nice!
Now that an alternative to "BeOS palette" exists, I see that GLA delivers much better results. I imagine that only very few have a need forthe BeOS palette...
How about we make GLA the default?
Also, since "palette" is already in the label, let's shorten to "BeOS" and - now that we have some space - rename "GLA palette" to "Generalized Lloyd's Algorithm".

dsizzle:
It does give really nice results, but my concern with making it the default is that it runs as soon as you open the add on, and for large images it can take time - I ran it on a large Hubble telescope image and it took about 5 minutes. 

I think maybe there should be an enhancement ticket with some or all of these:

- don’t run automatically; have a “go” button
- define max refinement iterations. Right now it’s hard coded to 100 but you might not see much difference with, 50 or 25?
- a button that stops the process. Right now I don’t think the Cancel button actually stops it. But you might want to hit stop after the image looks fine and not actually cancel the work that is done.

humdingerb:
> I think maybe there should be an enhancement ticket with some or all of these:

All good points. Shall I open a ticket or will you?
I suppose it's not trivial to multi-thread the algorithm?

What about the menu item renaming?

dsizzle:
Filed #207 

I'm not sure about the menu renaming.  "Generalized Lloyd's Algorithm" is quite long, and so the dialog gets bigger when that option is selected.  It's obviously a super easy change so if you feel strongly about it I'm happy to do it.

humdingerb:
When the "palette" part is removed, it's not _that_ much longer. There's also the old "Floyd Steinberg EDD" in the dithering menu... :)
I'd go with the long version, because it's less obscure than some three-letter acronym and that gives people the opportunity to research what it's about, which is more difficult with "GLA".

BTW, can we use this layout, which looks a bit more orderly:

```
BLayoutBuilder::Group<>(this, B_VERTICAL)
	.AddGrid(B_USE_DEFAULT_SPACING, B_USE_SMALL_SPACING)
		.Add(dither_mode_menu_field->CreateLabelLayoutItem(), 0, 0)
		.Add(dither_mode_menu_field->CreateMenuBarLayoutItem(), 1, 0)
		.Add(palette_size_menu_field->CreateLabelLayoutItem(), 0, 1)
		.Add(palette_size_menu_field->CreateMenuBarLayoutItem(), 1, 1)
		.Add(palette_mode_menu_field->CreateLabelLayoutItem(), 0, 2)
		.Add(palette_mode_menu_field->CreateMenuBarLayoutItem(), 1, 2)
		.End()
	.AddGroup(B_HORIZONTAL)
		.AddGlue()
		.Add(busy)
		.AddGlue()
		.End()
	.SetInsets(B_USE_SMALL_INSETS)
	.End();
```

dsizzle:
Thanks for the layout! 

I changed the strings and I used this grid layout, but I also added a min column width to column 1 so that the UI doesn't expand when the "Generalized Lloyd's Algorithm" option is chosen.  Seems to be some built-in spacing so the field looks wide enough, but this is the minimum width I could make it to avoid a noticeable "jump" when choosing this option.

dsizzle:
Oh, I’ll have to play around with that for other layouts. I thought SetColumnWeight was a proportional value, like between 0 and 1, whereas SetMin/MaxColumnWidth was an explicit pixel width value?

PR #208, opened 2022-06-16
Settings window: Layout changes

humdingerb:
* Rename *ControlViews to reflect their tab view labels and
  move them in the same order as the tab views in te GUI.
* Use BLayoutBuilder templates everywhere.
* Use consistent inset/spacing constants instead of hardcoded values.
* Remove insets around the tab view to avoid wasting space.
* Windows tab: alpha-sort the radio-buttons
* Transparency:
  - Use font size as measure for the background preview size and
    colour swatches.
  - Make the colour swatches a bit smaller and put them side-by-side.
  - Switch rows for "Colors" and "Grid size". Looks better and
    makes space for the...
  - "Defaults" button on the right side

### DIFF
--- a/artpaint/windows/GlobalSetupWindow.cpp
+++ b/artpaint/windows/GlobalSetupWindow.cpp
@@ -852,6 +852,12 @@ GlobalSetupWindow::GlobalSetupWindow(const BPoint& leftTop)
 		settings.FindInt32(skSettingsWindowTab, &activeTab);
 	}
 	fTabView->Select(activeTab);
+
+	float width = 0;
+	for (int i = 0; i < fTabView->CountTabs(); i++)
+		width += fTabView->TabFrame(i).Width();
+
+	fTabView->SetExplicitMinSize(BSize(width + 2 * be_plain_font->Size(), B_SIZE_UNSET));
 }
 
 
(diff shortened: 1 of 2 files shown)


humdingerb:
WRT: "Remove insets around the tab view to avoid wasting space."

This may be controversial, as it takes a bit to get used to. But it's used elsewhere, like Trackers "Get info", Locale, Screensaver, Time prefs, etc.

One thing though, I hope @dsizzle can help with: The "Miscellanous" is cut off a bit, because the window is too narrow. How to tell the bugger to open big enough to show all tab labels?

humdingerb:
Would my 2nd commit in this PR, 9907bc2140e542ae81fa9cdce44d28c8271b4070, be an acceptable solution? Seems to work sorta OK for font sizes from 8 to 30, though for 8pt it's starting to cut of and 30pt does waste a bit of space...

dsizzle:
It looks reasonable, but I’d just use StringWidth rather than be_plain_font->StringWidth because I believe it uses whatever the active font is?

Also it would be better to loop through the tabs and add up the StringWidth of their labels. Then it would still work if we add, remove, or change the wording of a tab. Does that make sense?

humdingerb:
Good call. In my updated PR I use the tab's width directly instead of the labels' StringWidth which gives more usable results. Adding 2x fontsize seems to give enough room for small and large system fonts.

dsizzle:
Ok, finally had a chance to try it and it looks great. ship it!

humdingerb:
Thanks for giving it a spin.

PR #209, opened 2022-06-17
Layer menu: Remove "Insert text..."

humdingerb:
The menu "Insert text..." in the Layer menu does the exact same thing as
the Text tool. If it created a new layer for it first, it may have some
reason for existing in the "Layer" menu, but it doesn't.


dsizzle:
Yes, this was redundant and works fine.

humdingerb:
Thanks

PR #210, opened 2022-06-18
TextManipulator: Fix layout

humdingerb:
Don't AddChild() the NumberSliderControls to the view or they'll
add some empty spacing at the top of the window. Instead, hide
them in the layout.
Use spacing constants instead of hardcoded values.

### DIFF
--- a/artpaint/viewmanipulators/TextManipulator.cpp
+++ b/artpaint/viewmanipulators/TextManipulator.cpp
@@ -600,45 +600,45 @@ TextManipulatorView::TextManipulatorView(TextManipulator* manipulator,
 			}
 		}
 	}
-	fFontMenuField = new BMenuField(B_TRANSLATE("Font"),
+	fFontMenuField = new BMenuField(B_TRANSLATE("Font:"),
 		fFontMenu);
 
 	BMessage *message = new BMessage(FONT_SIZE_CHANGED);
 	fSizeControl =
 		new NumberSliderControl(B_TRANSLATE("Size:"), "0",
 		message, 5, 500, false);
-	AddChild(fSizeControl);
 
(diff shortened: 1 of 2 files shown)


dsizzle:
Thanks, it looks a lot better!

humdingerb:
Thanks. I pushes a small additional change to add a colon after the "Font" label for consistency.
I'll update the catkeys at Polyglot. Let's give the guys another week for more translations and then do the release, OK? Might give us a bit of time for non-string related polishing or bugfixes... :)

dsizzle:
Waiting is fine; I’m going on a bit of a vacation this week so I won’t have my laptop for the next several days anyway. The downside is that I also won’t be able to fix anything either. 😆

humdingerb:
OK, no rush then. Enjoy your free days!

dsizzle:
I fixed the crash in #212 which was the only thing giving me concern about the release.  If we can get that merged I'd feel better, otherwise whenever we feel like doing it is fine by me.  The Hairy Brush code is, well, "hairy" so I don't want to dig into that before release unless you feel strongly about it.

humdingerb:
Nah, all in good time. With the easy updates via SoftwareUpdater, we can make more frequent point-releases  - with just tagging a release and do a minimal adjustment to the haikuports recipe - and can be sure it's delivered to all users without any efforts on their part.
So we'll stick to a new release for next weekend and hope for a few more translations.

dsizzle:
Sounds good!  

Oh, if you have time in the next week should we try 32-bit to make sure nothing broke?

humdingerb:
Good call. I'll have a look.

humdingerb:
@dsizzle: the 32build is indeed broken, see #215.

Issue #212, opened 2022-06-18
Crash while painting with the Brush tool


dsizzle:
Argh, was just trying out the newest updates and hit a crash while painting with the Brush tool. :(

![wahh](https://user-images.githubusercontent.com/10878750/174460224-c62f218c-2e5b-43e7-a358-f441556d1e19.png)

[ArtPaint-138826-debug-13-06-2022-06-13-47.report.txt](https://github.com/HaikuArchives/ArtPaint/files/8934510/ArtPaint-138826-debug-13-06-2022-06-13-47.report.txt)

On the plus side, the alpha is way better. :smile:

Issue #124, opened 2022-04-08
Clang static analysis


Vidrep:
Since ArtPaint is getting some recent love, I thought I'd run the Clang static analysis on the code and provide the report here
[scan-build-2022-04-08-ArtPaint.zip](https://github.com/HaikuArchives/ArtPaint/files/8455836/scan-build-2022-04-08-ArtPaint.zip)
 in case there are a few real bugs amongst the false positives.

dsizzle:
ooh nice, thank you!

Vidrep:
Here's a zip of the Clang Static Analysis for most recent build I did (June 17, 2022)
[scan-build-2022-06-17.zip](https://github.com/HaikuArchives/ArtPaint/files/8929905/scan-build-2022-06-17.zip)

dsizzle:
Thanks @Vidrep - I have a couple of changes queued up from last time but I appreciate a more up to date report!

Vidrep:
After the last batch of commits, the number of bugs found by Clang has dropped from 272 to 206. Here's the latest report.
[scan-build-2022-06-19.zip](https://github.com/HaikuArchives/ArtPaint/files/8936021/scan-build-2022-06-19.zip)

Vidrep:
We've gone from 206 bugs down to 90. Here's the most recent report. 
[scan-build-2022-06-26.zip](https://github.com/HaikuArchives/ArtPaint/files/8987064/scan-build-2022-06-26.zip)

Vidrep:
From 90 to 58 bugs as of most recent report (attached). 
[scan-build-2022-07-02.zip](https://github.com/HaikuArchives/ArtPaint/files/9035304/scan-build-2022-07-02.zip)

Vidrep:
Here's a clang static analysis build for the latest commit to ArtPaint. Reported bugs - 67. Up from 58 last time.
[scan-build-2022-09-25.zip](https://github.com/HaikuArchives/ArtPaint/files/9641423/scan-build-2022-09-25.zip)

Vidrep:
Attached is a clang static analysis as of October 2. Bugs found has dropped from 58  to  49. 
[scan-build-2022-10-02.zip](https://github.com/HaikuArchives/ArtPaint/files/9692861/scan-build-2022-10-02.zip)

Vidrep:
Attached a clang static analysis report Nov 20. Bugs found - 57
[...]

PR #213, opened 2022-06-19
Clang static analysis: Fixed several memory leaks and dead initializations

dsizzle:
This is related to #124 but I haven't gone through the whole report. 

Fixed a number of memory leaks that I added when I fixed the alpha drawing. :/   I am secretly hoping #212 is related to the memory leak as I was painting just fine for a while and then it crashed without warning.


humdingerb:
Only gave it a quick test run. Still works... :)

The "Hairy brush" tool produces different results than some time ago. Seems to be mixing black/dark lines in it.
Not related to this PR, but I imagine with the transparency work?
I haved checked out which commit broke it, as I'm short on time today.

dsizzle:
If I were to bet, I’d put my money on the transparency drawing fixes.

PR #214, opened 2022-06-20
Brush Tool: check for NULL brush_matrix in draw_brush function

### LINKED ISSUE #212: Crash while painting with the Brush tool

dsizzle:
The function call to brush->GetData returns NULL if the dx and dy are out of range.  The code checks if the spans returned are NULL, but never checks if the brush_matrix is NULL.  Also set spans to NULL when returning NULL from Brush::GetData()

Fixes #212

### DIFF
--- a/artpaint/tools/Brush.cpp
+++ b/artpaint/tools/Brush.cpp
@@ -191,6 +191,8 @@ Brush::GetData(span **sp,int32 dx,int32 dy)
 			return diff_11;
 		}
 	}
+
+	*sp = NULL;
 	return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


humdingerb:
\o/

PR #218, opened 2022-06-26
More static analysis fixes to prevent memory leaks and dead assignments

dsizzle:
Some code cleanup, memory leak prevention, and unused assignments to address some items in #124


humdingerb:
Still works, also checked on 32bit Haiku.

Issue #215, opened 2022-06-20
32bit build fails


humdingerb:
Tried with a current 32bit Haiku nightly:

```
/Source/ArtPaint/artpaint/paintwindow/PaintWindow.cpp: In function `static int32 PaintWindow::_AddAddOnsToMenu(void *)':
/Source/ArtPaint/artpaint/paintwindow/PaintWindow.cpp:1434: Internal compiler error.
/Source/ArtPaint/artpaint/paintwindow/PaintWindow.cpp:1434: Please submit info to Oliver Tappe <gcc@hirschkaefer.de>.
/boot/system/develop/etc/makefile-engine:300: recipe for target 'objects_artpaint/PaintWindow.o' failed
make: *** [objects_artpaint/PaintWindow.o] Error 1
```

The offending [line 1434](https://github.com/HaikuArchives/ArtPaint/blob/master/artpaint/paintwindow/PaintWindow.cpp#L1434):

`manip_map.insert(std::pair(BString(manipulator->ReturnName()), *it));`

humdingerb:
I'm stupid. Forgot to do a `setarch x86` first...

dsizzle:
Does this mean this 32 bit build is working? I hope so because that error is with bcc -so it really oughta work.

dsizzle:
I guess we shouldn't have dismissed this...

humdingerb:
Well, for me (and the haikuports buildbots) it was building fine when built with gcc11 on 32bit. I'll merge the fix anyway...

PR #221, opened 2022-06-28
Build fix: set template arguments for std::pair

dsizzle:
Fix build issue for 2.3.0 32-bit build...

...but what's different about the 64-bit build?  Is it more permissive?


dsizzle:
Fixes #215

PR #225, opened 2022-07-21
Layer Window: Remove composite view

### LINKED ISSUE #61: Composite image in Layer View doesn't update
Any drawing or other transformation in a layer appears immediately in the corresponding image in the Layers Window, but the composite image at the top is not updated.  The Layer Window must be closed and reopened to show the changes. (This wasn't the case in BeOS.)

dsizzle:
Removed useless composite view - didn't update layout or anything yet as I'm planning to revamp the layer window.

Fixes #61

### DIFF
--- a/artpaint/layers/LayerWindow.cpp
+++ b/artpaint/layers/LayerWindow.cpp
@@ -39,7 +39,6 @@ LayerWindow* LayerWindow::layer_window = NULL;
 // also the composite picture and target-window's title
 BWindow* LayerWindow::target_window = NULL;
 BList* LayerWindow::target_list = NULL;
-BBitmap* LayerWindow::composite_image = NULL;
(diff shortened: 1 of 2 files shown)


humdingerb:
Welcome back to ArtPaint developing. :)
I guess, since you're planing more changes for layers, it doesn' t make sense at this time to pretty things up, like center-aligning the project title and reduce the top box height.

dsizzle:
Thanks! The layer window is a bit of a beast with a lot of custom layout. I was trying to convert it to the Layout API and move things around, but I made such a mess of it that I decided to scrap all those changes, start over, and do it a little at a time - so this is just the first piece. If it takes a while and gets frustrating I may make it prettier as a temporary move. I saw there are some new translations so I don’t want to delay a release too long.

Issue #109, opened 2022-04-03
Layers are hard to drag


dsizzle:
The UI for Layers doesn't leave a lot of room to grab and drag to re-order. 
- grabbing the thumbnail and dropping on the canvas creates a duplicate layer.  Is it possible to use that same drag functionality within the Layer window for reordering? Is dragging a layer onto the canvas to duplicate the layer an intuitive action? 
- the dropdown for adding, deleting, and merging layers could be moved to a single toolbar/menu so it doesn't take up space in each layer, and therefore would make reordering easier

reference of Photoshop's Layer panel; taken from photoshopcafe.com:
![pshop-layers](https://user-images.githubusercontent.com/10878750/161452829-b219ebfa-1bda-42ff-8de7-ec2cfc67cbef.png)

humdingerb:
> grabbing the thumbnail and dropping on the canvas creates a duplicate layer. Is it possible to use that same drag functionality within the Layer window for reordering?

Sure. See e.g. the "Shape" list in Icon-O-Matic or the list of "Preferred languages" in the Locale prefs. Or the list of favourites of QuickLaunch.

> Is dragging a layer onto the canvas to duplicate the layer an intuitive action?

Not to me...

> the dropdown for adding, deleting, and merging layers could be moved to a single 
 toolbar/menu

+1, leaves space for an opacity setting (maybe like the main window's magnification, i.e. a slider that pops up when clicked on. Also for a blending mode menu ("lighten/darken/alpha/whatever) someday. 
IMO, the combined-layers-preview at the top of the Layers window  could be removed as well: You see all the separate layers below it anyway, plus on the main window's canvas itself.
Maybe I should create tickets for the above...

dsizzle:
> > Is it possible to use that same drag functionality within the Layer window for reordering?

>Sure. See e.g. the "Shape" list in Icon-O-Matic or the list of "Preferred languages" in the Locale prefs. Or the list of favourites of QuickLaunch.

Sorry - dragging already works, but what I meant was if dragging the thumbnail itself could have different behavior depending on whether it was dropped on the layer window or the main window.  

but, given this:

> >  Is dragging a layer onto the canvas to duplicate the layer an intuitive action?

> Not to me...

I am inclined to just eliminate that thumbnail drag so that the entire layer can be grabbed a lot more easily.  

[...]

Issue #20, opened 2013-09-27
Layers window labels and controls are cut off.


soakbot:
**Creator: Richard Sjusen Nyhus**
**URL: http://dev.osdrawer.net/issues/603**

_CAUTION: comments not extracted from issues. See OsDrawer or the WaybackMachine for issue comments._

humdingerb:
Fixed with 506db22d799469b84f5dd7565be9382a03594981

PR #226, opened 2022-07-27
Layers window updates

dsizzle:
I moved the layer operations up to the top and out of each layer, and added a transparency slider so you can adjust layer transparency easily. Also re-enabled(?) the layer name field and moved visibility to the left.  This is more like how Photoshop's layer panel is. 

oh and the changes to the rendering are in preparation for adding the ability to change to different compositing modes for the layers. 

However, the layouting is giving me headaches.  I'm positing this mostly for comments or suggestions - it's not done but wanted to get thoughts. 

Related to #109

### DIFF
--- a/artpaint/layers/LayerView.cpp
+++ b/artpaint/layers/LayerView.cpp
@@ -10,6 +10,7 @@
 
 #include "LayerView.h"
 
+#include "CustomGridLayout.h"
 #include "ImageView.h"
 #include "LayerWindow.h"
 #include "UtilityClasses.h"
@@ -138,28 +139,16 @@ LayerView::MouseDown(BPoint location)
 		if (layer_window)
 			layer_window->SetActiveLayer(the_layer);
 	}
-	// Here initiate a drag-session. If mouse was over the image, whole layer should
-	// be dragged. Otherwise we drag this view in it's parent to reorder the layers.
-	/*if (mini_image_frame.Contains(location)) {
-		// This should start a real drag'n'drop session
-		a_message.what = HS_LAYER_DRAGGED;
-		a_message.AddPointer("layer_bitmap",(void*)the_layer->Bitmap());
-		BBitmap *layer_mini_image = new BBitmap(the_layer->GetMiniatureImage());
-		DragMessage(&a_message,layer_mini_image,B_OP_ALPHA,location-BPoint(5,5));
-	}
-	else {*/
-		// We start reordering the layers.
-		thread_id reorder = spawn_thread(LayerView::reorder_thread,
-			"reorder layers", B_NORMAL_PRIORITY, (void*)this);
-		resume_thread(reorder);
-	//}
+	// We start reordering the layers.
+	thread_id reorder = spawn_thread(LayerView::reorder_thread,
+		"reorder layers", B_NORMAL_PRIORITY, (void*)this);
+	resume_thread(reorder);
 }
 
 
 void
 LayerView::MouseMoved(BPoint where,uint32 transit,const BMessage*)
 {
-
 }
 
 
@@ -234,28 +223,42 @@ LayerView::ReorderViews()
 	BPoint location;
 	BRect frame;
 
-	BWindow *the_window = Window();
+	LayerWindow* the_window = (LayerWindow*)Window();
 
 	int32 positions_moved = 0;
+
 	if (the_window != NULL) {
 		the_window->Lock();
 		BView *parent_view = Parent();
 		BView *exchanged_view;
-		GetMouse(&location,&buttons);
+		parent_view->GetMouse(&location, &buttons);
 		the_window->Unlock();
+
+		CustomGridLayout* layout = (CustomGridLayout*)the_window->GetListView()->GetLayout();
 
 		if (parent_view != NULL) {
 			while (buttons) {
 				the_window->Lock();
-				location = ConvertToParent(location);
-				frame = ConvertToParent(Bounds());
-				frame.OffsetBy(0,-LAYER_VIEW_HEIGHT);
+				frame = this->ConvertToParent(Bounds());
+				frame.OffsetBy(0, -Bounds().Height());
 				if (frame.Contains(location) == TRUE) {
-					exchanged_view = the_window->FindView(the_window->ConvertFromScreen(parent_view->ConvertToScreen(location)));
+					exchanged_view = the_window->FindView(
+							the_window->ConvertFromScreen(
+								parent_view->ConvertToScreen(location)));
 					if (exchanged_view != NULL) {
-						if (exchanged_view->Parent() == parent_view) {
-							exchanged_view->MoveBy(0,LAYER_VIEW_HEIGHT);
-							MoveBy(0,-LAYER_VIEW_HEIGHT);
+		 				int32 exchangedIndex = layout->IndexOfView(exchanged_view);
+		 				// make sure the selected view is one of the layer views,
+		 				// not another control like the layer name
+						while (exchanged_view != NULL && exchangedIndex < 0) {
+							exchanged_view = exchanged_view->Parent();
+							exchangedIndex = layout->IndexOfView(exchanged_view);
+						}
+
+						if (exchanged_view != NULL &&
+							exchanged_view->Parent() == parent_view &&
+							exchangedIndex >= 0 &&
+							exchanged_view != this) {
+							layout->SwapViews(this, exchanged_view);
 							positions_moved++;
 							if (parent_view->Bounds().Contains(Frame().LeftTop()) == FALSE) {
 								parent_view->ScrollBy(0,Frame().top);
@@ -264,13 +267,25 @@ LayerView::ReorderViews()
 					}
 				}
 				else {
-					frame.OffsetBy(0,2*LAYER_VIEW_HEIGHT);
+					frame.OffsetBy(0, 2 * Bounds().Height());
 					if (frame.Contains(location) == TRUE) {
-						exchanged_view = the_window->FindView(the_window->ConvertFromScreen(parent_view->ConvertToScreen(location)));
+						exchanged_view = the_window->FindView(
+							the_window->ConvertFromScreen(
+								parent_view->ConvertToScreen(location)));
 						if (exchanged_view != NULL) {
-							if (exchanged_view->Parent() == parent_view) {
-								exchanged_view->MoveBy(0,-LAYER_VIEW_HEIGHT);
-								MoveBy(0,LAYER_VIEW_HEIGHT);
+							int32 exchangedIndex = layout->IndexOfView(exchanged_view);
+							// make sure the selected view is one of the layer views,
+		 					// not another control like the layer name
+							while (exchanged_view != NULL && exchangedIndex < 0) {
+								exchanged_view = exchanged_view->Parent();
+								exchangedIndex = layout->IndexOfView(exchanged_view);
+							}
+
+							if (exchanged_view != NULL &&
+								exchanged_view->Parent() == parent_view &&
+								exchangedIndex >= 0 &&
+								exchanged_view != this) {
+								layout->SwapViews(this, exchanged_view);
 								positions_moved--;
 								if (parent_view->Bounds().Contains(Frame().LeftBottom()) == FALSE) {
 									parent_view->ScrollBy(0,Frame().bottom-parent_view->Bounds().bottom);
@@ -279,7 +294,8 @@ LayerView::ReorderViews()
 						}
 					}
 				}
-				GetMouse(&location,&buttons);
+
+				parent_view->GetMouse(&location,&buttons);
 				the_window->Unlock();
 
 				snooze(20 * 1000);
@@ -293,8 +309,8 @@ LayerView::ReorderViews()
 		a_message.what = HS_LAYER_POSITION_CHANGED;
 		a_message.AddInt32("layer_id",the_layer->Id());
 		a_message.AddPointer("layer_pointer",(void*)the_layer);
-		a_message.AddInt32("positions_moved",positions_moved);
-		image_window->PostMessage(&a_message,image_view);
+		a_message.AddInt32("positions_moved", positions_moved);
+		image_window->PostMessage(&a_message, image_view);
 	}
 
 	return positions_moved;
@@ -317,7 +333,6 @@ ThumbnailView::ThumbnailView(BBitmap* image)
 
 ThumbnailView::~ThumbnailView()
 {
-
 }
 
 
--- a/artpaint/layers/LayerView.cpp
+++ b/artpaint/layers/LayerView.cpp
@@ -11,13 +11,16 @@
 #include "LayerView.h"
 
 #include "ImageView.h"
+#include "LayerWindow.h"
 #include "UtilityClasses.h"
 
 
 #include <Bitmap.h>
 #include <Catalog.h>
 #include <CheckBox.h>
+#include <GroupLayout.h>
 #include <InterfaceDefs.h>
+#include <LayoutBuilder.h>
 #include <MenuItem.h>
 #include <PopUpMenu.h>
 #include <TextControl.h>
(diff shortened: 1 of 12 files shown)


dsizzle:
ok, I think I got the layouting fixed now.  

But is there something wrong with the CI build?  The message says it can't find 'make' ??

dsizzle:
oh, also Fixes #20

humdingerb:
Sorry by the late feedback.
A few suggestions:
- Add a horizontal separator under the file name.
- The transparency setting takes quite a lot of space. I think we may get away with a label "α" (= alpha symbol), the slider and a numbers box (just wide enough to fit "100"), all in one row.
- The "Action" menu could be moved into a regular menubar at the top. I expect we'll use the space for a "Blending mode" menu at one time.
- The down-arrow of the scrollbar is behind the resize-corner.
- It'd be nice if the window were resizable horizontally and made a bit wider by default.
- The borders of the box above the layers should merge with the window borders.
- Very long filenames stretch the window, instead the filename should be middle-truncated.

Kudos for tackling the layers window, though! :)

dsizzle:
I think all of your comments are addressed, except the horizontal scaling - it really messes things up to allow that.  I did make the window a little wider. 

@korli do you know why the CI builds are failing?  it complains about not being able to find the 'make' command?

humdingerb:
I thought I tested your last push, did a clean build... Anyway, now the scroll arrow is indeed visible.

More observations:
- The whole scrollbar needs to move 1 pixel to the right to avoid a "thick" border with the window border.
- The scrollbar behaves a bit strange, e.g. it gets a scroll-knob when resizing the window, even if it still isn't needed. Even if you have just one layer, after making the window bigger and smaller again, you can scroll the layer out of the visible area.
- For some reason, transparency isn't working. It's all black...
- The file name is replaced with random symbols when adding a layer.
- The layers are sorted the wrong way round. "Add layer in front" ends with a layer below the current layer. (I always wondered, if we should go from "front/behind" to "above/below". It fits better to the Layers window and is equally real-worldy, considering a stack of sheets.)
- I also wonder if we need to add/merge layers "in front" and "below". We could just "Add layer" and "Merge layer" and always just "add above" and "merge with below". Saves menu items, and may come in handy if we have icon-buttons for all actions for quicker access.

dsizzle:
> For some reason, transparency isn't working. It's all black...

 Which part is black? The canvas is black? It works perfectly on my end.

humdingerb:
Yep. Anything translarency related. If I launch ArtPaint and create a new canvas, it's black.
The only place I get the alpha-chckers is in the Layers window. I have completely removed all objects folders and built from scratch.

artpaint/layers/LayerView.cpp
```diff
@@ -28,76 +31,30 @@
...
-	popUpMenuField = new BMenuField(BRect(0,0,0,0),"popUpMenuField","",layer_operation_pop_up_menu);
-	popUpMenuField->ResizeToPreferred();
-	popUpMenuField->MoveTo(visibility_box->Frame().left,visibility_box->Frame().bottom+4);
-	popUpMenuField->MenuItem()->SetLabel("");
-	AddChild(popUpMenuField);
+	BGroupLayout* layerLayout = BLayoutBuilder::Group<>(this, B_HORIZONTAL, 5.0)
+		.Add(visibility_box)
+		.Add(thumbnail_view)
+		.Add(layer_name_field)
+		.SetInsets(5.0, 5.0, 5.0, 5.0);
```
humdingerb, line 57: It may be a good idea to use constants from the InterfaceDefs.h for all insets, e.g. B_USE_SMALL_INSETS.

artpaint/layers/LayerView.cpp
```diff
@@ -112,33 +69,39 @@ LayerView::~LayerView()
...
 {
-	DrawBitmap(the_image,BPoint(5,5));
+	thumbnail_view->Redraw();
 	if (is_active == TRUE) {
-		BRect a_rect = the_image->Bounds();
-		a_rect.InsetBy(-1,-1);
-		a_rect.OffsetTo(4,4);
-		SetHighColor(0,0,255,255);
+		BRect a_rect = thumbnail_view->Frame();
+		SetHighColor(0, 0, 255, 255);
```
humdingerb, line 90: Better use ui_color(B_CONTROL_HIGHLIGHT_COLOR). If that's the right constant...

humdingerb (commented):
Maybe my transparency issues stem from changes in paintwindow/Image.cpp ?

dsizzle:
I did change the main rendering loop, which is where we started this whole game - I think my very first fix in ArtPaint was the endianness fix for the black canvas. Maybe I will back out that part of the change for now. I am quite grateful for your testing!!

dsizzle:
Fixed some but not all:

- fixed filename truncation so it doesn't give weird symbols
- fixed layer order
- moved scroll bar over by 1 pixel

I still haven't seen the black canvas issue which is concerning. I'll still pull that change out but wish I could reproduce it because I want that change for the layer compositing modes.

humdingerb:
HA!
I found out how to the transparency working again, but I have no idea why that is.
I had all "Keep in front" checkmarks in the "Windows" settings disabled. After ticking a checkmark and pressing OK, the transparency was back; checkers and all.
Even after removing the checkmark again, the checkers stayed.
It's back to black, if I delete all settings.

humdingerb:
Oh yeah, your other fixes are all good. The scrollbar, though, needs to grow 1 pixel at the top and bottom, I think. The Magnify app is an unforgiving tool... :)

humdingerb:
The Layers window isn't resizable anymore.

dsizzle:
THANKS for digging into that!!  If I delete the settings I am able to reproduce the black canvas problem.  I submitted the fix as a separate PR as it doesn't have anything to do with this change and I think it should be pulled into the mainline separately. (#227)

Scroll bar / resizing still in progress...

dsizzle:
Ok, I fixed the resizing, and moved the scrollbar down 1 pixel.  I can't seem to change the top pixel for some reason, perhaps because of how the top and bottom layout butt against each other?

Also still need to figure out the scroll bar still.

humdingerb:
I still cannot manually resize the Layers window.

Which also rises a question: Can we somehow manage to simultaneously resize the window automatically, and respect a manual resize?
On the one hand, it's nice to make the Layers window as small as possible and auto-grow the height when a new layer is added. (Shouldn't grow outside the screen, of course.)
On the other hand, users may want to limit the window height and rather scroll up/down, in order not to obscure too much of the canvas.

Maybe ArtPaint can remember the window height it was last _manually_ resized to. Keep the window height as small as possible while auto-growing/shrinking with additional/removed layers. When the previous manual-height is reached, stop auto-growing.

Sounds reasonable?
Something for later, i.e. create an enhancement ticket?

korli:
The CI image is rebuilt [every week or so 1](https://github.com/hectorm/docker-qemu-haiku/issues/1#issuecomment-1086655709). Please be patient, or ignore this for a few days :)

dsizzle:
I just deleted everything (including the settings) and rebuilt, and resizing still works for me. There is a minimum size for the window, but otherwise it should resize  - see this horrific example:

1layer

You can file the enhancement for resizing as you describe but it sounds like a challenging bit of work.

dsizzle:
Ok, after a lot of head banging and a desperate reachout in the forums and then more head banging I got the scrollbar to work properly.  Whew.  

Mostly.  It doesn't update the scrollbar when you add a new layer.  But otherwise it works as it should.  If I can figure out the update-on-new-layer then I *think* that's the last piece here.

oh, i just remembered the "above/below" change too.

dsizzle:
ok, I figured out how to make the scrollbar update.

Also changed the language to use "above" and "below" and I think it makes way more sense now.  

I think this is all working; please give a try when you have a moment and I'm sure you'll find things to break. :laughing: 

also waiting on the CI build to work again...

dsizzle:
I am working on the layer name saving/loading; I feel like that should be part of this since I'm showing the names now.  

Also I think drag-to-reorder is a bit broken...will fix that too. 

Once I fix those I'm ok not waiting for the CI bot if you are ok with that.  Just didn't want to break protocol.

korli:
@dsizzle please rebase this PR on master to get the CI update. Thanks.

dsizzle:
rebased to get CI fix, thanks @korli !

hmm. I fixed the visual of layer reordering, but if you resize the window the layers revert to the original order.  I must've broken something with the layer ordering somewhere along the way. :frowning_face:

humdingerb:
It's really quite weird. The layer mayhem is not just visible in the Layers window, but is seem in the canvas as well. (You're most probably very aware of this, but I thought I mention it, because it read like you were focusing solely on the Layers window.)

Good luck hunting that bug!

dsizzle:
Thanks, yes, I was aware of the canvas having the same mayhem - good word for it, btw!

I think I fixed the overall window behavior for dragging, but somehow stuff still gets out of order in the canvas?  but - at least taking steps toward the goal.

dsizzle:
Ok, I think I solved it!! Please give it a try when you have a chance, I believe all layer reordering should work as expected now. 

I also re-added disabling the merge menu items when there aren't layers above or below, and added a tooltip to the slider.  Unfortunately due to the layouting I couldn't figure out how to get the tooltip on anything except the slider itself.

artpaint/layers/LayerWindow.cpp
```diff
@@ -46,28 +63,87 @@ sem_id LayerWindow::layer_window_semaphore = create_sem(1,"layer window semaphor
...
+
+	BMenuBar* menu = new BMenuBar("menu bar");
+
+	menu->AddItem(layer_operation_menu);
+
+	BMessage* message = new BMessage(HS_LAYER_TRANSPARENCY_CHANGED);
+	message->AddInt32("value", 0);
+
+	transparency_slider =
+		new NumberSliderControl(B_TRANSLATE("\xCE\xB1:"), "0",
```
humdingerb, line 113: The alpha symbol shouldn't need to be translatable. Without a comment, nobody will guess what it is anyway. :)

artpaint/layers/LayerWindow.cpp
```diff
@@ -46,28 +63,87 @@ sem_id LayerWindow::layer_window_semaphore = create_sem(1,"layer window semaphor
...
+	BGridLayout* transparencyLayout = BLayoutBuilder::Grid<>(0.0)
+		.Add(transparency_slider, 0, 0, 0, 0)
+		.Add(transparency_slider->LabelLayoutItem(), 0, 0)
+		.Add(transparency_slider->TextViewLayoutItem(), 1, 0)
+		.Add(transparency_slider->Slider(), 2, 0, 2);
+	transparencyLayout->SetMaxColumnWidth(1, font.StringWidth("1"));
+	transparencyLayout->SetMinColumnWidth(2,
+		font.StringWidth("SLIDERSLIDERSLIDER"));
+
+	transparency_slider->Slider()->SetToolTip(B_TRANSLATE("Set Layer Transparency"));
```
humdingerb, line 127: Sentence casing: "Set layer transparency" or maybe just "Layer transparency" as the whole widget also just shows the layer transparency, if the user doesn't twiddle the slider knob.

artpaint/layers/LayerWindow.cpp
```diff
@@ -143,9 +260,9 @@ void LayerWindow::ActiveWindowChanged(BWindow *active_window,BList *list,BBitmap
 		a_layer = (Layer*)list->ItemAt(0);
 	}
 
-	if (a_layer != NULL)
+	if (a_layer != NULL) {
```
humdingerb, line 263: No need for the curly brackets.

artpaint/layers/LayerWindow.cpp
```diff
@@ -198,19 +316,24 @@ LayerWindow::setFeel(window_feel feel)
 		layer_window->Lock();
 		layer_window->SetFeel(feel);
 		if (feel == B_NORMAL_WINDOW_FEEL) {
-			layer_window->SetLook(B_DOCUMENT_WINDOW_LOOK);
-			layer_window->scroll_bar->ResizeTo(layer_window->scroll_bar->Bounds().Width(),layer_window->list_view->Bounds().Height()-B_H_SCROLL_BAR_HEIGHT+1);
+			layer_window->SetLook(B_TITLED_WINDOW_LOOK);
```
humdingerb, line 319: No need for the curly brackets.

artpaint/layers/LayerWindow.cpp
```diff
@@ -238,68 +362,84 @@ void LayerWindow::Update()
 	if (must_update) {
 		layer_window->Lock();
 
-		// Here we must first remove any existing layer views.
-		while (layer_window->list_view->CountChildren() > 0) {
-			layer_window->list_view->RemoveChild(layer_window->list_view->ChildAt(0));
+		BGridLayout* layout = (BGridLayout*)layer_window->list_view->GetLayout();
+
+		while (layout->CountItems() > 0) {
```
humdingerb, line 367: No need for the curly brackets.

artpaint/paintwindow/ImageView.cpp
```diff
@@ -367,6 +367,11 @@ void ImageView::MessageReceived(BMessage *message)
 			}
 		}	break;
 
+		case HS_LAYER_TRANSPARENCY_CHANGED:
```
humdingerb, line 370: The other cases do have curly brackets...

artpaint/layers/LayerView.cpp
```diff
@@ -351,9 +306,57 @@ int32 LayerView::ReorderViews()
...
+	SetExplicitMaxSize(BSize(frame.Width(), frame.Height()));
+
+	frame.InsetBy(1.0, 1.0);
+	fThumbnailBitmap = image;
+}
+
+
+ThumbnailView::~ThumbnailView()
+{
+
```
humdingerb, line 333: Remove empty line.

humdingerb (commented):
Just a few minor comments.

humdingerb:
The layer ordering works very well now. Congrats for seeing this through! :)

What do you think about keyboard shortcuts for the Action menu items? Like:
* ALT+CursorLeft for  "Merge with layer above"
* ALT+CursorRight for  "Merge with layer below"
* ALT+CursorUp for  "Add layer above"
* ALT+CursorDown for  "Add layer below"
* ALT+D for "Duplicate layer"
* ALT+T for "Delete layer"

artpaint/paintwindow/ImageView.cpp
dsizzle: I left this one alone for now; about half of the case HAVE curly braces and half don't.  That file needs a cleanup overall.

dsizzle:
Thanks!  

I added the keyboard shortcuts, except I changed these two:
- ALT+SHIFT+CursorUp for "Merge with layer above"
- ALT+SHIFT+CursorDown for "Merge with layer below"

My reasoning was twofold: one, to keep with up and down arrows for above and below, and also adding shift makes the merging more deliberate and less likely to confuse with adding a new layer.

dsizzle:
Fixes #109 btw

humdingerb:
Makes sense.
So, how do you want to play it: Just merge #227 and build a quick bugfix hpkg? Or merge both #227 and this one, put the changes on Polyglot, wait a bit, then make a proper release?

humdingerb:
I found one little detail, just if you're not already fed up with the layer stuff...

Say you have 3 layers: L1, L2, L3. You select L2 and merge with above (L1), everything is perfect.
But if you merge with below (L3), L1 becomes the selected layer, instead of staying the now combined L2+L3.

dsizzle:
The benefit of merging just #227 with a quick fix is that we could also do a polyglot export. I think there are a few new languages. I’m not sure this layer change really justifies a release by itself, and it requires polyglot work for like 1 new string. Might wait until more features are added for a “full” release. Does this make sense? 

I would like to merge this change after that, in case I tire of layers and want to work on other stuff. I’m not sure if layer merging in general is correct/intuitive; if you merge L2 into L1 would you expect the layer to be called L1 or L2? It seemed to be the opposite of what I expected when I was playing with it.

humdingerb:
> Might wait until more features are added for a “full” release. Does this make sense?

Alright. Bugfix + translations it is.

> I’m not sure if layer merging in general is correct/intuitive; if you merge L2 into L1 would you expect the layer to be called L1 or L2? It seemed to be the opposite of what I expected when I was playing with it.

[Just for clarification, above I talk about what layer is selected after a merge, not the layer label.]
One could complicate things and distinguish if you're merging up or down, but IMO it's best to just be consistent and just keep the label of the layer that was selected when the merge was initiated.

humdingerb:
Arggl... I messed up on Polyglot by accidentally uploading the en.catkeys of this PR. That added the new strings and removed the old ones in all languages...  :(

Oh well, then we have just a quick bufix release and no new translations. Those come with the new layers window and whatever else you sprinkle your magic coder dust on.

dsizzle:
Yep, I see what you mean with the selecting the wrong layer when using "merge with below"

btw, ALT+T is already "Translate Layer" and ALT+CursorLeft/Up are shortcuts for flipping layers horizontally and vertically.  It was rather concerning when i was trying to move layers and they kept disappearing! :laughing:  Any other ideas for keyboard shortcuts?

also noticed that the Undo/Redo menu text still says "Merge with back/front layer" so I will fix that.

and finally "magic coder dust" is what's left when coders run out of tears. :smile:

humdingerb:
> btw, ALT+T is already "Translate Layer" and ALT+CursorLeft/Up are shortcuts for flipping layers horizontally and vertically. It was rather concerning when i was trying to move layers and they kept disappearing! 😆 Any other ideas for keyboard shortcuts?

Damn the scarcity of good shortcuts!
Truth be told, considering the small number of "actions" in that menu and that, as a graphics app, most users are quite mouse-centered, it may be best to use a tool bar with icon buttons instead.
We can whittle down the number of actions further by only doing a "merge down" and a "add layer above". Then we only need 4 buttons: add, delete, merge, duplicate.

I also experienced my shortcuts being eaten by focused "sub-windows". It'd be nice to have all shortcuts being sent to the main window, if they don't apply to a sub-window menu item (which should generally be avoided IMO, exception would be a to be added "DEL" press in the "Brushes" window to delete the current brush).
Is that possible?

> and finally "magic coder dust" is what's left when coders run out of tears. 

Oh dear. :)

BTW, whenever you want to draw a line under this PR, give the word.

dsizzle:
> Truth be told, considering the small number of "actions" in that menu and that, as a graphics app, most users are quite mouse-centered, it may be best to use a tool bar with icon buttons instead.
We can whittle down the number of actions further by only doing a "merge down" and a "add layer above". Then we only need 4 buttons: add, delete, merge, duplicate.

yeah, I initially made buttons but figured I'd stick with a menu for now.  I'm not sure how I feel about decreasing the number of actions as it's probably nicer to be able to add/merge in both directions.  

I also find myself trying to right-click to add/duplicate/merge so i wonder if having the menu pop up on right click would make sense even with buttons.

that said, there really SHOULD be a "merge selected" and "merge all" (aka "flatten image" in Photoshop-speak).  Currently this window only supports a single selection and really someday it should handle multi-select of layers for various kinds of actions

>  It'd be nice to have all shortcuts being sent to the main window...Is that possible?

Idk. it should be. There are currently unfortunate side effects to the way it works now.  For example if you merge layers and want to undo, you have to click on the main canvas. "undo" with the layer window selected does nothing. 

> BTW, whenever you want to draw a line under this PR, give the word.

I removed the shortcuts for now and I fixed the layer selection when you merge, so unless you find something broken I would be happy to see this merged. Thanks!

humdingerb:
> I also find myself trying to right-click to add/duplicate/merge so i wonder if having the menu pop up on right click would make sense even with buttons.

I'd prefer a context menu, too. Even more so currently, as all 'actions' relate to the selected layer. It's natural to right-click the layer to manilpulate and have all 'actions' right there.
I'm not sure sure I'd use it were there icon-buttons, but in case we get more 'actions' in the future, the menu would be a keeper.

> I'm not sure how I feel about decreasing the number of actions as it's probably nicer to be able to add/merge in both directions. [...]
> that said, there really SHOULD be a "merge selected" and "merge all" (aka "flatten image" in Photoshop-speak). Currently this window only supports a single selection and really someday it should handle multi-select of layers for various kinds of actions

Agreed to the multi-selection features. In case of merging multiple layers, the merging would be in one direction, i.e. merging with the below (the upper layer staying on top the ones below). I think it's OK to train the user that there's just one direction to merge: down.
I'm trained that way by Photoshop with it's CTRL+E shortcut when merging...

Anyway, I'll fire up the 32bit Haiku to check the build there and then merge.
The Layers window is much nicer now, and more functional. Well done!!

humdingerb:
Will merge after the 32bit buildfix. :P

dsizzle:
What's broken with the 32-bit build?  Do I need to fix something?

dsizzle:
sorry about that - should be fixed now.

humdingerb:
Thanks. Let's merge!

PR #227, opened 2022-07-31
Image: set default alpha for bg when there are no settings

dsizzle:
When the user has no settings for the alpha bg, the canvas may appear black.  This change sets the alpha for the default colors so that the bg renders correctly.


dsizzle:
Not sure if this should be merged urgently or not - it is an egregious bug for a new user to see an all black canvas. I’d hoped to get the layer changes in quicker but it is taking longer. Thoughts on doing a 2.3.1 for just this fix? I haven’t seen any complaints about it, though.

PR #228, opened 2022-08-06
Add 'targets' to build.sh script

humdingerb:
Adding 'targets' for:

main	:	Builds only the main app and moves it in the "dist" folder
addons	:	Builds only all addons and copies them in the "dist" folder
all		:	Builds main app and addons and puts them and the docs in the "dist" folder
clean	:	Removes all objects (main app and addons)
catkeys	:	Generates en.catkeys (main app and addons)

No argument assumes the target "all". Additional parameters are passed on to the
Makefiles, as before.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,8 +1,30 @@
 ArtPaint
 ======================
 ArtPaint is an award winning painting and image-processing program originally for the BeOS.      
-It now runs on Haiku:
+It now runs on Haiku (available through HaikuDepot):
 
 ![ArtPaintScreenShot](/artwork/ArtPaintScreenShot.png)
 
 See the [documentation](http://htmlpreview.github.io/?https://github.com/HakuArchives/ArtPaint/master/Documentation/index.html) for more information.
+
+### How to build
+
+Easy. Simply execute the script `build.sh` and you'll find ArtPaint, its addons and documentation in the folder `dist`.
+
+`build.sh` also takes optional parameters:
+
+```
+$> build.sh help
+A script to build ArtPaint
+Usage:  build.sh (action) [clean|debug|catkeys] target [all|main|addons]
+
+Targets (default is 'all'):
+        main    Builds only the main app and moves it in the dist folder
+        addons  Builds only all addons and copies them in the dist folder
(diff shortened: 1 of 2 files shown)


humdingerb:
@dsizzle: Would you be OK with this?

dsizzle:
Yes, I support this 200%!

Can you also add a “debug” target? That would save me some typing for sure.

humdingerb:
I added the option for a debug build.
Unfortunately, maybe I lack the right bash-fu, now specifying a target is mandatory. Otherwise I ran into issues when passing along more paramters for "make", like "-j8".

Anyway, e.g. "build.sh main" builds only ArtPaint, "build.sh debug main" builds it in debug mode.

humdingerb:
I think it'd simplify things quite a bit if we scrap the feature to pass along further switches and change
``` diff
- make OBJ_DIR="objects_addons" $*
+ make OBJ_DIR="objects_addons" -j
```
Or do you regularly use other switches?

dsizzle:
I am fine with that - I don't use any other switches.

humdingerb:
Cool. After a nice journey through bash scripting, I think it works quite well. For usage, see changed commit message.

humdingerb:
Currently a debug build is done in the regular "objects_artpaint" and "objects_addons" folders, clobering the plain built objects and the dist/ArtPaint executable. Should "objects_debug_artpaint|addons" be used instead (and "clean" remove all objects folders, plain and debug)? Shall the executable renamed "ArtPaint_debug" ?

dsizzle:
I think separate debug folders make sense.  Maybe put the debug at the end, like "objects_artpaint_debug" ?

humdingerb:
Done. The resulting debug ArtPaint executable is also named "ArtPaint_debug", so we can test side by side with the regular build.

humdingerb:
Guess we can merge. I added some info on building to the Readme as well.

dsizzle:
Thanks, sorry, i have been wanting to try it out but I am deep in this layer thing.

humdingerb:
A tough nut to crack, eh?
I'm not sure if you wanted to wait with merging #227 or not... Just give me a holler and it's done.

PR #232, opened 2022-08-19
Layers: Added color blend modes

dsizzle:
- Added color blend modes to Layers, very similar to Photoshop blending
modes:
	Normal, Multiply. Divide, Screen, Overlay, Hard light, Soft light,
	Vivid light, Pin light, Hard mix, Difference, Exclusion,
	Color dodge, Linear dodge, Color burn, Linear burn,
	Darken, and Lighten
- Blend mode is "applied" when merging layers, as is transparency.

### DIFF
--- a/artpaint/application/MessageConstants.h
+++ b/artpaint/application/MessageConstants.h
@@ -143,6 +143,4 @@
 
 #define	HS_MANIPULATOR_FINISHED			'MnFi'
 
-#define HS_LAYER_TRANSPARENCY_CHANGED	'LtXc'
-
 #endif
--- a/artpaint/application/PixelOperations.h
+++ b/artpaint/application/PixelOperations.h
@@ -11,13 +11,93 @@
 #define PIXEL_OPERATIONS_H
 
 
+#include <Catalog.h>
+
+
 #include <math.h>
+
+
+#undef B_TRANSLATION_CONTEXT
+#define B_TRANSLATION_CONTEXT "PixelOperations"
 
 
 union color_conversion {
 	uint8 bytes[4];
 	uint32 word;
 };
+
+
+enum BlendModes {
+	BLEND_NORMAL = 0,
+	BLEND_MULTIPLY,
+	BLEND_DIVIDE,
+	BLEND_SCREEN,
+	BLEND_OVERLAY,
+	BLEND_DARKEN,
+	BLEND_LIGHTEN,
+	BLEND_DODGE,
+	BLEND_BURN,
+	BLEND_LINEAR_DODGE,
+	BLEND_LINEAR_BURN,
+	BLEND_HARD_LIGHT,
+	BLEND_SOFT_LIGHT,
+	BLEND_VIVID_LIGHT,
+	BLEND_LINEAR_LIGHT,
+	BLEND_PIN_LIGHT,
+	BLEND_HARD_MIX,
+	BLEND_DIFFERENCE,
+	BLEND_EXCLUSION,
+	BLEND_DISSOLVE
+};
+
+
+inline const char* mode_to_string(BlendModes mode)
+{
+	switch(mode) {
+		case BLEND_NORMAL:
+			return B_TRANSLATE("Normal");
+		case BLEND_MULTIPLY:
+			return B_TRANSLATE("Multiply");
+		case BLEND_DIVIDE:
+			return B_TRANSLATE("Divide");
+		case BLEND_SCREEN:
+			return B_TRANSLATE("Screen");
+		case BLEND_OVERLAY:
+			return B_TRANSLATE("Overlay");
+		case BLEND_DARKEN:
+			return B_TRANSLATE("Darken");
+		case BLEND_LIGHTEN:
+			return B_TRANSLATE("Lighten");
+		case BLEND_DODGE:
+			return B_TRANSLATE("Color dodge");
+		case BLEND_BURN:
+			return B_TRANSLATE("Color burn");
+		case BLEND_LINEAR_DODGE:
+			return B_TRANSLATE("Linear dodge");
+		case BLEND_LINEAR_BURN:
+			return B_TRANSLATE("Linear burn");
+		case BLEND_HARD_LIGHT:
+			return B_TRANSLATE("Hard light");
+		case BLEND_SOFT_LIGHT:
+			return B_TRANSLATE("Soft light");
+		case BLEND_VIVID_LIGHT:
+			return B_TRANSLATE("Vivid light");
+		case BLEND_LINEAR_LIGHT:
+			return B_TRANSLATE("Linear light");
+		case BLEND_PIN_LIGHT:
+			return B_TRANSLATE("Pin light");
+		case BLEND_HARD_MIX:
+			return B_TRANSLATE("Hard mix");
+		case BLEND_DIFFERENCE:
+			return B_TRANSLATE("Difference");
+		case BLEND_EXCLUSION:
+			return B_TRANSLATE("Exclusion");
+		case BLEND_DISSOLVE:
+			return B_TRANSLATE("Dissolve");
+	}
+
+	return "";
+}
 
 
 // This function combines four pixels p1-p4 with weights c1-c4.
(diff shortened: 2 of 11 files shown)


dsizzle:
I don't have photoshop so I compared my results to what I found on the web, but I didn't have a reference for Vivid Light so I'm not sure if that one is correct. 

@humdingerb If you can compare Photoshop results to these I would appreciate knowing if any look odd.

dsizzle:
I tested with these two images:
normal
ducky

humdingerb:
Very nice!
Couldn't find any spoilers. :)
Apart from sentence casing "Blend mode", though maybe "Mode" is enough. Short is best, also when considering translations.

humdingerb:
WRT to comparing to PhotoShop. I can only do this at work next week. We also only have some cheaper, quite old version I'm not sure offers all those blending modes. But we'll see. Can also see what Krita etc. here on Haiku offers.

dsizzle:
Thanks! 

one thing I discovered; i shouldn’t allow/apply the blend mode on the bottom layer, because that doesn’t make any sense. There would be nothing to blend with.

also this gives more weight behind your comment that we should only merge down. The blending only works in one direction so merging up might not do what you want.

humdingerb:
I did comparisons with ArtPaint, Gimp and Krita:

[blending_artpaint,gimp,krita.zip](https://github.com/HaikuArchives/ArtPaint/files/9380430/blending_artpaint.gimp.krita.zip)

Best compared by flipping through them with ShowImage. There are some differences.
BTW, the amount of modes in Krita is insane...

dsizzle:
thank you for doing those tests!  this is hugely helpful. The ones that are wildly different are the ones i was worried about: color dodge, color burn, divide, vivid light, the banding on exclusion. 

I'll take a look at Krita's blending options - there are certainly more that can be added, and it's pretty easy at this point to add more - the trick is just getting the math correct.  Probably don't want TOO many more at least with the current single-dropdown menu interface...

anyway thanks again, this really will be great to compare when I try to fix the broken ones.

dsizzle:
I fixed color dodge/burn, divide, and exclusion - vivid light is being elusive.

humdingerb:
Yeah, Krita may be a good ressource to compare code or for inspiration. Never should ArtPaint include the same amount of blending modes...

dsizzle:
Ok, fixed Vivid Light, so all the modes now behave as per Krita. Gimp seems to be slightly different, not sure why, but my images seem to match Krita pretty much exactly. 

I also removed Merge with above layer and Add layer below, and changed the dropdown to "Mode:" instead of "Blend Mode:"

Krita is pretty crazy with the modes.  I don't think I want to add all of those.  One that Photoshop has that I didn't add is "Dissolve" - it's trivial to add because it just picks either the top or bottom layer randomly so you get a speckled look, but not sure how useful that is.

humdingerb:
> One that Photoshop has that I didn't add is "Dissolve" - it's trivial to add because it just picks either the top or bottom layer randomly so you get a speckled look, but not sure how useful that is.

I'm not a big image manipulator myself, but contrary to Krita that seems to suffer the common open source project's problem of a incorporating a hundred people's code/ideas, Photoshop normally has the needs of its users pinned down. So, there may be a good reason PS inculdes the "Dissolve" mode...

One thing: You may want to add separators in the Mode menu to group categories of blending modes together. Makes it easier to parse.

As usual, shout if you're ready to merge.

humdingerb:
Now that the Actions menu has slimmed down nicely.... :)
What do you think of renaming it to "Layer" and shorten the items to "Merge down", "Add", "Duplicate", "Remove"?

humdingerb:
The layout would look better IMO, if the labels of the Mode menu and the transparency slider were both in the grid layout's first column:

Layers

It looks best, if both labels are similarly wide, so we should retire the alpha-symbol and re-instate the word "Alpha". Though translations cannot guarantee that, so it'd be best to have the labels right-aligned, in case one label is wider than the other.
Can we also remove the filename-top_view? I think we're both of the opinion that it doesn't serve a purpose and it kinda disrupts the clean layout...

dsizzle:
- renamed Actions to Layer
- shortened menu item strings
- removed filename
- grid-aligned controls as above (though I realized you have mode on the top and I think I have mode on the bottom)
- added separators
- moved the blend modes to an enum/return strings so I can find the items in the menu based on string rather than relying on them always being in the right order.  Because of this i also grouped them the way I believe photoshop does, and added separators
- added Dissolve

I saw the bug about not having the mode change in the undo queue...def want to fix that before merging.

humdingerb:
A few more things (do tell me if I should open tickets for them):
* The "Delete" item should be disabled when there's only one layer left. The BAlert "Cannot delete the only layer..." can be removed.
* On 64bit, the Layers window can be made so narrow that the scrollbar disappears. For some reason that is not the case on 32bit. What?!
* I wonder... if the mode switching is added to the undo-history, should re-arranging layers be as well? How about the visibility checkbox and alpha setting?
* "Duplicate" appears as "Add layer" in the undo-history. No biggie if that is complicated to change...

humdingerb:
Even more comments, sorry...
* In the main window's Layer menu, the "Change transparency..." and the resulting setting window can be removed.
* The "Clear layer" menu can stay in the main window's Layer menu, but can be added to the Layers window's menu.

dsizzle:
>     The "Delete" item should be disabled when there's only one layer left. The BAlert "Cannot delete the only layer..." can be removed.

I fixed this - but left the code for the BAlert. I may go back and delete it later.

>     On 64bit, the Layers window can be made so narrow that the scrollbar disappears. For some reason that is not the case on 32bit. What?!

uh...please file a ticket for this one

>     I wonder... if the mode switching is added to the undo-history, should re-arranging layers be as well? How about the visibility checkbox and alpha setting?

yes - please file tickets for these.  It looks more complicated than I thought.

>     "Duplicate" appears as "Add layer" in the undo-history. No biggie if that is complicated to change...

Duplicate uses the Add layer function so imho it's not worth making the code aware of whether it's an add or duplicate. 

>     In the main window's Layer menu, the "Change transparency..." and the resulting setting window can be removed.

Hmm. I left that on purpose, in case someone has the Layer window closed?

>     The "Clear layer" menu can stay in the main window's Layer menu, but can be added to the Layers window's menu.

I was thinking of adding all of the Layer operations to the main window's Layer menu.  Can you think of any reason not to? I think this also would be a step towards removing the Layer window's menu and making it a pop up on right-click instead, and/or adding the buttons.

humdingerb:
[undo history]
> yes - please file tickets for these. It looks more complicated than I thought.

Isn't it always... :)

>> In the main window's Layer menu, the "Change transparency..." and the resulting setting window can be removed.

>Hmm. I left that on purpose, in case someone has the Layer window closed?

Even then, I think it's nothing gained for the users: they have to invoke a menu item, a window pops up, they have to move a slider. Plus clicking "OK". They can just as well invoke the menu item to open the Layers window and go from there. Also, IMO if you're dealing with multiple layers - otherwise there's no need to fiddle with the transparency - you really need to have the Layers window open to make sure you're drawing in the right layer.

> I was thinking of adding all of the Layer operations to the main window's Layer menu. Can you think of any reason not to?

I think that's a good idea! That avoids having to filter out which commands to except in the Layers window and everything can be forwarded to the main window. 

> I think this also would be a step towards removing the Layer window's menu and making it a pop up on right-click instead, and/or adding the buttons.

I'm all for adding buttons for the currently available, most used actions (add/delete/merge/duplicate). A pop-up menu would also be need, best if the Layer menu of the main window were cloned. Then you can directly right-click some layer and flip or whatever it.

dsizzle:
Ok, I added the Layer operations to the main Layer menu - but I also left them in the Layer window. 

I also agree with your thoughts on the Transparency manipulator so I commented out that code. 

I think if everything works fine on your end I'd like to merge this one and leave the undo/redo as the separate tickets.

humdingerb:
Alright. I found 2 more minor peculiarities with the 32bit build I'll open tickets for.
Greatly improved Layers window, well done! Merge!

PR #244, opened 2022-08-24
Global Settings: use width of localized "Grid size" string to set minimum window size

### LINKED ISSUE #230: Transparency tab of preferences has a weird layout
ArtPaint-prefs-trans

dsizzle:
Feels a bit arbitrary, but it works.

Fixes #230

### DIFF
--- a/artpaint/windows/GlobalSetupWindow.cpp
+++ b/artpaint/windows/GlobalSetupWindow.cpp
@@ -857,7 +857,9 @@ GlobalSetupWindow::GlobalSetupWindow(const BPoint& leftTop)
 	for (int i = 0; i < fTabView->CountTabs(); i++)
 		width += fTabView->TabFrame(i).Width();
 
-	fTabView->SetExplicitMinSize(BSize(width + 2 * be_plain_font->Size(), B_SIZE_UNSET));
+	width += be_plain_font->StringWidth(B_TRANSLATE("Grid size:"));
+
+	fTabView->SetExplicitMinSize(BSize(width, B_SIZE_UNSET));
 }
 
 


humdingerb:
Let's keep this one open a few days in case some layout sage (@pulkomandy ?) happens upon this PR and has knowledge of the why and how and maybe even has a better solution.

dsizzle:
The issue is that the current method of sizing is based on the with of the tab labels, and in the case of English it happens to be wider than the view. In French, the view is wider than all the tab labels added together.

I think the *right* solution is to find the maximum tab _view_ width, but I tried a couple of things and couldn’t seem to get the proper frame for the laid-out tab views.

dsizzle:
Also, rather than relying on serendipity(?) I might try making a small sample and posting it on the forums. That always seems to work.

humdingerb:
:+1:

pulkomandy:
I'm not sure what the question is for me. The new code makes sense, it is computing a view width from a string width instead of from a font height, so the two sizes are in the same direction.

Without knowing how the window looks and what it is made of, it's hard to say anything more.

I think if you don't do anything special, BTabView should already size itself to fit the content of all views inside it, and you can set an explicit min width to make sure the tabs themselves also fit if they have wide labels, but doing so overrides the default minimum width computed from the inner views. So, if you really need both, you may need to do more complex things (maybe overriding some of the size computation methods). I wouldn't know exactly what you need to do, however.

dsizzle:
I might ask you to merge this as-is @humdingerb ; I am not feeling motivated to fix this better at the moment when there are more interesting problems. :D  we could keep the ticket open but this PR would at least give people a better experience.

humdingerb:
Agreed.

PR #246, opened 2022-08-25
Straight Line: disable size slider if "Adjustable width" is checked

### LINKED ISSUE #173: Straight line tool: Disable width slider if "Adjustable with" is checked
...because the width is determined interactively, after the line's start and end points are set by the user.

dsizzle:
Fixes #173


humdingerb:
It's good!

PR #247, opened 2022-08-28
AntiDitherer addon: Deleted blank lines in pt.catkeys

dsizzle:
Portugese catkeys file had blank lines that were causing a "Bad data" error during build.


humdingerb:
Good catch. Errors pass by so quickly when building all... :)

PR #251, opened 2022-09-08
Color window redo

### LINKED ISSUE #231: Colors window not font-sensitive
Setting a large font size releveals the issue:

colors

* colour squares in the palette view need to resize
* arrow widget to switch colour sets need to resize
* spacing of colour sliders need to adjust

[Mentioning #45, to add this issue to the meta ticket.]

dsizzle:
TL;DR, reworked the color palette completely. 

Whew. Ok. What a long journey that was!!  Hopefully there aren't too many bugs. One thing is that you probably have to delete your settings otherwise you might get a crash due to the changes to the sliders.  Maybe I can try to figure a way to handle it gracefully? Anyway:

- Changed layout to use LayoutAPI
- Replaced color sliders with slider derived from BSlider
- Removed YIQ and YUV sliders, and replaced with CIELAB slider to more closely align with Photoshop
** I did work on the YUV slider a bit and so I could put it back but it seemed to not be very useful. 
- Added localizable tooltips to the sliders
- Make color swatch and sliders use alpha checker bg
- Removed palette view from StatusView as it's too small to be useful anyway

Fixes #231


artpaint/controls/ColorSlider.h
```diff
@@ -0,0 +1,122 @@
...
+
+
+#include <Bitmap.h>
+#include <Slider.h>
+
+
+class ColorSlider: public BSlider
+{
+public:
+ 								ColorSlider(BRect frame, const char* name,
```
korli, line 21: Why are these constructors not aligned with the other methods?

artpaint/controls/HSVColorControl.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+	float h, s, v;
+
+	rgb2hsv((float)c.red, (float)c.green, (float)c.blue,
+		h, s, v);
+
+	float r, g, b;
+
+	for (int i = 0; i < 6; ++i)
+	{
```
korli, line 69: goes on the previous line.

artpaint/controls/ColorSlider.h
dsizzle: That’s a good question. Not sure what I did there…?

artpaint/controls/ColorFloatSlider.cpp
```diff
@@ -0,0 +1,291 @@
...
+	kSliderModificationFinished	= 'ksmf'
+};
+
+
+ColorFloatSlider::ColorFloatSlider(const char* label, const char* text,
+		BMessage* message, float minRange, float maxRange, bool layout,
+		bool continuous, border_style borderStyle, thumb_style thumbStyle,
+		uint8 resolution)
+	: BBox(borderStyle, NULL)
+	, fMinRange(minRange)
```
korli, line 40: is this style (comma on the next line) consistent with the rest of the source files?

artpaint/controls/HSVColorControl.h
```diff
@@ -0,0 +1,132 @@
...
+
+	h /= 60.;
+ 	uint8 i = (uint8)floor(h);
+ 	float ff = h - i;
+ 	float p, q, t;
+ 	p = v * (1.0 - s);
+ 	q = v * (1.0 - (s * ff));
+ 	t = v * (1.0 - (s * (1.0 - ff)));
+
+ 	switch(i) {
```
korli, line 52: missing space after switch

artpaint/controls/ColorFloatSlider.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+
+void
+ColorFloatSlider::MessageReceived(BMessage* message)
+{
+	switch (message->what) {
+		case kNumberControlFinished: {
+			fSlider->SetValue(_FixValue(atof(fFloatControl->Text())) * fMult);
+			_SendMessage(atof(fFloatControl->Text()));
+		}	break;
```
korli, line 98: here break gets indented. while in other places, it's just a space.

artpaint/controls/ColorFloatSlider.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+void
+ColorFloatSlider::SetValue(float value)
+{
+	value = _FixValue(value);
+
+	if (fSlider)
+		fSlider->SetValue((int32)(value * fMult));
+
+	if (fFloatControl) {
```
korli, line 141: superfluous braces

artpaint/controls/ColorFloatSlider.cpp
```diff
@@ -0,0 +1,291 @@
...
+			BBox::MessageReceived(message);
+		}	break;
+	}
+}
+
+
+void
+ColorFloatSlider::SetEnabled(bool enabled)
+{
+	if (fSlider)
```
korli, line 125: should be fSlider != NULL

artpaint/controls/ColorFloatSlider.cpp
```diff
@@ -0,0 +1,291 @@
...
+		fSlider->SetEnabled(enabled);
+
+	if (fFloatControl)
```
korli, line 128: ditto

artpaint/controls/ColorSlider.cpp
```diff
@@ -0,0 +1,480 @@
...
+ColorSlider::DetachedFromWindow()
+{
+	BSlider::DetachedFromWindow();
+}
+
+
+void
+ColorSlider::MessageReceived(BMessage* message)
+{
+	switch(message->what) {
```
korli, line 121: missing space after switch

artpaint/controls/MultichannelColorControl.cpp
```diff
@@ -0,0 +1,157 @@
...
+MultichannelColorControl::Draw(BRect rect)
+{
+ 	BControl::Draw(rect);
+}
+
+
+void
+MultichannelColorControl::MessageReceived(BMessage* message)
+{
+ 	switch(message->what) {
```
korli, line 94: missing space after switch

korli:
I don't know exactly which style you follow (Haiku or another one) for new files. So my comments might actually not apply.
The class headers could be made more readable using correct alignment and maybe 100 characters width.

artpaint/controls/CMYColorControl.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+
+#undef B_TRANSLATION_CONTEXT
+#define B_TRANSLATION_CONTEXT "ColorSliders"
+
+
+CMYColorControl::CMYColorControl(rgb_color c)
+ 	: MultichannelColorControl(c, "C", "M", "Y", "A")
+{
+	slider1->SetResolution(1);
```
humdingerb, line 26: To make translations easier, how about a B_TRANSLATE_COMMENT("Cyan", "For color model CMY") etc. for all tool tips?
While it's quite obvious for the standard RGB and CMY, HSV and CIELAB may be more challenging when you stumble over a term in isolation.
Generally, it's not a bad idea to use B_TRANSLATE_COMMENT more often to give hints to the translators.

humdingerb:
Wow, massive! I always enjoy seeing a new ArtPaint PR in the morning! :)

I haven't found any real bugs yet, and I had no problems with old settings either.
Here's a list of things related to the Colors window or the palette generally. Do tell me for which to rather file separate issues.

Colors window:

1. New color set: after creating a new set (8/16/32/64 colors), switch palette to the newly created set. 
2. Redraw issues when  waving a window in front of colour well.
3. Slider values needlessly precise, no floats needed.
4. Slider value text boxes could be narrower.
5. Increase slider bar height with font size.
6. Slider bars show garbage pixels at right end with big font size (resize rounding errors?).
7. Resize arrow widgets to switch color sets with font size.
8. If you're on color set 1 and press the arrow-left widget, it stays depressed.
9. The padding around the palette could be smaller, esp. with large font sizes it doesn't have to grow much.

Main window:

10. Resize fg/bg colour wells with font size.
11. When using the colour picker tool, the Colors window gets updated live, the fg/bg colour wells of the main window dows not. IMO, since the colour picker tool already has a live preview, the Colors window should also only get updated on mouse-up.

dsizzle:
> I don't know exactly which style you follow (Haiku or another one) for new files. So my comments might actually not apply. The class headers could be made more readable using correct alignment and maybe 100 characters width.

I'm _trying_ to make it Haiku-like so I make changes over time, but some of it is bad style copy-pasted around, like those leading commas.  The ColorSlider alignment is odd because that was pretty much copy/pasted from BSlider.h and it looks perfectly fine in Pe so I'm not sure what's going on there yet.  

I try for 80 character width when I see long lines but there's still probably a ton of places that go over that length - and maybe in this modern age 80 characters is too narrow?

artpaint/controls/ColorFloatSlider.cpp
dsizzle: some - it's a copy/pasted bad style.  I fixed a couple of other files with the same issue.

artpaint/controls/CMYColorControl.cpp
dsizzle: I added comments - I wasn't aware of B_TRANSLATE_COMMENT but it's a great idea especially for ambiguous terms like "value"

dsizzle:
For the Colors window, I addressed:

> 2. Redraw issues when waving a window in front of colour well.
> 3. Slider values needlessly precise, no floats needed.
> 5. Slider values needlessly precise, no floats needed.
> 9. The padding around the palette could be smaller, esp. with large font sizes it doesn't have to grow much.

Probably will address in this CR:

> 1. New color set: after creating a new set (8/16/32/64 colors), switch palette to the newly created set.
> 6. Slider bars show garbage pixels at right end with big font size (resize rounding errors?).

Comments:

> 4. Slider value text boxes could be narrower.

I've tried to work on this, not just for this layout.  I seem to be unable to get BTextControls to change size.  There must be a way but I haven't found it. 

> 7. Resize arrow widgets to switch color sets with font size.

These buttons are tiny images that seem to be a fixed size.  I tried replacing them with normal BButtons with "Previous" and "Next" labels but they were awfully big.  I also tried "<" and ">" but wasn't sure how I felt about them.  I might try that again.  Thoughts?

> 8. If you're on color set 1 and press the arrow-left widget, it stays depressed.

Again not sure about this because it's a BPictureButton.  I think as a BButton I could also disable it if it's color set 1.

Please make separate tickets for these:

> 10. Resize fg/bg colour wells with font size.
> 11. When using the colour picker tool, the Colors window gets updated live, the fg/bg colour wells of the main window dows not. IMO, since the colour picker tool already has a live preview, the Colors window should also only get updated on mouse-up.

humdingerb:
> These buttons are tiny images that seem to be a fixed size. I tried replacing them with normal BButtons with "Previous" and "Next" labels but they were awfully big. I also tried "<" and ">" but wasn't sure how I felt about them. I might try that again. Thoughts?

You could use ⯇ (\xe2\xaf\x87) and ⯈ (\xe2\xaf\x88) or I suppose one can resize vector icons and use BButton's SetIcon()?

Opened tickets #253 and #254

humdingerb:
Something's not right with the Colors window. It's using quite a bit of CPU in an thread in app_server. After a while sitting idle, I had it crash with this: 
[ArtPaint-4664-debug-11-09-2022-15-57-43.report.txt](https://github.com/HaikuArchives/ArtPaint/files/9543049/ArtPaint-4664-debug-11-09-2022-15-57-43.report.txt)

I don't think I saw that with the original PR, so it must've been introduced in one of the later commits.

dsizzle:
That was just a dumb thing I did when I was fixing the redraw of the color well....I put an Invalidate call inside the Draw method, so it was just constantly redrawing.  Pushed the fix just now.

dsizzle:
also fixed garbage pixels at the end of color sliders - i was drawing 1 pixel short of the end, and at small sizes it happened to be covered by rounding values, but at larger sizes it was stretching into more than 1 pixel.

humdingerb:
Very good!
There are still artefacts though when slowly dragging a window in front of the Colors colour well.
Also, when horizontally resizing the Colors window, ideally only the sliders should grow, not the left hand pallete.

dsizzle:
> Also, when horizontally resizing the Colors window, ideally only the sliders should grow, not the left hand pallete.

oh, I thought one of the concerns was that the palette is too small with large color sets, so I thought it's a good thing to stretch it out?

humdingerb:
That's actually a good argument... I rarely use palettes, and mostly stick to 16-colours which are easy enough to hit.

dsizzle:
Ok, I changed the buttons to BButtons and used the unicode characters you suggested - it looks nice IMO.  I also moved the buttons to the side because I think it makes better use of space in the dialog.  I added tooltips to them as well. 

I messed with the sizing and resizing and ultimately made the color palette NOT resize  with the window as you mentioned.  It's larger so even 64-color palettes are pretty easy to click, and expands with text size, so I think it's good this way and certainly looks nicer than before.

The palette buttons also enable/disable based on the position in the list and whether there is a palette before or after.  

Also implemented the change to make it select the last color set when you add a new one.

I tried moving a window both slowly and quickly over the color well and I'm not seeing any artifacts so maybe it's just system differences?  Idk.  

I think this is all I planned to do here (and I covered almost all of your list), so please let me know if you have any additional comments or see odd behavior.

humdingerb:
Very nice! Good idea to move te arrow widgets to make the palette larger. Is the bit of padding to the right of the buttons intentional?
Can you change the tool tips to sentence casing, "Previous/Next color set"? :)

I should've given more detail WRT the artifacts: You need to have a half-transparent color, because only the alpha-checkers are affected. Her the Brushes window gets dragged diagonally over the color panel:
 
artefacts

dsizzle:
> Is the bit of padding to the right of the buttons intentional?

Yes, I felt like the arrow buttons needed a little separation from the slider labels, and the card layout doesn't have insets. I didn't want to add more spacing to the overall layout and that seemed like a simple way to group the buttons with the palette that looks ok.

> Can you change the tool tips to sentence casing, "Previous/Next color set"? :)

Fixed. One day I will do that without you having to ask. 

> You need to have a half-transparent color, because only the alpha-checkers are affected.

Thanks!  Fixed!

dsizzle:
Thanks, that was a super simple fix but I also realized there’s a math bug in the color sliders so some gradients are incorrect. I’m working on that.

dsizzle:
- Fixed arrow buttons on New color set
- Fixed color slider gradients
- changed CMY slider to CMYK since CMYK is more standard

dsizzle:
- added drop support for the color well (#257)

humdingerb:
Very nice! A re-generated en.catkeys and ready to merge? :)

dsizzle:
Sounds good to me - updated the catkeys and also removed an accidental debug message I left in there.

Issue #257, opened 2022-09-14
Accept dropped colours


humdingerb:
The large colour well at the right of the Colors window could accept dropped colors (and maybe also dropped text when it fits the hex format '#ffffff').
Bonus points for each color square in the palette to the left to accept the same. :)

BTW, the dropping is already working for the color wells of the main window.

dsizzle:
You can already drop colors onto the palette.

Ad mentioned in #258 I don't understand where you can drag a hex string from and if you have an example then I can try to make that work but otherwise not sure how to even drag a string of any kind?

PR #258, opened 2022-09-18
ColorPalette: add hex color field

### LINKED ISSUE #256: Add a way to specify colors by hex code
Most art programs seem to have a way to specify colors from a hex code; we should find some real estate in the Color palette window to add a hex code field. Perhaps shrink the color well and add the hex field below it?

dsizzle:
Added text control below color well in color palette window. Accepts:
- with or without '#'
- 3 digits, like html - assumes alpha of FF -- '123' -> '#112233FF'
- 4 digits, as above but with alpha -- '1234' -> '#11223344'
- 6 digits - assumes alpha of FF -- '123456' -> '#123456FF'
- 8 digits -- '12345678' -> '#12345678'

One caveat: setting the hex color via text does NOT update the selected color in the main window.  The problem is that ArtPaint uses whichever button was used to click on the wells and sliders to decide whether to update the fg or bg color, and in this case we have no button information.  I didn't want to assume that the user always wants to edit the fg color. After you type in (or copy/paste) a hex value, you can either right- or left-click on the palette entry.  

Fixes #256

### DIFF
--- a/artpaint/controls/CMYColorControl.h
+++ b/artpaint/controls/CMYColorControl.h
@@ -62,7 +62,7 @@ CMYColorControl::rgb2cmyk(float r, float g, float b,
 	b /= 255.;
 
 	k = 1.0 - max_c(r, max_c(g, b));
-	float inv_k = 1.0 - k; // + 0.00001;
+	float inv_k = 1.0 - k + 0.00001;  // guard against divide-by-zero
 	c = ((1.0 - r - k) / inv_k) * 100.;
  	m = ((1.0 - g - k) / inv_k) * 100.;
  	y = ((1.0 - b - k) / inv_k) * 100.;
--- a/artpaint/controls/ColorPalette.cpp
+++ b/artpaint/controls/ColorPalette.cpp
@@ -165,11 +165,43 @@ ColorPaletteWindow::ColorPaletteWindow(BRect frame, int32 mode)
 		B_SIZE_UNSET));
 	colorPreview->SetExplicitMaxSize(BSize(color_control->Bounds().Height(),
 		B_SIZE_UNSET));
+	hexColorField = new BTextControl("", "hex-color", new BMessage(HEX_COLOR_EDITED));
+	for (uint32 i = 0; i < 256; ++i)
+		hexColorField->TextView()->DisallowChar(i);
+	hexColorField->TextView()->AllowChar('0');
+	hexColorField->TextView()->AllowChar('1');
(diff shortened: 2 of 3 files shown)


humdingerb:
> One caveat: setting the hex color via text does NOT update the selected color in the main window

Not sure that restraint is necessary. After all, if you twiddle a slider after entering the hex, it's the fg colour in the main window that gets auto-updated.

What are your thoughts of using SetModificationMessage() to react on every entered/pasted value and thereby avoid having to press ENTER or TAB to effect the colour change?
Drag'n'dropped hex values could also be accepted (actually similar to how dropped rgb_colors are already handled).

dsizzle:
Ok, fixed so that it updates the fg color.

> What are your thoughts of using SetModificationMessage() to react on every entered/pasted value and thereby avoid having to press ENTER or TAB to effect the colour change?

I'd rather not do this. Since the field accepts 3, 4, 6, and 8 digits, it would be very crazy as the field tried to react while the user is typing.  I'd rather it's explicit confirmation.

> Drag'n'dropped hex values could also be accepted (actually similar to how dropped rgb_colors are already handled).

You mentioned this also in the dragging colors ticket #257 too but I don't understand where you might drag a hex string from?  Can you give me an example?  You can already paste a value into the hex field.

dsizzle:
also rolled the fix for #263 in here btw

humdingerb:
> Since the field accepts 3, 4, 6, and 8 digits, it would be very crazy as the field tried to react while the user is typing. I'd rather it's explicit confirmation.

Good point. :)

> I don't understand where you might drag a hex string from? Can you give me an example? You can already paste a value into the hex field.

You may have an HTML document open in a text editor. From there you could doubleclick a colour value and drag'n'drop it into the ArtPaint palette. You'd save pressing ALT+C and ALT+V and keep the contents of the clipboard. Plus, pervasive d&d support is cool. :)

The other way around would also be useful: drag a color from ArtPaint into an editor to drop the hex value. This can already be done with Pe and Colors!.

As always, tell me when ready to merge.

dsizzle:
>  This can already be done with Pe and Colors!.

sorry if I wasn’t clear - I understood the idea but I needed a concrete example like this so I can see how it’s implemented. The d&d involves putting data into a BMessage but if you don’t name it properly etc it won’t be understood by whatever you’re dropping it on. I can look at Pe and colors! And match how they do it. 

But I can do that under #257, please merge this pr if it passes your testing. Thanks!

humdingerb:
Looks good!

PR #260, opened 2022-09-18
ToolWindow: add stroke to selected tool button to make it more obvious

### LINKED ISSUE #33: Tools window needs some form of 'focus' on the selected tool.
At the moment, you can't tell which tool you're using (other than the text at the bottom of the canvas window). It would be nice to have a border around the currently selected tool in the brushes/tools window.

dsizzle:
Added border around selected tool using ui color B_KEYBOARD_NAVIGATION_COLOR - I tried different UI colors and this seemed the best compromise between visibility and obtrusiveness,  B_WINDOW_TAB_COLOR was too much, B_MENU_SELECTED_BACKGROUND_COLOR was still a bit subtle.  

Fixes #33


humdingerb:
Good. I tried with only a 1-pixel border, but the 2-pixel border makes it more obvious.
Merge!

Issue #252, opened 2022-09-11
Resize fg/bg colour wells with font size


humdingerb:
The wells of the foreground/background colours should resize with increased font size.

dsizzle:
This ticket can be closed now due to #261

humdingerb:
Right'o!

PR #261, opened 2022-09-18
StatusView: main color wells and mag buttons resize based on font size

### LINKED ISSUE #254: +/- Zoom buttons empty with large font size
The main window's +/- buttons appear empty when using a system font size >25.

dsizzle:
Fixes #254 and #252


humdingerb:
Nice.
Q: should the tool icons be resized as well?

Issue #24, opened 2013-09-27
Canvas > Translate feature skews/skids drawing


soakbot:
**Creator: Richard Sjusen Nyhus**
**URL: http://dev.osdrawer.net/issues/607**

_CAUTION: comments not extracted from issues. See OsDrawer or the WaybackMachine for issue comments._

dsizzle:
I can't reproduce this at all.

richienyhus:
If my memory serves me correctly, this bug was found by a student in GCI 2012 and I was simply reporting it for them.

So I think we can close this issue as either the bug was never there to begin with or it's been fixed in the last decade.

dsizzle:
Cool, thanks for the background on this issue. It likely was fixed somewhere along the way.

humdingerb:
Not so fast. :)

I'm not sure the cryptic issue title is trying to say the same thing, but I do see artefacts when dragging a selection around with "Layer | Translate": http://0x0.st/obwM.mpg

* create a selection
* ALT+T to translate
* move the snippet around
--> it leaves a trace of pixels

If you don't translate a selection, but the whole layer, the artifacts only appear on the canvas bottom row of pixels.

dsizzle:
I can now reproduce this too. Seems to be easier to reproduce with a loaded image vs. a drawn image for some reason?  

- load image
- zoom out if the entire image isn't visible
- Alt+T to translate the image

![translate-smear](https://user-images.githubusercontent.com/10878750/165026154-24fdd310-d090-4ef2-8be4-7b53c274d740.png)

PR #265, opened 2022-09-21
Alpha selection fix: Render transparent selections properly when transforming

### LINKED ISSUE #120: Selection: alpha is not respected for selections in Layer mode
Select an area like this:
select-alpha01

...then use Layer->Translate:
select-alpha02

...the background of the selection is white, when it should be transparent.  In this instance the red stroke should also be translucent.

dsizzle:
Translation before this change; note how the transparent portion of the selection occludes the image:
before01
before02

Translation after this change; note how the transparent portion of the selection is respected:
after01
after02

Rotation before this change; again the transparent portion of the selection is occluding the image:
before01
before03

Rotation after this change; transparency of the selection is respected:
after01
after03

Also cleaned up the code style for TranslationManipulator and RotationManipulator a bit.

Fixes #120

### DIFF
--- a/artpaint/viewmanipulators/RotationManipulator.cpp
+++ b/artpaint/viewmanipulators/RotationManipulator.cpp
@@ -8,6 +8,7 @@
  *
  */
 
+#include "BitmapUtilities.h"
 #include "HSPolygon.h"
 #include "ImageView.h"
 #include "MessageConstants.h"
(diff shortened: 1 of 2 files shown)


humdingerb:
Good fix. BTW, do you see artifacts when moving a selection, too? Not when rotating though...

dsizzle:
I haven’t seen any artifacts on my end. Do you mean like #24? I had a hard time reproducing that issue too.

also are the artifacts only after this fix, or were they present before it?

humdingerb:
Yep, it's the same as #24, no change with this fix. It's still like in the mpg linked from there. The artifacts only appear when moving the selection up.

Let's merge this then.

PR #271, opened 2022-09-21
Minor GUI string changes

humdingerb:
The help texts of the tools didn't give much insight, most just saying "Click to use %tool%"...

Now we only show the name of the currently selected tool and only add useful hints, e.g. changed behaviour with modifier keys (which we may have more of in the future).


humdingerb:
Good thing is, the tool names are already translated, so there are only strings removed from the catkeys. No translation work involved. :)

dsizzle:
Looks good to me! Agree the “click to use” is rather silly.

humdingerb:
Thanks!

PR #272, opened 2022-09-22
Color Palette/Number controls: trimmed the fat

### LINKED ISSUE #268: A more compact Colors window
While updating the documentation, I noticed that the Colors window put on a bit of weight recently...

Before:
windows_colors

Now:
colors_new

Some things can't be helped much, but every pixel counts in the sum of all.
Suggestions:
* make the hex field and colour well above a few pixels narrower
* less horizontal padding between slider label, value box, slider, colour well
* the old problem of the rather bigger than required value boxes...
* the minimal width of the sliders coud be reduced. In case someone only really needs to switch between colours of the palette the window can be made smaller because high slider precision isn't needed.

### LINKED ISSUE #269: Colors: arrow widgets disappear for font size <12
With 11pt the widget labels disappear, with 10pt they are replaced with "...", lower and they disappear again. 
I think the Haiku default is 11pt Noto Sans Display. I normally run 12pt . Them old eyes...

dsizzle:
- decreased size/distance between controls for Color Palette window
- set max width of number controls to be maxBytes * stringwidth of a character

Not sure how much it helped- I tried to trim down the controls and I successfully shrank the text controls - this had the side effect of shrinking them throughout the app, but I looked around and mostly they look fine - except maybe the "canvas size" window.

Unfortunately I can't really shrink the hex value field...the widest value is #DDDDDDDD, and unfortunately most color values are far narrower than #DDDDDDDD, but in order to accommodate that color the field needs to be that wide.

Also, to make the arrow buttons not disappear, I had to make them a tiny bit wider.

Can't seem to get the sliders to shrink more.  The more I think I know about Layout API the less I think I actually know?

Fixes #269 
Fixes #268 ??

### DIFF
--- a/artpaint/controls/CMYColorControl.cpp
+++ b/artpaint/controls/CMYColorControl.cpp
@@ -29,7 +29,7 @@ CMYColorControl::CMYColorControl(rgb_color c)
 {
 	BGridLayout* mainLayout = (BGridLayout*)GetLayout();
 
-	BMessage *message = new BMessage(SLIDER_CHANGED);
+	BMessage* message = new BMessage(SLIDER5_CHANGED);
 	slider5 = new ColorFloatSlider("A",
  		"0", message, 0, 255, false);
  	slider5->Slider()->SetBarThickness(slider1->Slider()->BarThickness());
@@ -62,11 +62,33 @@ CMYColorControl::CMYColorControl(rgb_color c)
 }
 
 
+CMYColorControl::~CMYColorControl()
+{
+	slider5->RemoveSelf();
+
+	if (slider5 != NULL)
+		delete slider5;
+}
+
+
+void
+CMYColorControl::AttachedToWindow()
+{
+ 	slider5->SetTarget(this);
+
+ 	MultichannelColorControl::AttachedToWindow();
+}
+
+
 void
 CMYColorControl::MessageReceived(BMessage* message)
 {
- 	switch (message->what) {
- 		case SLIDER_CHANGED: {
+	switch (message->what) {
+ 		case SLIDER1_CHANGED:
+ 		case SLIDER2_CHANGED:
+ 		case SLIDER3_CHANGED:
+ 		case SLIDER4_CHANGED:
+ 		case SLIDER5_CHANGED: {
  			uint32 buttons;
  			BPoint point;
  			GetMouse(&point, &buttons);
--- a/artpaint/controls/CMYColorControl.h
+++ b/artpaint/controls/CMYColorControl.h
@@ -14,13 +14,18 @@
 #include "MultichannelColorControl.h"
 
 
+#define SLIDER5_CHANGED		's5Ch'
+
+
 using ArtPaint::Interface::ColorFloatSlider;
 
 
 class CMYColorControl : public MultichannelColorControl {
 public:
-		CMYColorControl(rgb_color c);
+					CMYColorControl(rgb_color c);
+					~CMYColorControl();
 
+		void		AttachedToWindow();
 		void		MessageReceived(BMessage* message);
 virtual	void		SetValue(rgb_color c);
 virtual	void		SetValue(float one, float two,
--- a/artpaint/controls/ColorFloatSlider.cpp
+++ b/artpaint/controls/ColorFloatSlider.cpp
@@ -47,11 +47,13 @@ ColorFloatSlider::ColorFloatSlider(const char* label, const char* text,
 {
 	_InitMessage();
 
+	BFont font;
+
 	fMult = pow(10, resolution);
 
 	fFormat.SetToFormat("%%0.%df", resolution);
 	fFloatControl = new (std::nothrow) FloatControl(label, text,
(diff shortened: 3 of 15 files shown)


dsizzle:
0palette
1palette

looks like I actually made it wider... :sob:

humdingerb:
Can the +/- buttons made smaller by using SetExplicitSize()?
Maybe we can shave off a few pixels from the palette after all. One pixel off each colour of the 64 colour set would yield 8 pixels...
The sliders are now wider than before and cannot be made smaller when resizing the window. Their minimal width could be reduced.
The hex box could use lower case letters, D0D0D0D0 is wider than d0d0d0d0.

humdingerb:
One thing I notice, the spacing between slider label and value box is rather large. As one example:

eraser

label-spacing

Granted, is has been large before this PR, but with the narrower value box width it's somehow more striking...

humdingerb:
Oh, and we have our familiar 32bit build failure, twice:

```
artpaint/controls/ColorPalette.cpp: In member function 'bool ColorPaletteWindow::openControlViews(int32)':
artpaint/controls/ColorPalette.cpp:687:34: error: call of overloaded 'SetVisibleItem(int)' is ambiguous
    sliderLayout->SetVisibleItem(0);
```
and:

```
artpaint/controls/HSVColorControl.cpp: In member function 'virtual void HSVColorControl::SetSliderColors(rgb_color)':
artpaint/controls/HSVColorControl.cpp:92:55: error: call of overloaded 'RemoveItem(int)' is ambiguous
    uint32* hue_word = (uint32*)(colorList.RemoveItem(0));
```
Cast it out. :)

dsizzle:
ok, I made it so you can shrink the sliders horizontally to be rather small. 

also changed hex field to lowercase.

I also added the SHIFT modifier to snap the sliders, but it really only makes sense for RGB and CMYK.  It makes no sense for CIELAB or HSV.  Also it doesn't work on the BeOS control because I have no control over that. Let me know if you think it's worth it.  I don't know about the CTRL because I don't think I have knowledge of how far the slides move.

Fixed the 32-bit casts.

> One thing I notice, the spacing between slider label and value box is rather large

Yes - I didn't want to go through and touch every place that used one of these fields but I suppose I will have to.  Pretty sure all the tool setup windows use exactly the same column widths for uniformity's sake, so whether the label was "Transparency:" or "Size:" the spacing is the same. As you point out it does look odd now.  I also need to fix the Create Canvas window. Maybe a separate PR from this?

dsizzle:
Also fun fact, the Alpha slider on the CMYK mode was totally not working and neither of us caught it... :smile: ...but it's fixed now.

humdingerb:
> I also added the SHIFT modifier to snap the sliders, but it really only makes sense for RGB and CMYK. It makes no sense for CIELAB or HSV.

I guess you're right, though for consistency, maybe HSV and CIELAB should just do the same(?). At least currently, they behave differently (possibly a bug):
HSV: SHIFT+click the H-slider throws S- and V-sliders to max. but keep increasing to >100.
CIELAB: Difficult to describe... The other sliders jump a bit, then follow the moving slider. Kinda like what was suggested for CTRL-clicking a slider.

> Let me know if you think it's worth it. I don't know about the CTRL because I don't think I have knowledge of how far the slides move.

The SHIFT-click does come in handy to set a grey tone for RGB and CMYK. So I'd say it is worth it.
The CTRL-click makes obvious sense only for RGB to change brightness. It may be useful to find "related" colour hues when used with the other colour models, who knows? When seen in action, it may turn out that moving those sliders in unison finds colours that in some way fit together...
It's your decision. You can also put it in a ticket for maybe later...

dsizzle:
For the HSV and CIELAB the problem is the vastly different scales. For example, Hue is 0-360, and S and V are 0-1.0. What’s the right thing to do? You can’t set the same value (which is what it tries to do now) but even setting the same horizontal position doesn’t make sense. Like HSV of (180, 0.5, 0.5) isn’t meaningful/useful, nor are (270, 0.75, 0.75) or (90, 0.25, 0.25) - essentially changing the saturation and lightness based on hue? 

What I may do is only enable shift click for rgb and HSV and disable for others.

for ctrl+click unless there’s another use I would say the RGB use case doesn’t make sense - just pop over to HSV and tweak S and V - that’s what they do. 😄

humdingerb:
Sounds reasonable.
Maybe some time in the future there will be the ultimate amalgamation with Colors!, but for now, I think we can be very happy with what we got .

dsizzle:
Ok, I actually disabled SHIFT+click for CMYK as well, so it's ONLY available for RGB.  The reason is because it doesn't really add any value to CMYK either. The way to get gray values with CMYK is to set CMY to 0, then just use the K slider. Setting CMY to the same value gives gray but then the K slider overrides it...so it's really an odd use of CMYK.  So, it's only active on RGB.

humdingerb:
Alright. Thanks for being so thorough for such a niche feature... :)
Read to merge then?

dsizzle:
Sure, go for it!

humdingerb:
Yay, wish I could trim fat with a click of a button... :)

PR #273, opened 2022-09-24
ColorPalette: more dragging/dropping goodness

### LINKED ISSUE #257: Accept dropped colours

dsizzle:
- add hex color value to drag message so you can drop it in a text window
- also accept hex values on color well

drag/drop between Colors! and ArtPaint works. I'm not sure how much worked before, but you can drag *from* ArtPaint to Colors! which I don't think worked before.  You could actually use Colors! as your color picker, because you can drag/drop from Colors to the palette and the fg/bg color slots on the main window too.  Interesting.

Fixes #257 

not sure if there are other drag/drop color scenarios still outstanding

### DIFF
--- a/artpaint/application/UtilityClasses.h
+++ b/artpaint/application/UtilityClasses.h
@@ -11,6 +11,8 @@
 #ifndef UTILITY_CLASSES_H
 #define UTILITY_CLASSES_H
 
+
+#include <String.h>
 #include <View.h>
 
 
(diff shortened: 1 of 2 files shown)


dsizzle:
I think dragging to and from colors! Already worked. Ah well. At least I added hex vale support to the color well.

humdingerb:
The magic of BMessages :)
The d&d to/from Colors! seems to have worked before this PR, too.
What I can't make work is the d&d of hex-values... neither from a text editor, nor the Colors! hex box, not with or without a "#", and not on ArtPaints color panels or hex-value box.

humdingerb:
FWIW, in Terminal I get this output when dropping a hex-value from Pe onto the colour pane of ArtPaint's COlors window:

```
BHandler w>Colors: MessageReceived() couldn't understand the message:
BMessage('DATA') {
        text/plain = (type = 'MIME')(size = 7)
        be:clip_name = string("Text Snippet", 13 bytes)
        buttons = int32(0x1 or 1)
        _drop_point_ = BPoint(x:724, y:301)
        _drop_offset_ = BPoint(x:19, y:11)
}
```

```
This is when dropping from Colors!' hex-value box:

BHandler w>Colors: MessageReceived() couldn't understand the message:
BMessage('MIME') {
        be:originator = (type = 'PNTR')(size = 8)
        be_actions = int32(0x4444524d or 1145328205)
        text/plain = (type = 'MIME')(size = 6)
        application/x-vnd.Be-text_run_array = (type = 'MIME')(size = 60)
        buttons = int32(0x1 or 1)
        _drop_point_ = BPoint(x:755, y:179)
        _drop_offset_ = BPoint(x:15, y:11)
}
```

dsizzle:
this PR broke other stuff so I'm closing it for now - but will come back to this issue.

PR #276, opened 2022-09-25
Clang static analysis: Cleaned up various potential issues

dsizzle:
Partially addresses #124


humdingerb:
It builds and I don't see any regressions after a short test drive...

PR #277, opened 2022-09-25
Translation: adjust bounds of drawing to prevent drawing artifacts

### LINKED ISSUE #24: Canvas > Translate feature skews/skids drawing

### LINKED ISSUE #160: Layer translate: Artefacts when dragging
* Open an image
* Choose _Layer | Translate..._
* Move the layer (or a selected area)

-> There are artefacts from the bottom edge when you're moving up

dsizzle:
I *think* fixed the translation artifacts, both with a selection and without - please give it a try. 

Fixes #24

### DIFF
--- a/artpaint/viewmanipulators/TranslationManipulator.cpp
+++ b/artpaint/viewmanipulators/TranslationManipulator.cpp
@@ -239,6 +239,10 @@ TranslationManipulator::PreviewBitmap(Selection* selection, bool full_quality,
 	else
 		last_calculated_resolution = lowest_available_quality;
 
+	// DEC: TEST disable low-res preview
+	last_calculated_resolution = 1;
+	//
+
 	if (last_calculated_resolution > 0) {
 		union color_conversion background;
 		// Transparent background.
@@ -271,10 +275,10 @@ TranslationManipulator::PreviewBitmap(Selection* selection, bool full_quality,
 //			uncleared_rect.PrintToStream();
 			uncleared_rect.left = max_c(0, x_translation_local);
 			uncleared_rect.top = max_c(0, y_translation_local);
-			uncleared_rect.right = min_c(width - 1,
-				width - 1 + x_translation_local);
-			uncleared_rect.bottom = min_c(height - 1,
-				height - 1 + y_translation_local);
+			uncleared_rect.right = min_c(width,
+				width + x_translation_local);
+			uncleared_rect.bottom = min_c(height,
+				height + y_translation_local);
 			to_be_cleared.Exclude(uncleared_rect);
 			for (int32 i = 0; i < to_be_cleared.CountRects(); i++) {
 				BRect rect = to_be_cleared.RectAt(i);
@@ -326,17 +330,18 @@ TranslationManipulator::PreviewBitmap(Selection* selection, bool full_quality,
 			uncleared_rect = uncleared_rect & preview_bitmap->Bounds();
 			for (int32 i = 0; i < to_be_cleared.CountRects(); i++) {
 				BRect rect = to_be_cleared.RectAt(i);
+				rect = rect & preview_bitmap->Bounds();
 				int32 left = (int32)(floor(rect.left /
 					last_calculated_resolution) * last_calculated_resolution);
 				int32 top = (int32)(floor(rect.top /
 					last_calculated_resolution) * last_calculated_resolution);
-				int32 right = (int32)(floor(rect.right /
+				int32 right = (int32)(ceil(rect.right /
 					last_calculated_resolution) * last_calculated_resolution);
-				int32 bottom = (int32)(floor(rect.bottom /
+				int32 bottom = (int32)(ceil(rect.bottom /
 					last_calculated_resolution) * last_calculated_resolution);
 
-				for (int32 y = top; y <= bottom;y += last_calculated_resolution) {
-					for (int32 x = left; x <= right;x += last_calculated_resolution) {
+				for (int32 y = top; y < bottom;y += last_calculated_resolution) {
+					for (int32 x = left; x < right;x += last_calculated_resolution) {
 						*(target_bits + x + y*target_bpr) =
 							*(source_bits + x + y * source_bpr);
 					}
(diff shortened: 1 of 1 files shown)


humdingerb:
I'm afraid I see no difference. It's still reproducible as described in #24. Good thing those artifacts aren't permanent; confirming or cancelling redraws the result correctly...

dsizzle:
Do you have an easy repro case? I find it very hard to reproduce but I can reproduce the problem 100% of the time with one particular image before this PR, and not afterwards

dsizzle:
I downloaded that Walter image like in your video in #24 and it works perfectly fine on my machine ???

humdingerb:
Strange, reproducible with any image. My go to is this (small and has transparency):

Walter_the_OS

* Make a selection
* ALT+T
* Move down and up -> artifacts when moving upwards
* OK or Cancel -> artifacts get cleaned up

dsizzle:
Quite unfortunate. I was only able to reproduce it with one image so far, and so I used that to test. I got artifacts before the PR, and they went away with the fix. I tried multiple times reverting and redoing the fix. :(

Are you using “real” Haiku or a VM?
What’s your screen resolution?
And it happens with any image? Even that yellow duck from the blend tests?

humdingerb:
I use Haiku hrev56454, 64bit on real metal.
Resolution 1920x1080x32, Noto Sans Display 12pt
I see those artifact with every image, Mr. Duck, too.

humdingerb:
Dunno if that helps, this is a when
* making a selection
* ALT+T, enter "20" in the "X" coordinate box and hitting TAB

ducky_translated

The bottom 12 pixels appear to to be moved with the rest of the selected part of the image. And a 2-pixel wide border of the original selection border position is transparent.

dsizzle:
I updated to hrev56475 as I was several revs behind, and I upped my resolution to 1600x1200 (I can't do 1920x1080 on this laptop) but I am using VirtualBox. 
Walter is fine, as is the Duck.  

I also tried your steps:
20ducky

yes, there is a pixel or two of space at the bottom/right of the selection but not a huge chunk of pixels like yours.

humdingerb:
I can confirm that the amount of screwed pixels decrease with the resolution. Here a shot at 1152x864:

screenshot3_1152x864

I think you should be able to run VirtualBox-Haiku in a window and just up the resolution in Haiku's Screen prefs for testing purposes. Or maybe try qemu?

dsizzle:
That’s interesting for sure… but I didn’t see a difference between 1152x864 and 1600x1200 on my screen. I am using Haiku’s screen prefs to change the res, I assume my laptops res is limiting my virtual box res somehow

humdingerb:
I just tried 1600x1200 and curiously, I don't see the issue there. Maybe my panel has a native 1920x1080 resolution and there's some squashing to 1080? No idea.

I'd have thought you could have any resolution when running Virtualbox-Haiku in a window. You should be able to force a resolution with https://www.haiku-os.org/guides/virtualizing/virtualbox#part_customVESA
It's been a long while since I ran Virtualbox...

dsizzle:
Hmm. Maybe it’s a 16:9 issue… I’ve only been using 4:3 resolutions. ArtPaint uses a lower resolution bitmap for various operations (like translation) so I wonder if the math doesn’t work right for widescreen… although it should depend on the resolution of the image, not the screen? 🤯

dsizzle:
I tried setting my VirtualBox to 1920x1080x32 and it worked fine - no artifacts.  I'm out of ideas.

humdingerb:
Frustration.

> ArtPaint uses a lower resolution bitmap for various operations (like translation) so I wonder if the math doesn’t work right for widescreen… 

Can that "lower resolution bitmap" thing be deactivated easily. Then you could test if it' the culprit. Also, that may be a remnant of ArtPaint's decades of history. Is such as trick still useful for today's maschines.? Maybe it can be only activated when dealing with huge images.

humdingerb:
Could there be a threading component to the issue as well?
I don't know how many cores you use with Virtualbox, but I think there are issues when running with multiple CPUs.
I have an 8 core i7 and when deactivating 7 cores via Pulse (running natively), I see much fewer lines affected by this bug. Bigger images show more than smaller ones.

dsizzle:
I tried adding more CPUs to my VM and I still can't repro the issue.

ok, I disabled the lower res for the translation manipulator; see if you notice a difference.

if there's no difference I might just toss this issue back on the shelf for now.

humdingerb:
I still always get those artifacts when moving a selection with the mouse. The amount of pixels that get messed up at the bottom of selection changes. I seems to depend where in the image I make the selection. I seem to get more when the selection is in the middle of the image than further to the top or bottom.
If I create several non-continguous selections (tool set to rect selections, add area), only the bottom selection produces the artifacts.

Finally, once more a zoomed in screenshot:

screenshot1

If I press ALT+T, a 1-pixel line (1) inside the bottom and right-hand selection border appears to be transparent.
Once I enter X=20 and TAB, the selection moves to the right, but some lines (2) of the image don't move with the selection.
When I press OK, those lines (2) jump to their correct position.

Anyhoo, before you hurt yourself banging your head against that problem much longer, feel free to put it on the backburner. Maybe inspiration strikes when you're thinking about something else.
Shall this PR be merged anyway? It might have improved the situation (at least I did have a few instances where I thought so)...

humdingerb:
Just to add: I don't see that transparent-line-border - the (1) above - in current master, i.e. without the changes in this PR.

dsizzle:
The reason for the transparent line is that I pulled back/rounded down the boundaries of the selection. If I can push the boundary back without losing the benefits then I’ll say yes to the merge. 

Thanks for all the time you’ve spent testing this!  I do appreciate it a ton. 

I wouldn’t hold the release for this - when you feel like we have enough translations just go for it.

humdingerb:
The trickiest thing is that you don't see the issues yourself, so I can imagine how hard it is to blindly poke at things.
It's all so chaotic. Sometimes I see no issues at all e.g. with ducky.png when at 100% zoom, or 200%, but they appear when at some odd factor or 300%. Next time I open it, the issue manifests at 100%, too...

> The reason for the transparent line is that I pulled back/rounded down the boundaries of the selection. If I can push the boundary back without losing the benefits then I’ll say yes to the merge.

It looks like with this PR's changes, things are just as haphazard as without, just with the additional transparent line thrown in... So maybe not worth it after all?

> I wouldn’t hold the release for this - when you feel like we have enough translations just go for it.

Yeah, I'd say we wait until after this weekend and then include all >90% finished translations.

dsizzle:
abandoning this PR; hoping #398 works!

PR #278, opened 2022-09-26
GlobalSettings: fix messages for cursor settings and remove quit confirm

### LINKED ISSUE #132: Global settings: Cursor setting not saved and doesn't work
The radio setting between "Tool cursor" and "Cross-hair cursor" doesn't seem to change the mouse pointer appearance. The setting isn't saved either.

I haven't seen the "Cross-hair cursor", I imagine it'll just be the small cross-hair cursor that's also there with the "Tool cursor". IMO, the small additional tool symbol at the mouse pointer never actually distracts or obscures anything important.
I propose to just remove the whole setting.

### LINKED ISSUE #133: Global settings: Remove "Confirm quitting" option
IMO it's never a good idea to close an unsaved document without confirmation.

If this setting gets removed, and #132 is removed as well, the whole tab "Miscellaneous" can be removed.

dsizzle:
Ok, the cursor setting problem was the silliest thing - the messages were opposite of the buttons!  So if you clicked "tool cursor" in the code it selected cross-hair mode, and vice versa.  That's why it didn't seem to ever work.  Now it works, you can choose cursor modes.  

possible improvements: 
- seems like the cursor stays as the finger pointer unless you leave the canvas and return. It'd be nice if it updated whenever you're on the canvas. Some kind of refresh issue?
- maybe my eyes are getting older but I find the cursors hard to see on the alpha checkerboard, even when I tried different colors. Maybe they can be made 2 pixels thick instead of 1?  

I also removed the quit confirm while in there.  I agree it's pretty useless.

Fixes #132 
Fixes #133

### DIFF
--- a/artpaint/windows/GlobalSetupWindow.cpp
+++ b/artpaint/windows/GlobalSetupWindow.cpp
@@ -702,7 +702,6 @@ private:
 
 		BRadioButton*	fToolCursor;
 		BRadioButton*	fCrossHairCursor;
-		BCheckBox*		fConfirmShutdown;
 };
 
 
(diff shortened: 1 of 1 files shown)


humdingerb:
oops, didn't want to merge just now...

Issue #282, opened 2022-09-29
Improve UX of Manipulator windows (Translate/Rotate/Crop/Scale)


humdingerb:
* Currently the windows can be lost under their canvas window.
  -> Float above their canvas window (canvas windows of other projects can move in front)

* Currently you can enter a value, but press ENTER or click OK and nothing gets moved/rotated/cropped/resized. You have to cycle the focus with TAB or click into another text box.
  -> When OK is clicked or ENTER triggers OK as the default button, the values should be applied.
  -> ESC should cancel and close the window.

* No need to have the window resizable

dsizzle:
What do you think about not having explicit confirmation of transformations?  Since ArtPaint has undo, a user can just translate or rotate, and then undo if they don't want it?  It feels non-standard to have to say Ok for every manipulation...I often forget to click Ok and then go do something else and the manipulation gets undone - especially if the window is hiding behind something.

Also, to be fair to the original author of ArtPaint, the window bit is my fault - some of these used to show up as controls in the Status View of the main window but I moved them out when cleaning up the layout.  Maybe there's a way to fit it back in the status view and then they won't get hidden. I'd still vote for applying the manipulations immediately, but don't want to lose the fields to enter the values.

Thoughts?

humdingerb:
> What do you think about not having explicit confirmation of transformations?

Not sure about that.
For one, every addon with options has OK/Cancel buttons. Those would need to be adjusted as well.
Then the user would have to click the small window-close-button to get rid of the window instead of the bigger OK button. Because close it, you must, as it blocks using another tool.
Having  OK/Cancel buttons isn't that bad. Optics-wise they make otherwise tiny windows a bit bigger (e.g. Translation, X/Y). ArtPaint could also offer to switch "Cancel" to "Revert" to the setting when the window was opened when holding CTRL or ALT, like Photoshop. I like that, saves cancelling and reopening the panel.
I think I prefer floating/modal windows. Functionally, they are modal already. Currently they can just get lost...

> Maybe there's a way to fit it back in the status view and then they won't get hidden.

Popping up a window for all panels may still be nicer. It's more in-your-face obvious. It's consistent. It offers easily extendable space if more options for the functions gets added in the future.

dsizzle:
I think this is the last challenge left:
> Currently you can enter a value, but press ENTER or click OK and nothing gets moved/rotated/cropped/resized. You have to cycle the focus with TAB or click into another text box.
[...]

PR #293, opened 2022-10-05
Crop manipulator: code cleanup, added shaded area, other features

### LINKED ISSUE #39: 'Crop to selection' feature

dsizzle:
Cleaned up code for Crop manipulator, and added some new features:

- dragging a corner of the crop box maintains the aspect ratio
- shade area to be cropped
- added "use selection" button that sets the crop rectangle to the bounds of the current selection
- added "reset" button that resets the crop rectangle to the full image bounds

Fixes #39 

C&C welcome with respect to wording and UI layout.  Note that the manipulator window is a predefined layout with the Ok and Cancel at the bottom so the Reset and Use Selection buttons can't be moved down to the bottom.

I know in Photoshop, "Crop To Selection" does an instant crop but this was easier to implement and felt more flexible?

### DIFF
--- a/artpaint/viewmanipulators/CropManipulator.cpp
+++ b/artpaint/viewmanipulators/CropManipulator.cpp
@@ -14,12 +14,14 @@
 
 #include "MessageConstants.h"
 #include "NumberControl.h"
+#include "PixelOperations.h"
+#include "Selection.h"
 
 
 #include <Bitmap.h>
 #include <Catalog.h>
 #include <GroupLayout.h>
-#include <GroupLayoutBuilder.h>
+#include <LayoutBuilder.h>
 #include <StatusBar.h>
 #include <Window.h>
 
@@ -34,11 +36,15 @@
 using ArtPaint::Interface::NumberControl;
 
 
-CropManipulator::CropManipulator(BBitmap *bm)
-	:	WindowGUIManipulator()
+CropManipulator::CropManipulator(BBitmap* bm)
+	: WindowGUIManipulator()
 {
 	settings = new CropManipulatorSettings();
 	config_view = NULL;
+	preview_bitmap = NULL;
+	copy_of_the_preview_bitmap = NULL;
+	use_selected = TRUE;
+	lock_aspect = FALSE;
 
 	min_x = 0;
 	max_x = 0;
@@ -48,12 +54,13 @@ CropManipulator::CropManipulator(BBitmap *bm)
 	if (bm != NULL) {
 		BRect rect = bm->Bounds();
 		settings->left = min_x = rect.left;
-		settings->right = max_x = rect.right;
+		settings->right = max_x = rect.right + 1;
 		settings->top = min_y = rect.top;
-		settings->bottom = max_y = rect.bottom;
+		settings->bottom = max_y = rect.bottom + 1;
 	}
-}
 
+	SetPreviewBitmap(bm);
+}
 
 
 CropManipulator::~CropManipulator()
@@ -67,7 +74,8 @@ CropManipulator::~CropManipulator()
 }
 
 
-void CropManipulator::MouseDown(BPoint point,uint32,BView*,bool first)
+void
+CropManipulator::MouseDown(BPoint point, uint32, BView*, bool first)
 {
 	previous_left = settings->left;
 	previous_right = settings->right;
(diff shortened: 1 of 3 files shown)


humdingerb:
I think it would look much better and isn't really a big loss to abandon the "anatomically correct" placement of the left/top/bottom/right text controls. Quick mockup:

cropper

I would also suggest to keep Left, Top but use Width and Height instead of the Bottom/Right.
You may often want an image of a specific width/height (say 256 pixels). Using all 4 coordinates, you have to mentally add the 256 to the Left or Top values.

Also: sentence casing... :)

Also 2: the frame and grabbers of the cropping selection are quite difficult to see, depending on the background (try the transparent checkers and they are practically invisible on the light greys).
I have a suggestion that I'll file in a new ticket.

dsizzle:
100% on all of this - I wasn't a fan of the "anatomically correct" :rofl: layout either. 

I will also add in #295 because it is super easy to do.

dsizzle:
Updated per above comments; lmk if you see any problems.  Thanks!

humdingerb:
A few comments:

* Doing the crop sometimes shears the resulting image. I haven't yet figured out how to 100% reproduce it. It looks like it happens most often if you move the crop-frame outside the canvas (negative coordinates).

shear

* Always keeping the aspect ration when manipulating the corner grabbers feels a bit odd. I find myself often not wanting that, and then having to move to a non-corner grabber for correction.
I think it'd be nicer to have a "Lock" checkbox beside the Height/Width like the Scale dialog has. There the button label is "Restore" where it's here "Reset". I think we should pick one, you decide. The "Use selection" and "Reset/Restore" buttons could move to the top to make room for the "Lock" checkbox.
Additionally, as for any tool, holding SHIFT could also temporarily lock the aspect ratio.

* Currently, clicking in the crop-area moves its center to the mouse-down point. This isn't so nice, because the area jumps quite jarringly. It would be better if you could grab the area anywhere and the area just moves together with the mouse.

* If you move the crop-area without changing its size, the crop appears not to be executed.

dsizzle:
- Fixed moving of the crop area so it follows the mouse movement instead of jumping to the mouse-down point. Feels a lot nicer!
- Fixed bug where if you only moved the crop without resizing, the crop was ignored
- Changed "Reset" to "Reset to canvas" as it's more descriptive. I don't like "Restore" for Scale and will likely end up changing it
- Added "Lock aspect ratio" to keep the same aspect ratio when resizing either via typing or interactively with the mouse. Holding CTRL also locks the aspect ratio

Any other ideas on the shearing?  I was only able to reproduce it one time, and tried many many more times unsuccessfully. :(

humdingerb:
WRT shearing I haven't found any pattern either... But today I find it harder to trigger it. Only once in over 20 tries, I'd say. Exausting... :)
How about we create a ticket for it, and see if we can get more info later? Maybe by actual user feedback.

* I think it may be a good idea to "use selection" by default when opening the Crop window, if there's a selection. People may be used to "Crop to selection" with other apps and create an appropriate selection by habbit. If they don't find a "Crop to selection" menu item, they settle for "Crop..." and are happy that their selection is already respected, without having to find, read, understand, and click the "Use selection" button.

* The "Use selection" button could be disabled when there's no selection.
The "Reset to canvas" button could be disabled when the crop area is equal to the canvas, but there's not as much potential for confusion as with the selection button ("What selection? There is no selection...").

* We should keep one label, "Lock" or "Lock aspect ratio", in both Scale and Crop windows. "Lock" is nice and short, though I don't mind either way.

* You said before that the Cancel/OK buttons of Scale/Crop are fixed. Could a "Revert" button be added there that would then be available wherever that kind of window is used? Then we had a revert feature everywhere and wouldn't need "Restore" buttons in the Scale window. Maybe a separate ticket...

* If you have your Crop window open and resized/moved the crop area, then choose any other manipulation window (Resize/Rotate/Translate/any addon/Crop again), the crop action is executed.
I think, as long as there are OK buttons, it's unexpected to have anything applied without explicitly clicking OK.
This is true for all these functions, Resize/Rotate/Translate/any addon etc. Something for a separate ticket?

* Scale reports "width" and "height", Crop reports "width - 1" and "height - 1". The file attribute set by ShowImage agrees with the Scale window.

humdingerb:
One more thing: Can we have SHIFT as aspect locking modifierer? It's already used for this with the Rectangle and Ellipse tools, and seems to be used by other apps as well for that sort of thing.

dsizzle:
> WRT shearing I haven't found any pattern either... But today I find it harder to trigger it. Only once in over 20 tries, I'd say. Exausting... :)
How about we create a ticket for it, and see if we can get more info later? Maybe by actual user feedback.

Perhaps. I have a suspicion as to what causes it, so I put in a fix for it, but I just can't confirm it.  

>  I think it may be a good idea to "use selection" by default when opening the Crop window, if there's a selection. People may be used to "Crop to selection" with other apps and create an appropriate selection by habbit. If they don't find a "Crop to selection" menu item, they settle for "Crop..." and are happy that their selection is already respected, without having to find, read, understand, and click the "Use selection" button.

I think this is a great idea, as there is no other real reason to have a selection before choosing Crop. It is mplemented. 

> The "Use selection" button could be disabled when there's no selection.
    The "Reset to canvas" button could be disabled when the crop area is equal to the canvas, but there's not as much potential for confusion as with the selection button ("What selection? There is no selection...").

This is unfortunately not possible right now - the UI has no idea if there is a selection or not, and at the point where the selection is handled it's not the right time to access the UI. 

> We should keep one label, "Lock" or "Lock aspect ratio", in both Scale and Crop windows. "Lock" is nice and short, though I don't mind either way.

I switched it to "Lock" - I was thinking there might be a difference between locking the ratio between the values and locking the two values to the same number, but it's really not. One is just a square aspect ratio. 

> Scale reports "width" and "height", Crop reports "width - 1" and "height - 1". The file attribute set by ShowImage agrees with the Scale window.

Fixed.

> One more thing: Can we have SHIFT as aspect locking modifierer? It's already used for this with the Rectangle and Ellipse tools, and seems to be used by other apps as well for that sort of thing.

Fixed.

>     You said before that the Cancel/OK buttons of Scale/Crop are fixed. Could a "Revert" button be added there that would then be available wherever that kind of window is used? Then we had a revert feature everywhere and wouldn't need "Restore" buttons in the Scale window. Maybe a separate ticket...

>    If you have your Crop window open and resized/moved the crop area, then choose any other manipulation window (Resize/Rotate/Translate/any addon/Crop again), the crop action is executed.
    I think, as long as there are OK buttons, it's unexpected to have anything applied without explicitly clicking OK.
    This is true for all these functions, Resize/Rotate/Translate/any addon etc. Something for a separate ticket?

Both of these should go in the UX ticket imho (#282)

humdingerb:
> [Shearing]. I have a suspicion as to what causes it, so I put in a fix for it, but I just can't confirm it.

Well, at least I couldn't accidental-shear when I just cropped a dozen times. So, probably, "Well done!". We'll see... :)

I amended ticket #282.
I suppose we're ready to merge here.

PR #297, opened 2022-10-08
Cursor: Made cross-hair cursor more visible

dsizzle:
With the checkboard alpha bg the cross-hair cursor was nearly invisible. I made it more solid to save my already bad eyes from getting worse.


humdingerb:
I like it!

PR #298, opened 2022-10-08
ImageView: code cleanup

dsizzle:
Nothing but code cleanup.  Tried to make the code more standard and Haiku-like.

PR #301, opened 2022-10-13
ImageView: add panning if the user holds down the space bar

### LINKED ISSUE #98: Pan canvas when holding SPACE
A handy shortcut many graphics app support.

dsizzle:
If you hold SPACE, drag to pan, then release the mouse button but keep holding SPACE, clicking again to pan doesn't always work.  Releasing SPACE and pressing it again works.  

I discovered that WonderBrush has exactly the same problem - not sure if it's a Haiku issue or my setup? 

Fixes #98

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -39,6 +39,7 @@
 #include <Catalog.h>
 #include <ClassInfo.h>
 #include <Clipboard.h>
+#include <Cursor.h>
 #include <MenuBar.h>
 #include <MenuItem.h>
 #include <Message.h>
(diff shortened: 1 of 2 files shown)


humdingerb:
Must be your setup. Works perfectly here.
With WonderBrush I sometimes get a "hanging" space key, or so it appears. I release the space bar, but the cursor is still the panning hand, and the mouse still drags the canvas. Once more hiting the space bar solves this. I haven't seen that with ArtPaint yet while panning madly for a few minutes...

BTW, you may want to try to split the coding style stuff into its own commit in the future. Makes discerning actual code changes easier for the generations to come. I know that it's very tempting to make a tiny style fix while doing code changes. And always commiting every little thing and later squash stuff with "git -i rebase" can be a drag... :)

Anyway, looks good. Let's merge!

dsizzle:
Yay, I was hoping it was just on my end!

Wrt style fixes, I agree, that’s why I did #298 separately- but I should do more that way.

Brunobasta:
The middle mouse button is nice to have too for pan

dsizzle:
That already existed, I didn’t remove that functionality when I added this.

humdingerb:
Strangely, I never tried the middle button... It does indeed pan the image,but IMO it's a bit weird. You move the mouse to the right and the image moves to the left. Can this work just like the SPACE panning, or is this - to me strange - panning behaviour appreciated by anyone?

Brunobasta:
The image should move to the right and the image should move to the right, too.
Wonderbrush, Blender, and Photoshop and Illustrator it using it this way.

PR #302, opened 2022-10-14
Selection: don't clear selection if either all points are included or excluded

### LINKED ISSUE #300: Selection dismissed when start- and endpoints outside canvas.
Until there's a "Select all" menu item... :)

The way most would probably select the whole canvas is:
* if needed zoom out until you see the striped background all around the canvas.
* click-and-hold the mouse to somewhere top-left outside the canvas (X/Y: -20/-20)
* move the mouse to somewhere bottom-right outside the canvas (X/Y: width+20/height+20).
* release the mouse-button.

If you release the mouse-button when at least one coordinate is within the canvas, the selection is made. If both are outside, as described above, there's no selection.

dsizzle:
only clear selection if BOTH are true.

Fixes #300


humdingerb:
Thanks!

Issue #126, opened 2022-04-10
Support common modifier key functions


humdingerb:
There are some modifier key functions that are commonly used by graphics apps:

* +CTRL:
    * lock aspect ratio to create circles and squares (already does for rect/ellipses tools)
    * draw lines with a 0°/45°/90° angle (already does for straight line tool)
    * rotate by 0°/45°/90° angle
    * keep aspect ratio when resizing something
    * add to a selection
 * +SHIFT:
    * remove from a selection
* +ALT:
    * when creting an ellipse or rectangle or when selecting, take the point where the mouse button was pressed as center. (Some tools - e.g. Rectangle tool - already have that option as "Center to corner" instead of the usual "Corner to corner)
    * with mouse wheel-up: zoom in (formerly #96)
    * with mouse wheel-down: zoom out (formerly #96)

dsizzle:
I think these are good ideas. 

One question I have is, what is the behavior when a modifier key conflicts with the chosen option?  I assume the key modifier should override the option, but does the option change temporarily to reflect the key press?

Lt-Henry:
Issue #96 can be merged here. Don't you think?

humdingerb:
> One question I have is, what is the behavior when a modifier key conflicts with the chosen option? I assume the key modifier should override the option, but does the option change temporarily to reflect the key press?

IMO, once these modifier behaviours are implemented, the options should be removed from the tool setup menus. It's most consistent and the result will always be predictable. Otherwise ALT or SHIFT can do different things with different tools if their options are toggled differently. Chaos! :)

dsizzle:
ugh. These might be super easy to do (at least I tried ellipse) but it makes #32 MUCH worse when you switch between the "center" mode and "corner-to-corner" mode. gray lines everywhere! :(

also turns out Alt is some kind of 'eyedropper' shortcut that grabs the color under the cursor.  That would need to be disabled or remapped.

Brunobasta:
shift +i (to inverse the selection is a very handy shortcut)

Brunobasta:
I agree with hundingers modifier key functions

dsizzle:
these are all done now, yes?

humdingerb:
Tools wise, yes. Only modifier relates feature still missing seems to be the SHIFT snaps rotation to 22,5° angles for the rotation of canvas and layer.

PR #303, opened 2022-10-14
ImageView: use scroll wheel + alt key to zoom in and out

dsizzle:
So: this *should* be alt+mouse wheel to zoom in and out - for some reason it works with shift, control, etc., but not alt.

@humdingerb can you please try this PR to see if it works for you?  I wonder if this is another VirtualBox thing since every other modifier key works for this feature except Alt. :shrug:  if it doesn't work then I will need to try to figure out what's different with the alt key I guess.

Part of #126


humdingerb:
Good news: it does work on a native Haiku.:)

Bad news: I already have an enhancement request... :)
It'd be much nicer, if the "zoom-center" would become the mouse pointer, once the image fills the window. Similar to ShowImage.

humdingerb:
Oh, and we lost a feature on the way: The Haiku standard of scrolling vertically with the mouse wheel and horizontally with CTRL+mouse wheel went missing.

humdingerb:
Sorry for the piece meal reporting...
Another casualty: ALT+mouse-click should temporarily switch to the colour picker. (Which reminds me to file #304).

dsizzle:
I restore the standard scrolling, and I also put back the color picker shortcut - I had purposely removed that because I wasn't sure how useful it is.

Thanks for testing!   

I will see if I can do the zoom-to-cursor part...

humdingerb:
> I also put back the color picker shortcut - I had purposely removed the color picker because I wasn't sure how useful it is.

Totally useful. :)
It lets you pick a colour of the image practically while you're painting. Otherwise you'd have to switch tools back and forth.

One more thing: When you're zoomed out a bit, it takes quite a few mouse wheel turns to zoom back into the picture. Already, the zoom-level steps increase as you zoom in. That works fine, but I think the initial steps could be a bit higher.

dsizzle:
Ok, I added "zoom-to-cursor" and tweaked the scroll speed a lot and it seems to feel ok to me at low and high zoom levels. 

Now, regarding zoom-to-cursor... it only works when the canvas is zoomed in so that it's larger than the window. The reason for this is because ArtPaint keeps the canvas centered in the window unless it's scrolled, and since the canvas is smaller than the window there's no content to scroll:  

center-duck

Notice that the scrollbars are inactive because there's no content outside the window so you can't move the canvas at all.  

So a question for the future:  is there any use to be able to slide the canvas around in the view when you're zoomed out?   Or should the "background" be a large scrollable area much larger than the canvas?

humdingerb:
> Ok, I added "zoom-to-cursor" and tweaked the scroll speed a lot and it seems to feel ok to me at low and high zoom levels.

Much better. One idea, please dismiss if it's too much of a hassle: Could an additional SHIFT increase the zoom-step to maybe 2x or 4x (needs to me tested out)?
That is used in various places in Haiku, e.g. Tracker or StyledEdit to do pagewise scrolling with the mouse-wheel.

> Now, regarding zoom-to-cursor... it only works when the canvas is zoomed in so that it's larger than the window.

That's perfectly OK and makes sense.

> So a question for the future: is there any use to be able to slide the canvas around in the view when you're zoomed out?

IMO there isn't. Panning only make sense if you've zoomed in enough so that the image doesn't fit into the window.

dsizzle:
> Could an additional SHIFT increase the zoom-step to maybe 2x or 4x?

it *should* be trivial to just double the scroll wheel delta, as long as it doesn’t break the fancy math for the scroll speed.

dsizzle:
ok, added doubling of zoom when SHIFT is also held.

humdingerb:
Good! Though IMO the zoom steps are still too small, esp. if you have larger images and zoom in the 300+ % range. Can we try doubling the step size for normal and shift-zooming?
I don't think it's a problem if the 50-100% range gets a bit more rougher/faster zooming as a result.

dsizzle:
ok, doubled the step for both shift- and non-shift case...lmk how it feels!

humdingerb:
Very nice now, I think.
Let's merge!

PR #305, opened 2022-10-14
Changed/added string, updated en.catkeys

humdingerb:
* Added "Color picker window", it appears as window title in the Deskbar and should be translated, however short lived the pop-up window is.

* Updated help texts to fit better to the tools ones.


dsizzle:
Looks fine to me. I like having the tool/manipulator name in the help text.

humdingerb:
Me too.

PR #310, opened 2022-10-16
Select-all: add 'select all' function to select entire canvas

### LINKED ISSUE #154: Edit menu: Add "Select all" item
Add a "Select all" (ALT+A) menu item to the _Edit_ menu down by the other selection related commands.

dsizzle:
Here's the ALT+A to select the entire canvas.  Haven't decided what to do about the "select all non-transparent pixels" yet as that's an awfully long menu item. :smile: 

Fixes #154

### DIFF
--- a/artpaint/application/MessageConstants.h
+++ b/artpaint/application/MessageConstants.h
@@ -103,6 +103,7 @@
 // these constants are used in the Edit-menu
 #define	HS_INVERT_SELECTION		'InSl'
 #define	HS_CLEAR_SELECTION		'ClSl'
+#define HS_SELECT_ALL			'SlAl'
 #define HS_GROW_SELECTION		'GrSl'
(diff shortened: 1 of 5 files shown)


humdingerb:
> Haven't decided what to do about the "select all non-transparent pixels" yet as that's an awfully long menu item.

I think keeping "Select all" as menu label is still OK. The help text in the status bar can elaborate, "Selects all non-transparent pixels of the _canvas_".

I just noticed that _this_ needs to be clarified: does it select all non-transparent pixels of the canvas, or the active layer? Makes more sense to limit it to the current layer, I think.
Otherwise we'd have to add a submenu, as it's done with Edit| Copy:

Select all
 > Active layer (ALT + A)
 > All layers (SHIFT + ALT + A)

Anyway, merging as is for now.

Issue #175, opened 2022-05-31
Ellipse tool: Add "Enable rotation" checkmark


humdingerb:
Similar to the feature of the "Rectangle tool", to rotate the ellipse interactively after it's drawn.

dsizzle:
In addition, I think both the Ellipse and Rectangle tools should have an angle slider/numeric field so you can set an explicit angle.

PR #316, opened 2022-10-18
Ellipse/Rectangle: Use modifiers for SHIFT and ALT instead of modes

dsizzle:
- SHIFT maintains "square" aspect ratio
- ALT uses center-to-corner mode
- removed radio buttons for modes; added to help string
- also draw with current fg color instead of just black

Part of #126

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -1419,7 +1419,7 @@ ImageView::PaintToolThread()
 
 	int32 tool_type = ToolManager::Instance().ReturnActiveToolType();
 
-	if (modifiers() & B_COMMAND_KEY)
+	if (modifiers() & B_CONTROL_KEY)
 		tool_type = COLOR_SELECTOR_TOOL;
 
 	if (tool_type != TEXT_TOOL) {
--- a/artpaint/tools/EllipseTool.cpp
+++ b/artpaint/tools/EllipseTool.cpp
@@ -73,15 +73,17 @@ EllipseTool::UseTool(ImageView *view, uint32 buttons, BPoint point, BPoint)
 		BPoint original_point;
 		BRect bitmap_rect,old_rect,new_rect;
 		window->Lock();
+		rgb_color old_color = view->HighColor();
 		old_mode = view->DrawingMode();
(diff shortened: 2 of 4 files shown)


humdingerb:
Nice change!
However, we need to use "CTRL for centered" as ALT is already taken by the instant-colour-picker.

dsizzle:
Personally I’d rather keep ALT for this and change the color picker, since I believe other tools use ALT for centering. What do you think?

humdingerb:
Fine by me.

dsizzle:
ok, changed color selector to CTRL. 

also added SHIFT to lock rotation of rectangle to 0/45/90 degrees.

artpaint/paintwindow/ImageView.cpp
```diff
@@ -1419,7 +1419,7 @@ ImageView::PaintToolThread()
 
 	int32 tool_type = ToolManager::Instance().ReturnActiveToolType();
 
-	if (modifiers() & B_COMMAND_KEY)
+	if (modifiers() & B_CONTROL_KEY) {
```
humdingerb, line 1422: This stray curly foils the build. :)

humdingerb:
Aside from the little build spoiler, it works as advertised. Nice!
To be consistent over tools, see #175. :)

artpaint/paintwindow/ImageView.cpp
dsizzle: dang. that was a merge conflict...I swear I don't usually push changes that don't build (aside from the 32-bit casts)

dsizzle: fixed, and my face is red.

dsizzle:
re: #175 as I mentioned the Ellipse code needs to be re-done.  It doesn't know how to render rotated ellipses.  don't worry, it's on my to-do list...

humdingerb:
Well, it is early morning... over here at least. :P
Merge!

humdingerb:
> also added SHIFT to lock rotation of rectangle to 0/45/90 degrees.

I wonder if this is too restrictive. Have you tried 45°/4 = 11,25° or - if that turns out too be to fine - 45°/2 = 22,5°?

PR #319, opened 2022-10-19
ColorSelectorTool: check if value is in range of sqrt table to avoid crashes

dsizzle:
ran across a random crash where the value was out of bounds for the array.


humdingerb:
:+1:

Issue #309, opened 2022-10-16
Selection.cpp: SimplifySelection() needs to be improved


dsizzle:
Selection::SimplifySelection() is a method that deconstructs a pixel selection into polygons but currently it doesn't do a great job.  Even for a simple rectangular selection, it generates a huge number of points. The lowest I saw in quick testing was 500 and I've seen multiple thousands.  A rectangular selection should ideally be a 4-point polygon! 

Look into raster->vector conversion algorithms, or add a way to deduplicate/reduce (Douglas-Peucker?) the points in selection.

humdingerb:
This is probably off-topic for this ticket, but would it be at all possible to keep selections as vector? That is, always have handles around the rect/ellipse like in the crop manipulator. Then you could easily adjust a selection pixel-precise, resize, rotate.

I suppose it may not be possible for the other selection tools. Or maybe a sort of bounding box could work.

dsizzle:
Yes, I don’t think this is a good idea overall. However, I do think we need the ability to transform selections. #156 is a prerequisite however.

dsizzle:
This might actually be the cause of #286.

dsizzle:
I’m working on “Moore’s Neighborhood” contour tracing - seems to be well known and at least my first attempt feels a lot easier to understand vs. the existing method. Not sure about performance.

humdingerb:
> I’m working on “Moore’s Neighborhood” 

Careful. Make sure to stay off Moore's Law(n)...
/me ducks

dsizzle:
After a bunch of work I decided SimplifySelection actually works really well.  I refactored a bit in #366 but still think there should be a way to reduce the number of points generated - but I don't think it needs a different algorithm now that #286 is solved.

dsizzle:
I have working code that does this simplification, but I can't decide if it's actually worth anything.  Whenever a pixel is tested for inclusion in a selection, we just check the selection map pixels directly, which is likely always faster than a polygon check.  Therefore I don't know if having a minimal number of points (like 4 points for a rectangular selection) really matters at all. It may provide some performance benefits in some operations but might not be worth it overall... still thinking about this one...

humdingerb:
Can't help you there...

dsizzle:
Ok, I have been looking into this because I think this same code can be used to generate UI elements for any current and future brushes (i.e. a representation of the brush when painting) to solve #128.  I originally thought the current code was good enough as stated above

[...]

PR #320, opened 2022-10-20
SelectorTool: code cleanup and refactor of Check*Spans code.

dsizzle:
massively cleaned up and reduced the span-checking code...probably will do that for FillTool as well as I think it's almost the same code...

### DIFF
--- a/artpaint/tools/SelectorTool.cpp
+++ b/artpaint/tools/SelectorTool.cpp
@@ -12,6 +12,7 @@
 
 #include "SelectorTool.h"
 
+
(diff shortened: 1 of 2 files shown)


humdingerb:
Builds and still works. Merge!

dsizzle:
I could be wrong but I *think* this could be related to #309. A new selection probably has a lot fewer points to cover. 

Or else it could be that a subtraction in your example is like doing a select-all behind the scenes and then it has to walk over the entire image? 

Agreed it could use optimization though.

PR #326, opened 2022-10-21
SelectorTool: add SHIFT and ALT modifiers for rectangle and ellipse

### LINKED ISSUE #318: Add SHIFT/CTRL/OPT modifiers to selection tools
The rectangular and ellipse selection tools could use the same modifiers as the Rectangle and Ellipse drawing tools:
* SHIFT makes square/circle
* CTRL for centered

Additionally, OPT could do "Substract area", making "Add area" the default, so the "Behavior" radio buttons can be removed. Possibly holding OPT would change the cursor to include a "-".

dsizzle:
- SHIFT maintains square ratio
- ALT centers the square or ellipse on the start point
- also combined rectangle/ellipse code since it was 95% the same
- OPTION key is subtract mode; removed add/subtract options

Fixes #318


dsizzle:
because of the way the tools work, changing the cursor to + and - is non-trivial.  I might do some work on the cursors in general as I don't think they're easy to see/read in a lot of cases - but for this change I don't want to mess with that.

humdingerb:
Something weird is going on:
* Set zoom to 100%
* Make e.g. a rect selection
* Zoom in to 150%
* Add another rect selection
--> The seclection frame is opening quite a bit below the mouse pointer.

Otherwise, I very much approve of this change. A few comments:

* Now that there's no add/substract behaviour to choose in the window any more, the "Mode" BBox label could be removed.
* Same with the "Wand tolerance" which could move to it's proper place under the "Magic wand" radio button (same as the "Undo steps" in the app settings window).
* The "OPTION" in the help text could be abbreviated as "OPT". Saves some pixels and people should be familiar with it from shortcut depictions in menus and the Haiku user guide.

humdingerb:
One more:
* The "_SHIFT maintains aspect_" in the help text is not quite right, as it works now. Currently, it makes square/circular.
    If you can make it work as you described - i.e. you draw out a rectangle/ellipse, then hold SHIFT and the current aspect ratio is held - that would be even better. To get a square/circle you'd have to first hold SHIFT, then start drawing out the selection rect/ellipse.
    Otherwise the help text could change to something similar to the rect/ellipse tool: "_SHIFT for square/circle_".

dsizzle:
I’m thinking I should add a width/height field for rect/ellipse and then actually make it aspect instead of square…

humdingerb:
That would be an additional option that would be independent from the SHIFT-aspect, I think.
If I understand yoyu correctly, that would mean a new checkbox "Lock aspect [____] : [_____]" under the rectangle and ellipse radio buttons. (Or not even a checkmark, but with default values "0 : 0" meaning disabled.)
If the user enters values, one click into the canvas would produce a selection of that size in pixels. When dragging the mouse the selection would grow, keeping that aspect ratio.

Good future feature, but doesn't solve what happens if the user holds SHIFT. :)

dsizzle:
No, I just mean adding “width” and “height” fields, like crop, but yes that’s the idea.

Without this, SHIFT only makes sense for circles and squares. As an aspect lock it’d be difficult to get a perfect circle or square without fields to be able to type the same value in both.

humdingerb:
You can already create a circle/square if you start holding SHIFT before clicking. So it would be possible to have SHIFT lock aspect more generally. If you draw out your rectangle before holding SHIFT, SHIFT will lock the current aspect ratio.

Having aspect ration textcontrols that double as width/height in pixels, as I described above, has its charms.
I suppose if that works for selections, it could be added to the rect/ellipse tool as well.

dsizzle:
Yes but if you lose that square aspect by lifting SHIFT it’d be impossible to regain.

humdingerb:
Well, just don't lift that SHIFT before releasing the mouse-button. :)

dsizzle:
I made the decision to keep it as square/circle for now, and will think about the aspect part more. 

The cursor works too!

Still looking into the weird bug with the offset rectangle.  Very odd.

dsizzle:
ok, I think I fixed the rectangle-zoom-bug, so I think this is good to go

or at least good to test. :laughing:

dsizzle:
btw I think this one is ready too, unless there are other bugs.

humdingerb:
Works nicely! However, there are some artifacts when dragging the mouse around to make the rect selection frame bigger and smaller. The rectangle tool sows the same.
New issue?

dsizzle:
Is it only when you use the modifier keys? I should probably fix it with this PR since it’s the most recent change and the likely culprit.

humdingerb:
No artifacts at <=100% zoom.
No artifacts at 200%, 300%, 400% etc. zoom. Artifacts appear with the zoom level between...
Best seen with center mode (ALT), but at least one line (top or left) also artifacts with no or the other modifiers.

humdingerb:
Oh, and current master also shows those artifacts... so no blame onthis PR.

dsizzle:
I *think* I fixed that issue, and yes, it was completely unrelated to this PR.

humdingerb:
Fixed! \o/
Let's quickly merge before something else pops up...

PR #327, opened 2022-10-21
Snapping: factored out snap code to UtilityClasses.

dsizzle:
- also adjusted Rectangle angle snapping to 22.5 instead of 45 degrees per comment in #316 
- changed BrushEditor to use new snap code too


dsizzle:
Also please don’t merge this. I want to move the code that normalizes the angle to -90-90 range into the snap function

humdingerb:
Noice. While you're at it, the Straight Line tool should also use this instead of the 45° snapping it does now.

humdingerb:
... oh and the Rotate... in the Canvas menu doesn't snap yet.

dsizzle:
Ok, I made the code change I wanted to make, and I fixed the Straight Line tool.

The RotationManipulator is different and I will file a ticket for that separately.

humdingerb:
Very good!

PR #333, opened 2022-10-23
Selection: code cleanup only, no functionality changed

dsizzle:
- but changed "Dilatate" to "Dilate" because it bothered me


humdingerb:
One dilettanted mistake less.

PR #335, opened 2022-10-24
Minor string changes

humdingerb:
* Always bothered me: "Clear layer" and "Clear canvas" removes all pixels, "Clear selection" only unselects all. Make it clear: "Select none" as the opposite of "Select all"

* Re-arrange the "Selection" menu items.

* Shorten "Clear layer" and "Clear canvas" to simply "Clear" as they are already in the appropriately named menus.

* Fix menu item names in comments refering to "Edit" that were moved to "Selection".

* Update en.catkeys.

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -612,7 +612,7 @@ ImageView::MessageReceived(BMessage* message)
 			}
 		} break;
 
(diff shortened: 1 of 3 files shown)


dsizzle:
One question: should the help text for “select all” be “Selects entire layer” instead of “Selects entire canvas”?  It doesn’t select all layers.

humdingerb:
You're right. PR updated.

dsizzle:
looks good to me!

PR #345, opened 2022-10-25
Selection: add option to hide selection borders.

### LINKED ISSUE #284: Selection menu: Add "Hide border" item
It's be nice to be able to hide the selection border "ants".
I suppose it would be a checkmarkable item "Hide border".
To acknowledge peoples' Photoshop muscle-memory, I propose to have the shortcut ALT+H.

dsizzle:
Fixes #284

### DIFF
--- a/artpaint/application/MessageConstants.h
+++ b/artpaint/application/MessageConstants.h
@@ -100,12 +100,13 @@
 #define	HS_SHOW_DATATYPE_SETTINGS	'SdtS'
 
 
-// these constants are used in the Edit-menu
-#define	HS_INVERT_SELECTION		'InSl'
-#define	HS_CLEAR_SELECTION		'ClSl'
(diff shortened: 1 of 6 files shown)


humdingerb:
Changing menu items' labels isn't so nice.  It's better to use a checkmarked menu item and keep the label constant. It's even in the under-maintained HIG...
One advantage is that you always see at a glance if the bottom menu item has a checkmark or not, without having to read any text.

humdingerb:
Also, what do you think about adding a separator above the "Hide borders"?
There are quite a few items in the Selection menu already, so it's nice to have something the eye can skip to. Also, it would separate the items changing the selection in any way and "non-destructive" actions.

dsizzle:
Excellent point about the checkmark instead of the label - fixed that! 

Also added the separator.

humdingerb:
Perfect.

PR #349, opened 2022-10-26
Inset "Tolerance" slider to be under "Magic wand"

humdingerb:
Otherwise the tolerance slider could be mistaken to be applied to other selection tools.

### DIFF
--- a/artpaint/tools/SelectorTool.cpp
+++ b/artpaint/tools/SelectorTool.cpp
@@ -629,6 +629,7 @@ SelectorToolConfigView::SelectorToolConfigView(DrawingTool* tool)
 		fTolerance =
 			new NumberSliderControl(B_TRANSLATE("Tolerance:"),
 				"10", message, 0, 100, false);
+		fTolerance->SetEnabled(FALSE);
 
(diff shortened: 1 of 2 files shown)


humdingerb:
I didn't find a way to disable the "Tolerance" slider when "Magic wand" isn't the active tool. It seems the actually selected tool is only checked when its been used. Of course then it's too late to dis/enable the tolerance slider...
No biggie, I suppose.

dsizzle:
If you look at StraightLineTool.cpp, in the StraightLineToolConfigView::MessageReceived function, you can see how the Width slider gets disabled when the checkbox is selected - I expect the same thing should work for disabling the Tolerance slider here.

humdingerb:
That appeared to be pretty straight forward. I just more or less copied the MessageReceived().
PR updated.
Unfortunately, it crashes when slecting the Selection tool: 
[ArtPaint-3439-debug-28-10-2022-06-40-30.report.txt](https://github.com/HaikuArchives/ArtPaint/files/9885898/ArtPaint-3439-debug-28-10-2022-06-40-30.report.txt)

Any ideas?

Issue #286, opened 2022-10-01
Freehand selections aren't smooth


dsizzle:
Freehand selections are not very nice with a lot of uneven edges:

![0hairy](https://user-images.githubusercontent.com/10878750/193422666-1e93701a-1755-466d-86c1-2db9273d5607.png)

They should be smooth (along pixel boundaries at least)

Brunobasta:
they should become a blur edge or a smooth edge feature too (feather edge?)

PR #350, opened 2022-10-28
Selection: add option to select all non-transparent pixels

dsizzle:
Added option to select all non-transparent pixels in the current layer.  I called it "All pixels" in the menu but open to suggestions!

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -135,6 +135,35 @@ BitmapUtilities::ConvertColorSpace(BBitmap *inBitmap, color_space wantSpace)
 }
 
 
+BBitmap*
+BitmapUtilities::ConvertToMask(BBitmap *inBitmap, uint8 color)
+{
+	BBitmap* out_map = new BBitmap(inBitmap->Bounds(), B_GRAY8);
+	uint8* out_bits = (uint8*)out_map->Bits();
+	int32 out_bpr = out_map->BytesPerRow();
+
+	uint32 *in_bits = (uint32*)inBitmap->Bits();
+	int32 in_bpr = inBitmap->BytesPerRow()/4;
+
+	union color_conversion c;
+
+	out_map->LockBits();
+	uint32 pos = 0;
+	for (int32 y = 0; y < out_map->Bounds().IntegerHeight(); y++) {
+		for (int32 x = 0; x < out_map->Bounds().IntegerWidth(); x++) {
+			c.word = *(in_bits + x + y * in_bpr);
+			float alpha = (float)c.bytes[3] / 255.;
+
+			*(out_bits + x + y * out_bpr) = (uint8)((float)color * alpha);
+		}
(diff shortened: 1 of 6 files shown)


humdingerb:
That is supposed to select all pixels that are fully, 100% non-transparent, right?
Since pixels could have an alpha of <255 and are still visible as pixels, "All non-transparent" or "All opaque" would be better.

There's a strange effect when using the Airbrush tool with a half transparent colour and selecting "All pixels":

dsizzle:
Actually I am trying to preserve the alpha and select any pixel that is not *fully* transparent. The idea is that if you have something with an antialised edge it’d be nice to select those pixels if you transform the selection instead of leaving behind some dots.

For some reason GitHub isn’t showing me your screenshot but I will experiment with some transparent stuff when I have a chance.  I didn’t try an airbrush stroke, but it works great on Ducky and Walter. :D

humdingerb:
> Actually I am trying to preserve the alpha and select any pixel that is not fully transparent. The idea is that if you have something with an antialised edge it’d be nice to select those pixels if you transform the selection instead of leaving behind some dots.

Right, that makesmore sense. Although, since that only works when the object with the antialiased edge is on an otherwise transparent layer, you could just as well just "Select all"...
I suppose it'll really become usefull when ArtPaint gets something like "Stroke/Fill selection".
"All non-transparent" seems to be an OK label. Or "All visible"? - that may be misunderstood though, e.g. all in the currently visible main-window-clipping of the canvas.

> For some reason GitHub isn’t showing me your screenshot

Nor me... this was the shot:
Airbrush

> but it works great on Ducky and Walter. :D

Well... it does look a bit frayed (this is pasting the "All pixel"ed Walter as a new layer on top of some image), or is that related to #286 :

all-pixels-pasted

dsizzle:
That airbrush selection screenshot is terrible, and is related to #309 …

I don’t think the Walter edges are the same issue as #286 - looks like the antialiasing didn’t all get selected?  Hmm.

dsizzle:
It might actually be the same issue as #286?  

Also - this is exposing other issues...I used the "erase" tool and the "select non-transparent" is selecting the part I erased! 
0ghost

humdingerb:
Spooky! Trick or treat, ArtPaint?!

dsizzle:
so actually fixing #286 *did* fix this too.  Works way better on Walter and paintbrush strokes.  Also airbrush strokes.

Although it does expose the issue that apparently the brushes and eraser are painting almost-but-not-quite transparent pixels?

humdingerb:
Nice! Merge now?

> Although it does expose the issue that apparently the brushes and eraser are painting almost-but-not-quite transparent pixels?

Looks like it, "Clear" canvas/layer also create non-transparent pixels. The colour picker shows an alpha of 1 instead of 0.

dsizzle:
Yes, I think it works well enough to merge if you haven’t found any other concerns

humdingerb:
It's good. Merge!

PR #369, opened 2022-11-03
ImageView: fix undo message for 'select all non-transparent'

dsizzle:
Apparently the undo message for 'All non-transparent' said 'Grow selection' ... copy/paste error, oops.


humdingerb:
Good catch!

PR #370, opened 2022-11-03
Selection: draw selection_map on view before adding polygons

### LINKED ISSUE #367: Freehand selections don't work properly with certain others
Steps to reproduce:

1. Use the magic wand and click on an empty canvas to select the entire thing
2. Draw a selection with the freehand tool while holding OPT

The selection subtracts the entire bounding-box of the freehand selection, not the selection itself:
0bad

Expected result:\
0good

It works fine if you subtract from a rectangle selection or select-all, but for magic wand and elliptical selections, trying to use a freehand selection on top behaves badly.

It also is problematic for additive selections:
0bad2

But again, it's ONLY on top of magic wand or elliptical selections; rectangles work fine with freehand on top, both additive and subtractive: 
0good2

dsizzle:
Ok, this was the correct fix after all, but I deleted some lines that I shouldn't have. 

Fixes #367

### DIFF
--- a/artpaint/application/Selection.cpp
+++ b/artpaint/application/Selection.cpp
@@ -200,6 +200,7 @@ Selection::AddSelection(HSPolygon* poly, bool add_to_selection)
 		selection_view->SetLowColor(255, 255, 255, 255);
 		selection_view->SetHighColor(0, 0, 0, 0);
 
+		selection_view->DrawBitmap(selection_map);
 		BPolygon* p = poly->GetBPolygon();
 		if (!add_to_selection) {
 			if (bound_poly) {
@@ -899,8 +900,6 @@ Selection::SimplifySelection()
 						}
 
 						next_point = BPoint(new_x, new_y);
-						if (bounds.Contains(next_point) == FALSE)
-							direction = -1;
 					}
 
 					HSPolygon* new_polygon;
--- a/artpaint/application/Selection.h
+++ b/artpaint/application/Selection.h
@@ -192,7 +192,8 @@ Selection::ContainsPoint(int32 x, int32 y)
 		return false;
 
 	return (selection_bits == NULL ||
-		((*(selection_bits + y * selection_bpr + x)) != 0x00));
+		(image_bounds.Contains(BPoint(x, y)) &&
+		((*(selection_bits + y * selection_bpr + x)) != 0x00)));
 }
 
 


humdingerb:
After using the magic wand on an empty layer so that all is selected, I get high CPU usage on one core. Some thread in app_server pegs. It's OK when using the magic wand on something in a non-empty image.

dsizzle:
Can you please try merging #371 before testing this change?  I think it may solve it.

humdingerb:
I did that now, it now pegs the ImageView thread. Unhelpful debug report:
[ArtPaint_debug-3507-debug-04-11-2022-15-52-16.report.txt](https://github.com/HaikuArchives/ArtPaint/files/9940035/ArtPaint_debug-3507-debug-04-11-2022-15-52-16.report.txt)

dsizzle:
Might not be unhelpful - looks like maybe lock contention on the selection bitmap or similar. I’ll check it out. I was hoping that bounds check would fix it bc I was seeing freezes before that fix but not after. :(

humdingerb:
I have complete confidence in you. Weekend's coming up and the weather's getting seasonally worse. :)

humdingerb:
Oh, should I merge #371 anyway, to have that one out of the way?

dsizzle:
Yes please, I don’t think there is any downside to #371 

well - possibly that it masks/avoids the bigger issue with off-by-one sizes. I see a single column of pixels on the right edge of the image trying to get selected (which is what causes the out-of-bounds) problems… but I don’t want to tackle that issue just yet. 

So I said alll that to say yes please merge #371 but this note is that there is still a bigger problem to solve in ArtPaint overall

humdingerb:
Hmm... apparently I already merged #371. All these PRs you fire left, right and centre - I'm dizzy...

> ... but this note is that there is still **a** bigger problem to solve in ArtPaint overall

Oh well, if ArtPaint has just that _one_ problem overall... )

dsizzle:
Well, I wasn’t feeling well yesterday so I was off work - in between naps I worked on ArtPaint, perfect medicine? 🤣 

> Oh well, if ArtPaint has just that one problem overall... 

exactly why I need to fire off so many PRs! We are like a decade behind Adobe and need to catch up!

dsizzle:
whoa yikes, the magic wand selected every pixel as a separate polygon or something...

humdingerb:
That's not a performance miracle. :)

Totally off-topic for this PR, are you familiar with the great Captain Disillusion?
He's more a video effects guy, but I recently saw this episode which - if you can follow it (I can't really, maybe need to rewatch a few times...) - may be of interest to ArtPaint's alpha handling:
https://www.youtube.com/watch?v=XobSAXZaKJ8

dsizzle:
Yeah #371 is problematic. But I still think it needs to happen, just differently. Like you said, may be a rainy weekend…

I hadn’t heard of CD but it’s a fun video. I knew all that already but changing from straight to premultiplied alpha isn’t something I want to tackle at the moment. Maybe someday.

dsizzle:
ok, I fixed it.  I need to be more careful; I missed a bounds check when I did the refactor in #366.  Now it seems to work as it should.

humdingerb:
I confirm, works nicely, and feels much quicker!

PR #372, opened 2022-11-03
PixelOperations: if resulting alpha is 0, just return 0 for pixel value

dsizzle:
- was changing 0 alpha to 1 to avoid div-by-zero errors, but the effect was returning an alpha of 1, so it was not really transparent (but 1/255 is pretty darn invisible)!


dsizzle:
this fixes the "ghost erasing" btw

humdingerb:
Helloween's officially over.   >:E

PR #373, opened 2022-11-04
LayerWindow: add message filter for B_KEY_DOWN to layer window

### LINKED ISSUE #296: Layers: Add BMessageFilter AppKeyFilterFunction
To react to ALT+TAB to un/hide all panel windows, I think the BMessageFilter to call AppKeyFilterFunction() should be added in the Layers window.
See e.g. [ToolSetupWindow.cpp#L48](https://github.com/HaikuArchives/ArtPaint/blob/master/artpaint/tools/ToolSetupWindow.cpp#L48)

dsizzle:
Fixes #296


humdingerb:
The Layers window finally plays nice with hide&seek. Thanks!

PR #374, opened 2022-11-04
ImageView: clearing layer, selection, or canvas should clear it

### LINKED ISSUE #354: Clear layer/canvas should make transparent (=remove pixels)
Currently, invoking "Clear" from Layer or Canvas menu fills the layer/selection or canvas with the background colour. In most situations - pretty much all but if you're on the bottom layer - it's not helpful at all. And given the name "Clear", it's downright counter-intuitive.

I think it makes sense to always actually remove the pixels of the selection/layer/canvas.
People can always create a bottom-layer and fill it with their background colour.

Related: Does the "Background color" option of the Eraser tool still make sense?
Given the above, it'd more consistent to also always erase to transparent and forget about "background colour".

I think the whole "background colour" things cater to a situation where there is no transparency and layers. We do have those, though. Things sure were confusing, before you implemented the checkered-transparency indication...

dsizzle:
- moved Layer->Clear to Edit->Delete as it's more accurate
- this also fixes Edit->Cut

I tried to use a shortcut but all menu shortcuts HAVE to start with ALT, and also with B_DELETE and B_BACKSPACE, nothing shows up in the menu shortcut, so it looks weird to just show "ALT" when the shortcut is "ALT+DEL"  - so I left it with no shortcut for now.

Fixes #354

### DIFF
--- a/artpaint/application/MessageConstants.h
+++ b/artpaint/application/MessageConstants.h
@@ -109,10 +109,8 @@
 #define HS_HIDE_SELECTION_BORDERS	'HbSL'
 #define HS_SELECT_LAYER_PIXELS		'SlPx'
 
-// These constants are used to inform about clearing of
-// the canvas or a individual layer.
-#define	HS_CLEAR_CANVAS			'Clcv'
-#define HS_CLEAR_LAYER			'Clla'
+// These constants are used to inform about deleting a layer or selection
+#define HS_EDIT_DELETE			'Clla'
 
 // This constant is used in a message that contains
 // an int32 'mode' that describes the new mode for ImageView.
--- a/artpaint/paintwindow/Image.cpp
+++ b/artpaint/paintwindow/Image.cpp
@@ -493,7 +493,7 @@ Image::ClearCurrentLayer(rgb_color &c)
 	cleared_layer->Clear(c);
 
 	// Store the undo.
-	UndoEvent *new_event = undo_queue->AddUndoEvent(B_TRANSLATE("Clear layer"),ReturnThumbnailImage());
+	UndoEvent *new_event = undo_queue->AddUndoEvent(B_TRANSLATE("Delete"),ReturnThumbnailImage());
 	if (new_event != NULL) {
 		for (int32 i=0;i<layer_list->CountItems();i++) {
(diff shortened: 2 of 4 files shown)


humdingerb:
Good change, one thing though: 
Currently, if there's a selection, only the selected area is cleared when calling "Canvas | Clear". I'm not sure if it's a bug or a feature. One might expect a "Clear canvas" always empties the whole canvas, all layers. Now, with a selection, you practically punch a hole through all layers.

I'm not sure... what do you think?

WRT ALT-less menu shortcuts, see the only 12 year old ticket https://dev.haiku-os.org/ticket/7078 :)

You could use a BMessageFilter to properly react on a DEL key press to invoke "Edit | Delete", even if there won't a shortcut icon. IMO, having DEL as the typical "delete the selected object" is extremely useful for working quickly. Nobody ever uses the mouse to find and click "Edit | Delete" if there's a keyboard shortcut.

dsizzle:
Honestly I considered removing the Canvas | Clear as I think it’s just as easy and more likely to just open a new project. I’ve never wanted to clear a whole image in other tools. Also as it currently stands it doesn’t delete all the layers, does it?

Cutting a hole through all layers is interesting but I can’t think of a real use case for it.

humdingerb:
It does clear all layers, provided there's no selection. And yeah, who wants to punch holes in layers?
I'd say good riddens. Every saved menu item is a win.

dsizzle:
What I actually meant was whether Canvas | Clear deleted all the layers - so if you have 37 layers and you choose Canvas | Clear, it clears all 37 but then you have 37 blank layers, which is terrible. 

So, it's now removed. One fewer menu item to worry about.

humdingerb:
Very good. 2 things, though:
* As the new command is a generic "Delete", the undo/redo text should also be "Delete". There's still "Clear canvas" there.
* The constant HS_CLEAR_LAYER could also be renamed "HS_DELETE" or maybe "HS_EDIT_DELETE".

dsizzle:
Fine, fine, I fixed those, and also removed the functions that clear the canvas

humdingerb:
Merge!

PR #375, opened 2022-11-04
PaintWindow: fix bugs in resize-to-fit, and remove menu item

### LINKED ISSUE #322: Window menu: "Resize to fit" bugs/enhancements
The menu item "_Window | Resize to fit_"  is supposed to resize the main window so that it shows all of the canvas (or at least the whole width) at the current zoom level.

* If the image (at the current zoom level) is wider than the screen resolution, the window is moved into the top left corner of the screen and is resized to fill the whole screen. **OK.**
* If the image (at the current zoom level) is _not_ wider than the screen resolution, the window is only moved vertically to the top of the screen, it stays at its X-coordinates. The window gets resized to fit the image width, even if that extends it outside the screen. **Not OK.**
* Zoom out until you see the whole canvas in a smallish main window. "Resize to fit" does not make the window large enough to disable the scroll bars. It's a few pixels to narrow (the amount of the window borders?), and many pixels too short (the amount of the status bar?).
* As  "Resize to fit" does the same as the window's Zoom button, it could also share its shortcut CTRL+ALT+Z. Saves us the ALT+Y for something else. Come to think of it, we could even remove the menu item  "Resize to fit"...

dsizzle:
Fixed all the mentioned issues as far as I can tell - it resizes to fit with full scrollbars, and doesn't end up in situations where the window is partially off the screen - if that's going to happen it centers the window. 

Also removed the menu item because it really does behave exactly the same as the widget on the window tab, so no need to waste a menu item. 

Fixes #322

### DIFF
--- a/artpaint/paintwindow/PaintWindow.cpp
+++ b/artpaint/paintwindow/PaintWindow.cpp
@@ -542,13 +542,6 @@ PaintWindow::MessageReceived(BMessage *message)
 			be_app_messenger.SendMessage(message);
 		}	break;
 
-		case HS_RESIZE_WINDOW_TO_FIT: {
-			// this comes from fMenubar->"Window"->"Resize to fit" and
-			// informs us that we should fit the window to display exactly the
-			// image use a private function to do resizing
-			_ResizeToImage();
-		}	break;
-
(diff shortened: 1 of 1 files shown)


humdingerb:
Perfect!

PR #377, opened 2022-11-05
RotationManipulator: added SHIFT to snap to 22.5 degree angles

### LINKED ISSUE #126: Support common modifier key functions

### LINKED ISSUE #330: The Rotate manipulator should use SHIFT to snap to angles
...like the other operations, a user should be able to hold shift to rotate the canvas and selections to 22.5 or 45 degree angle increments.

dsizzle:
...also (by necessity) updated SnapToAngle() function to allow different max angles rather than just 90

Point of note, the angle values in the manipulator dialog might seem strange but they are strange even without snapping.  I don't know how much it matters as you can always type in a specific angle if you wish, and the visual is what's really most important.  

Fixes #330 

Fixes #126

### DIFF
--- a/artpaint/application/UtilityClasses.cpp
+++ b/artpaint/application/UtilityClasses.cpp
@@ -115,14 +115,14 @@ MakeRectFromPoints(const BPoint& point1, const BPoint& point2)
 
 
 float
-SnapToAngle(const float snap_angle, const float src_angle)
+SnapToAngle(const float snap_angle, const float src_angle, const float max_angle)
 {
 	float new_angle = src_angle;
 
-	if (src_angle > 90)
-		new_angle = -(90. - ((int32)src_angle % 90));
-	if (src_angle < -90)
-		new_angle = 90. - ((int32)-src_angle % 90);
+	if (src_angle > max_angle)
+		new_angle = -(max_angle - ((int32)src_angle % (int32)max_angle));
+	if (src_angle < -max_angle)
+		new_angle = max_angle - ((int32)-src_angle % (int32)max_angle);
 
 	float abs_src_angle = fabs(new_angle);
 	float abs_snap_angle = fabs(snap_angle);
@@ -132,7 +132,7 @@ SnapToAngle(const float snap_angle, const float src_angle)
 	if (new_angle < 0)
 		sign = -1;
 
-	for (float i = 0; i <= 90; i += abs_snap_angle) {
+	for (float i = 0; i <= max_angle; i += abs_snap_angle) {
 		if (abs_src_angle > i - half_angle && abs_src_angle < i + half_angle) {
 			new_angle = i;
 			return new_angle * sign;
--- a/artpaint/application/UtilityClasses.h
+++ b/artpaint/application/UtilityClasses.h
@@ -49,7 +49,8 @@ public:
 BRect FitRectToScreen(BRect source);
 BRect CenterRectOnScreen(BRect source);
 BRect MakeRectFromPoints(const BPoint& point1, const BPoint& point2);
-float SnapToAngle(const float snap_angle, const float src_angle);
+float SnapToAngle(const float snap_angle, const float src_angle,
+	const float max_angle = 90.);
 
 
 inline uint32
--- a/artpaint/viewmanipulators/RotationManipulator.cpp
+++ b/artpaint/viewmanipulators/RotationManipulator.cpp
@@ -15,6 +15,7 @@
 #include "PixelOperations.h"
 #include "RotationManipulator.h"
 #include "Selection.h"
+#include "UtilityClasses.h"
 
 
 #include <Catalog.h>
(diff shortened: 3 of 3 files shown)


humdingerb:
Some observations:
1. Holding SHIFT to snap to angles doesn't yet work smoothly. 
2. Rotating canvas not yet 100% OK when there's a selection:
* Create a project with 3 layers, e.g. by putting 3 different images on each layer
* Create a selection
* a) Canvas | Rotate + 90°  --> The whole canvas is rotated, instead of just the selection on all layers.
   b) Canvas | Rotate...  --> Here the selection on all layers are rotated, but only on the current layer the selected area is cut out and rotated. On the other layers a copy of the selected area is rotated.

dsizzle:
Should Canvas | Rotate respect a selection? Similar to punching a hole in all layers, would you ever actually want to translate, rotate, and/or scale a cutout across all layers? 

Also what do you mean about the snapping not working smoothly yet?

and while we are asking rotation related questions, do you think the rotation pivot should default to the center of the selection instead of wherever it defaults to now?

dsizzle:
Also what is the point of Canvas | Translate?

humdingerb:
> Should Canvas | Rotate respect a selection? Similar to punching a hole in all layers, would you ever actually want to translate, rotate, and/or scale a cutout across all layers?

For example, I could imagine having an object on one layer and a drop-shadow for it on te one below, and possibly other effects on another layer. For course, It'd be better if we could link layers and only move/rotate etc. those... :)

> Also what do you mean about the snapping not working smoothly yet?

It just doesn't really rotate sometimes, as if the mouse didn't touch the snapping-angle. Here's a video where I first rotate freely, then I kept holding SHIFT: http://0x0.st/oETz.mpg [8.7 MiB]

humdingerb:
> Also what is the point of Canvas | Translate?

I suppose you could have been working on something, created a bunch of layers in the course of it. Then you realize, you've been working only in a small area in the middle of the canvas, but need more space to the bottom right. Do Canvas|Translate and move it all to the top left corner and continue work in the now empty bottom right corner... I don't know...

BTW, I find "Move" instead of "Translate" a more down-to-earth term. Is that a non-native speaker opinion and graphic app users are totally used to "translate" obejcts?

humdingerb:
> do you think the rotation pivot should default to the center of the selection instead of wherever it defaults to now?

Yes. Center of selection would be nicer.

dsizzle:
> It just doesn't really rotate sometimes, as if the mouse didn't touch the snapping-angle. 

I see, thanks for the video. I guess I spent more time with selections vs. the entire canvas.

> I suppose you could have been working on something, created a bunch of layers in the course of it. Then you realize, you've been working only in a small area in the middle of the canvas, but need more space to the bottom right. Do Canvas|Translate and move it all to the top left corner and continue work in the now empty bottom right corner... I don't know...

Personally, in my experience, I'd just do either a crop outside the bounds (aka "crop" the image larger) or resize the canvas (which I realized ArtPaint doesn't have aside from Crop. Photoshop has "Image Size" which scales the image, or "Canvas Size" which resizes the canvas without scaling (adding blank space if needed).  You can do this with Crop, although I just realized the Crop manipulator doesn't draw on top of the background and probably should.

> BTW, I find "Move" instead of "Translate" a more down-to-earth term. Is that a non-native speaker opinion and graphic app users are totally used to "translate" obejcts?

I could be persuaded either way. "Translate" is more the mathematical term but I don't remember what most graphical apps use.  I seem to recall Photoshop calling it the "Move" tool but not sure what it's called in the "Transform" dialogs or menus.  Blender uses "Grab" which makes sense once you're used to it but at first is confusing.

dsizzle:
I created #378 for the rotation pivot and #379 for the inconsistent Canvas rotation with a selection, as those are 2 issues outside the realm of this ticket.  I think I need to do a bit of refactor for the way Manipulators deal with/are aware of selections, for one thing.  The 'rotate all layers' one might be tricky too.

dsizzle:
Please give this a try; it feels a lot better to me. When I was trying to reproduce what you showed in that video, I found that with a little practice I could move almost 360 degrees with no rotation/snap!  Yikes!  Thanks for your testing and critical views.

Now I can't reproduce that behavior so I am hoping it's fixed.

humdingerb:
WRT "Canvas | Translate": I suppose using "Crop" to move the crop-box without resizing it is an good way to achive the same thing. So I'm OK with removing the "Canvas | Translate".

WRT "Move" vs. "Translate", maybe we should keep "Translate" so we can keep ALT+T (ALT+M is already taken). Also, I hope we (read: "you") unify Transform and Rotate (and Scale) into one "Transform" action. Using one "transform-box" where depending where you grab the handles it moves/rotates/resizes (and possible also shear and free transform with some modifier keys).
Yes, yes, that is remarkable similar to Photoshop... :)

> I created https://github.com/HaikuArchives/ArtPaint/issues/378 for the rotation pivot and https://github.com/HaikuArchives/ArtPaint/issues/379 for the inconsistent Canvas rotation with a selection, as those are 2 issues outside the realm of this ticket.

OK. I cannot reproduce case b) of #379 with this updated PR. Is that possible?

Merging this one, as it seems to rotate and snap nicely. Thanks!

dsizzle:
> OK. I cannot reproduce case b) of https://github.com/HaikuArchives/ArtPaint/issues/379 with this updated PR. Is that possible?

Shouldn't be possible - this only adjusts the rotation angle, nothing to do with how the pixels are manipulated.

dsizzle:
> WRT "Move" vs. "Translate", maybe we should keep "Translate" so we can keep ALT+T (ALT+M is already taken). Also, I hope we (read: "you") unify Transform and Rotate (and Scale) into one "Transform" action. Using one "transform-box" where depending where you grab the handles it moves/rotates/resizes (and possible also shear and free transform with some modifier keys).
Yes, yes, that is remarkable similar to Photoshop... :)

baby steps - I spent some time trying to get Scale working for a selection and it's a bit harder than I hoped. There is some commented-out code for a "Free Transform" manipulator but I haven't looked at it.

PR #381, opened 2022-11-07
Manipulators: refactor so selection is a class member

dsizzle:
Refactored so that each GUI manipulator has a class member called "selection" that holds the current selection when the manipulator is instantiated. This is instead of passing in the selection only when the bitmap is actually manipulated, because the selection should never change during a manipulation. It allows other functions in the manipulator to access the selection, for example if we wanted to draw a transform box around the transform region, or set the rotation pivot to the center of the selection.

This is a relatively large code change but there should be no functionality changed - I tried to test the different manipulators but if you could test the different manipulators to see if they behave no worse than before I'd appreciate it.  

I tried to minimize stylistic changes.

### DIFF
--- a/addons/AddOns/AdaptiveHistogramEqualization/AHEManipulator.cpp
+++ b/addons/AddOns/AdaptiveHistogramEqualization/AHEManipulator.cpp
@@ -40,7 +40,8 @@ Manipulator* instantiate_add_on(BBitmap *bm,ManipulatorInformer *i)
 
 
 AHEManipulator::AHEManipulator(BBitmap*)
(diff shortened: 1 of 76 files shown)


dsizzle:
ugh, forgot that a bunch of AddOns are also the same manip type...

dsizzle:
Feel free to test this but I did find one odd behavior that I'm trying to track down, so please don't merge yet.

humdingerb:
I did some layer / canvas resizing, cropping, flipping, tried our tools with /without selections.
Looks like it still all behaves as before. Let's merge.

PR #386, opened 2022-11-10
RotationManipulator: when selection is set, set origin to center point

### LINKED ISSUE #378: Rotation pivot should center on selection by default
When using the Rotation tool the center point defaults to the center of the canvas/layer.  Instead, it should default to the center of the selection if there is one.

dsizzle:
Fixes #378


humdingerb:
Very good.
A related question: what do you think about no resetting the rotated area when the user sets a new rotation center? That way it'd be possible to first rotate around some point A and then change to rotate further around point B, etc. Currently the rotated selection always snaps back to the start position when setting a new center.

Issue #224, opened 2022-07-02
Add-ons: Apply only to current selection


humdingerb:
If there's a selection, apply the addon just to it, not the entire layer.

Most add-ons do that already, these don't:
* AHE
* Anti-dither
* Color reducer
* Color separator
* Marble

dsizzle:
Oh I fixed Marble with #381. I considered trying to do the others but decided it was a bad idea to lump even more changes in there.

humdingerb:
Alright! Marble: strike!

PR #388, opened 2022-11-10
Anti-Dither addon: only affect selected areas if there's a selection

dsizzle:
- use selection in Anti-Dither addon
- also refactored anti_dither function to remove nearly duplicate code

I couldn't help myself. 

Addresses #224

Still a few more addons to do for that ticket


humdingerb:
Works as advertised. Founda new issue while testing: #389.

Issue #155, opened 2022-05-03
Layer menu: move action to Edit menu & change behaviour


humdingerb:
The translation related actions are inconsistent:
* _Rotate_ and _Translate_ also work on selections
* _Flip horizontally/vertically_ only work on the whole layer

Suggestion:
Move all 4 actions to the _Edit_ menu and make them work on the current selection. If anyone needs to apply these actions to the whole layer, simple "Select all" first (#154).

dsizzle:
Agreed!

PeteCA:
On Tue, May 03, 2022 at 11:18:30AM -0700, humdinger wrote:
 * _Rotate_ and _Translate_ also work on selections
I never knew that!
> 
> Suggestion:
> Move all 4 actions to the _Edit_ menu and make them work on the current selection. If anyone needs to apply these actions to the whole layer, simple "Select all" first (#154).

Not sure I like that.  I've never used rotate or translate on a selection -- always generated
a layer to hold just what I wanted to move or translate, and worked on that.  Tends to
avoid the ugly blank areas that occur if you work on the selection itself.   

What I'd find *extremely* useful on occasion would be the ability to scale a single
layer rather  than the whole canvas.

humdingerb:
> > If anyone needs to apply these actions to the whole layer, simple "Select all" first ([#154](https://github.com/HaikuArchives/ArtPaint/issues/154)).

Also, having no selection could also rotate/translate/flip/size the whole layer. 

> Not sure I like that.  I've never used rotate or translate on a selection -- always generated
a layer to hold just what I wanted to move or translate, and worked on that.

Well, sometimes having to create a new layer isn't needed. Imagine two separate objects on a transparent or single-colour background. You could just make a selection on one of them and rotate/translate/flip/size it.
Implementing the suggestion of this ticket would still allow you to keep working the way you decribed, but is more flexible for some situations or workflows.

PR #391, opened 2022-11-11
FlipManipulator: make horiz and vertical flip respect selections

dsizzle:
This is part of #155.


dsizzle:
this has exposed something else wrong with ellipses...sometimes the flip axis is NOT the center of the selection but something...else.  Only saw the issue with ellipses.

humdingerb:
Indeed, the first flip of an ellipse selection moves it to the right (h-flip) or down (v-flip).
Separate issue, oryou wanna try fixing that in this PR?

dsizzle:
It’s separate. I filed #393 for it

humdingerb:
And we merge...

Issue #382, opened 2022-11-07
Selection oddness


dsizzle:
Steps to reproduce
1. Load image
2. Make an elliptical selection:

![0a](https://user-images.githubusercontent.com/10878750/200230505-94db05c1-6bf3-48c5-b8a7-6768f40af8d0.png)

3. Choose AddOn->Blur, but then Cancel the operation
4. Make a second elliptical selection
Expected result:
![0c](https://user-images.githubusercontent.com/10878750/200230518-ec282bcf-edb0-48a2-9e56-e5f3db838943.png)

Actual result appears to be an offset "all non-transparent" selection:
![0b](https://user-images.githubusercontent.com/10878750/200230528-3f89b241-ee08-45ba-b275-308d4aee5852.png)

Suspecting #370 perhaps

dsizzle:
I don't think it's #370 necessarily - might just be the elliptical selection itself.  Even easier to repro is to just do an ellipse selection, then change tools, then change back and do another ellipse.  Seems to only affect ellipses...

dsizzle:
ok, not sure if it's good or bad news, but I traced this all the way back to the first Ellipse selection checkin - so nothing new broke, but Ellipse selections appear to be corrupting memory or doing something to the selection map.

PR #394, opened 2022-11-12
Ellipse selection: fix alpha for background when adding bitmap selection

### LINKED ISSUE #393: Ellipse selections sometimes select 1 pixel edge on right/bottom of image
I saw this before but thought it was fixed in #382 - seems to be throwing off the bounding box calculation for flipping.

dsizzle:
Fixes #393 

This is even sillier than the one line change from #382...

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -109,8 +109,8 @@ BitmapUtilities::ConvertColorSpace(BBitmap *inBitmap, color_space wantSpace)
 
 				out_map->LockBits();
 				uint32 pos = 0;
-				for (int32 y = 0; y < out_map->Bounds().IntegerHeight(); y++) {
-					for (int32 x = 0; x < out_map->Bounds().IntegerWidth(); x++) {
+				for (int32 y = 0; y <= out_map->Bounds().IntegerHeight(); y++) {
+					for (int32 x = 0; x <= out_map->Bounds().IntegerWidth(); x++) {
 						c.word = *(in_bits + x + y * in_bpr);
 						float alpha = (float)c.bytes[3] / 255.;
 						float value = 0.21 * ((float)c.bytes[2] / 255.) +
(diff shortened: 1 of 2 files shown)


humdingerb:
I don't see how the alpha channel plays into it... but I trust you know what you're doing. :)

Unfortunately, while it appeared at first to have solved the offset-while flipping, it's still happening when playing around long enough. Upon further testing: It appears that it works for the first image you open after launch, but if you close it and open another (without quitting the app), the bug re-appears...(Found #395 while doing that, btw.].

dsizzle:
these ellipse selections are gonna be the death of me.  it doesn't make sense at all that it would happen upon opening a second window???  

but thanks, I will investigate more!!

humdingerb:
Simple solution, as ArtPaint is for artists, get rid of the ellipses and have them draw their perfect circles with the freehand tool. :D

Observation, you have probably made yourself already: The further the ellipse is from the center of the image, the further it gets displaced on the first flip.

dsizzle:
> Simple solution, as ArtPaint is for artists, get rid of the ellipses and have them draw their perfect circles with the freehand tool. :D

hahaha, good idea!

> Observation, you have probably made yourself already: The further the ellipse is from the center of the image, the further it gets displaced on the first flip.

right, this makes sense if the symptom is that there’s an errant 1 pixel selection along the right/bottom edge because the axis of flip is the center of the selection. So if you draw an ellipse touching the left side of the canvas, it oughta flip almost all the way over to right edge because the axis would practically be the center of the canvas

dsizzle:
I think I fixed it - I was able to reproduce it simply by drawing an ellipse in the center, and then one on the far left.  The bounding box for the selection was getting calculated incorrectly. I can't break it anymore.

humdingerb:
You have to be very string now...

I still see the same phenomenon:
* Launch ArtPaint
* Open image 1
* Select an ellipse and flip -> works
* Close image 1 (without quitting ArtPaint)
* Open image 2
* Select an ellipse and flip -> selected area gets displaced

(I deleted the objects folder and checked out PR 394 afresh before building.)

dsizzle:
interesting, I can reproduce this using the steps you posted.

dsizzle:
ok, now I'm pretty positive this is actually fixed.

humdingerb:
Confirmed fixed!
Amazing how one stupid little pixel can wreak havoc...

PR #396, opened 2022-11-17
ImageView: allow space panning even when manipulator/addon is active

### LINKED ISSUE #351: SPACE-panning when translating/rotating/cropping
It'd be nice if SPACE-panning worked while translating, rotating or cropping. Currently every click-drag in the canvas does the ranslating/rotating/cropping, even if you hold SPACE.

### LINKED ISSUE #389: No spacebar-panning when in transformer or add-on mode
Spacebar-panning doesn't work when there's a transformer (rotate/scale/crop) or add-on window open.
While it's probably acceptable for the transformers, when using an add-on you may want to zoom in to preview the effect up close and pan to different areas of the canvas.

dsizzle:
I tried moving the space-panning code to whether there's a manipulator/addon active or not.  It seems to work although it's not totally perfect - like if you are translating and you click INSIDE the selection, it doesn't do anything.  If you click OUTSIDE the selection with space down it works. 

Same with addons, some addons might change the center of the effect when you click to pan but that might be unavoidable.  Overall it seems to function - give it a try? 

Fixes #351 
Fixes #389

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -189,6 +189,8 @@ ImageView::AttachedToWindow()
 
 	// Make this view the focus-view (receives key-down events).
 	MakeFocus();
+
+	SetEventMask(B_KEYBOARD_EVENTS);
 }
 
 
@@ -295,7 +297,7 @@ ImageView::KeyDown(const char* bytes, int32 numBytes)
 		if (bounds.bottom + delta > bitmap_rect.bottom)
 			delta = bitmap_rect.bottom - bounds.bottom;
 		ScrollBy(0, delta);
-	} else if (*bytes == B_SPACE && fManipulator == NULL) {
+	} else if (*bytes == B_SPACE) {
 		if (space_down == FALSE) {
 			space_down = TRUE;
 			be_app->SetCursor(fGrabCursor);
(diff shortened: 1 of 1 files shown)


humdingerb:
> like if you are translating and you click INSIDE the selection, it doesn't do anything. If you click OUTSIDE the selection with space down it works.

It's working here, so it's either you running in a VM or this:
When the "Translate" window pops up, it steals the focus from the main window. Now, if you click-drag anywhere in the canvas - with or without SPACE - it moves the selection. If, however, you first give the main window the focus (e.g. by clicking the yellow tab), the SPACE bar is heeded and you pan.

This phenomenon isn't limited to these transformation windows, of course. Any focus change does it: Try changing tools and SPACE pan directly without giving the focus back to the main window first. You'll paint with the tool instead of SPACE-pan.

Can this be solved? Can every action check if SPACE is held before starting their action (paint/select/                       transform...)?

dsizzle:
not sure. the focus thing is tricky. I *think* the KeyDown has to happen when the window has focus.  And since SPACE isn't a modifier key, it requires a KeyDown event to activate it.  So the order is wrong. 

in other words, you must click to get focus, and the SPACE bar requires focus, so you cannot press SPACE until *after* you clicked to get focus.  pressing it beforehand sends the KeyDown(SPACE) to whatever *does* have focus.  

I'm not aware of any "IsKeyDown" functionality, only a "KeyDown" and "KeyUp" that fire when a key is pressed or released.  only the modifiers (ALT, CONTROL, OPTION, SHIFT) can be detected as down at any point in time. 

I think they only way to get that to work would be to disable B_ACCEPT_FIRST_CLICK which then requires a click-to-focus before the paint window would accept any clicks to take action - and I think that's an annoying experience, having to repeatedly click on the paint window every time something else takes the focus.

humdingerb:
Hmm. Have a look at this KeyCatcher BView I use in Clipdinger: https://github.com/humdingerb/clipdinger/blob/master/src/KeyCatcher.cpp

I use SetEventMask(B_KEYBOARD_EVENTS) to get all key events, even if the window doesn't have focus. If SHIFT + ALT + v is pressed, Clipdinger jumps into action.
Maybe ArtPaint could do something similar to intercept B_SPACE and set a flag that all tools can check or something.

dsizzle:
I tried adding the SetEventMask() to ImageView and it *seems* to work (but of course I have the weird VM key interactions) - let me know if it's better.  Thanks for the tip btw!

humdingerb:
Works nicely! :+1:

PR #398, opened 2022-11-18
TranslationModifier: adjust "clear" rect to be larger to avoid artifacts

### LINKED ISSUE #24: Canvas > Translate feature skews/skids drawing

dsizzle:
Another try at this pesky ticket. 

Fixes #24 

Never give up hope!


humdingerb:
Very much better! Cannot reproduce any artifacts when moving selections.
Almost...
Only when the whole image is selected, and only with very few images, a copy of the bottom row of 7 pixels height stick to the canvas.
It's only if the whole image is selected, translating the whole canvas without a selection doesn't have that issue.

Merge as is? The most obvious and annoying artifacts are gone after all. Good work!

dsizzle:
Please merge as-is if it’s mostly better. Thanks!

PR #399, opened 2022-11-18
ImageView: decrease change counter on undo. clear changes if no more undos

### LINKED ISSUE #357: "Undo" seems not to decrease the "change counter"
* Open an image
* Use some pixel altering tool, e.g. create a rectangle. The change is registered, a "*" is added to the window title.
* Undo

Now we're back to the state after opening the image.
Still, the "*" indicator is still there, and when closing you're prompted to save, because "you made 2 changes since last time the project was saved". Apparently the drawing action is count, plus the undo action.

Undoing should decrease the "change counter", so that prompt when closing isn't triggered.

dsizzle:
Fixes #357


humdingerb:
Fixed indeed!

PR #401, opened 2022-11-19
ImageView: only capture keypresses when window is active (except space)

### LINKED ISSUE #400: Unfocused ArtPaint catching all key presses
The recent introduction of SetEventMask() to capture SPACE for panning has some unintended side-effect:
Even if no ArtPaint window has focus, e.g. while typing in Terminal, the key input is captured and interpreted as tool changes. Oops. :)

dsizzle:
oops indeed.  I think this change still allows space panning, but filters out everything else unless the window is active

Fixes #400

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -271,39 +271,43 @@ ImageView::DrawManipulatorGUI(bool blit_image)
 void
 ImageView::KeyDown(const char* bytes, int32 numBytes)
 {
-	if (*bytes == B_LEFT_ARROW) {
-		BRect bounds = Bounds();
-		float delta = bounds.Width() / 2;
-		if (bounds.left - delta < 0) delta = bounds.left;
-		ScrollBy(-delta, 0);
-	} else if (*bytes == B_UP_ARROW) {
-		BRect bounds = Bounds();
-		float delta = bounds.Height() / 2;
-		if (bounds.top - delta < 0) delta = bounds.top;
-		ScrollBy(0, -delta);
-	} else if (*bytes == B_RIGHT_ARROW) {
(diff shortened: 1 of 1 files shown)


humdingerb:
Is it at all possible to have those shortcuts active if _any_  ArtPaint window is active, not just the canvas window?
One may have fiddled in the Layers window and hit a shortcut to change tools while the Layers window is still active.

Generally, I find myself often invoking normal ALT shortcuts when the canvas window doesn't have focus, which then of course don't work... Can all windows forward commands to the canvas window? I imagine one would have to keep track of the last active canvas window, so it's sent to the right one, if you have more than one image open?

Also, can we use the same SPACE-panning trick to make DEL do Edit|Delete?

dsizzle:
> Is it at all possible to have those shortcuts active if any ArtPaint window is active, not just the canvas window?
One may have fiddled in the Layers window and hit a shortcut to change tools while the Layers window is still active.

Sounds tricky - how do you know if someone is just typing a layer name in the Layer window?  In that case you wouldn't want to have tools changing as you type?  So you'd have to filter out keypresses when a text field is selected - not sure if that's possible?  

> Can all windows forward commands to the canvas window? I imagine one would have to keep track of the last active canvas window, so it's sent to the right one, if you have more than one image open?

maybe?  definitely outside of the scope of this change. 

> Also, can we use the same SPACE-panning trick to make DEL do Edit|Delete? 

Probably.

humdingerb:
> Sounds tricky - how do you know if someone is just typing a layer name in the Layer window?

Not sure... :)
Maybe the BTextControl's MakeFocus() can be used to signal to the main window(s) to ignore KeyDown messages until the BTextControl is invoked or loses focus.

Anyway, quite rightly not the scope of this PR. Merging!

PR #404, opened 2022-11-19
ImageView: properly cancel manipulators when interrupted

dsizzle:
When a manipulator is active (includes Add-Ons), and another manipulator is started, the previous behavior was to execute the changes of the active manipulator. This new behavior cancels the previous manipulator before starting the new manipulator. For example, if a user chooses "translate" and begins moving a selection, and then chooses "rotate", the previous behavior acted as if "Ok" was clicked, and now it acts as if "Cancel" was clicked.

Part of #282


humdingerb:
Works fine!

PR #406, opened 2022-11-20
PaintWindow: quit canvas size window when closing paintwindow

### LINKED ISSUE #344: Stray "Create canvas" popup window
Apparently it's still possible to have the "Create canvas" window pop up...:

* Launch ArtPaint.
* Open some image (opens in a new window).
   **Side-request:** Can we just close the empty start window + "Create canvas" window in such a case?
  If users open an existing image and don't create a new canvas at the start, chances are they will just work on the opened image.
* Close the empty start window.
* Invoke the Instant Color Picker by holding CTRL and click-dragging in the opened image.

--> Up pops the "Create canvas" window. Disappears when you let go of the mouse button.

dsizzle:
- also hide window when image is opened.

Fixes #344


humdingerb:
Been on a successful bug hunting, eh? :)
Merging.

PR #408, opened 2022-11-21
ImageView: If delete key is pressed, delete selection or layer

### LINKED ISSUE #402: Add DEL as shortcut for Edit > Delete
As the title says - the DELETE key should be a shortcut for Edit > Delete.

dsizzle:
...only works when the canvas is active for safety reasons...I wouldn't want a spurious delete causing havoc.

Fixes #402

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -305,6 +305,9 @@ ImageView::KeyDown(const char* bytes, int32 numBytes)
 			if (bounds.bottom + delta > bitmap_rect.bottom)
 				delta = bitmap_rect.bottom - bounds.bottom;
 			ScrollBy(0, delta);
+		} else if (*bytes == B_DELETE) {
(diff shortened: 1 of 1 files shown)


humdingerb:
Deleting the whole layer if there's no selection is a bit inconsistent. If "no selection" == "all selected", that should also apply to Edit|Copy/Cut.  Which would be fine by me. After all, that's how things work with manipulators (Translate,Rotate) and add-ons.

dsizzle:
I agree.  I didn’t change any functionality for this change though, merely wired the delete key to the Edit -> Delete function.

I think the problem is that we changed “clear layer” to “delete”, because it also works with selections. Also we didn’t have the distinction between select all and select none until recently- I bet this got lost in all the shuffle

humdingerb:
OK, then let's merge this and I open a ticket to make Copy/Cut do copy/cut whole layer (or even all layers with SHIFT) if there's no selection.

PR #409, opened 2022-11-21
Selection: rewrote Erode to be true "erode" instead of invert + dilate

### LINKED ISSUE #336: "Select all"-selection can't be shrunken
* ALT+A to select all
* SHIFT+ALT+G to shrink the selection

Doesn't shrink.

dsizzle:
Per the definition of how erode/shrink works, the edges of an image *should* get removed - so shrinking a "select all" selection is correct. 

Fixes #336


humdingerb:
That works! While testing, I found #410.
Merge!

Issue #156, opened 2022-05-03
Edit menu: Add a "Scale..." action


humdingerb:
ArtPaint misses a way to scale the currently selected part of an image.
Should be similar to the "Scale" window of the _Canvas_ menu, though I think we only need the Width and Height text controls (BSpinner?) and the "Lock" checkbox

PeteCA:
Ohh.. snap! (:-))

I think again, though, I'd always prefer to do it to a layer, rather than a selection.

humdingerb:
I think it's easy to do a quick ALT+A to select the whole layer and do the resizing.  OTOH, if you could only scale the whole layer, you'd need to first copy a selection, paste it to a new layer and then do the resizing and possibly afterwards merge the layer back to the layer below.

Maybe a good compromise: If there's nothing selected, scale the whole layer.

In any case, currently there's only scaling for the whole canvas, i.e. you couldn't even scale just one layer.

PeteCA:
> Maybe a good compromise: If there's nothing selected, scale the whole layer.
> 
Actually, this makes a lot of sense.  For the resize/rotate/etc operations we're
talking about, you're not likely to want to apply them to an empty selection(!),
so having  the default be  the whole layer seems sane.

dsizzle:
ugh, this is not trivial.  I spent the last several days on this.  will revisit again later.

PR #414, opened 2022-12-04
ScaleManipulator: add ability to scale selections

dsizzle:
at long last, I *think* this works in every case.  The code isn't as pretty as I'd like but I think it works and so here we go.

- caveat, I did contract covid-19 yesterday and so my brain is either free of its normal restrictions, or completely off.  

Fixes the hardest part of #156 

hopefully nothing else broke along the way

### DIFF
--- a/artpaint/application/HSPolygon.cpp
+++ b/artpaint/application/HSPolygon.cpp
@@ -143,7 +143,9 @@ HSPolygon::ScaleBy(BPoint origin, float sx, float sy)
 {
 	for (int32 i = 0; i < point_count; i++) {
 		points[i].x = (points[i].x - origin.x) * sx + origin.x;
+		points[i].x = min_c(100000, points[i].x);
 		points[i].y = (points[i].y - origin.y) * sy + origin.y;
+		points[i].y = min_c(100000, points[i].y);
 	}
 }
 
--- a/artpaint/application/Selection.cpp
+++ b/artpaint/application/Selection.cpp
@@ -586,6 +586,39 @@ Selection::ScaleBy(BPoint origin, float dx, float dy)
 
 
 void
+Selection::ScaleTo(BPoint origin, float x_scale, float y_scale)
+{
+	acquire_sem(selection_mutex);
+	if (original_selections != NULL) {
+		for (int32 i = 0; i < selection_data->SelectionCount(); i++) {
+			delete original_selections[i];
+		}
+		delete[] original_selections;
+		original_selections = NULL;
+	}
+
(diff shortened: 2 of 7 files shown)


humdingerb:
Sorry to hear about your corona infection. Hope you come through quick and unscathed! You're a bit late to the game, aren't you, if that's your first infection? Personally, I'm still a covid-virgin myself... :)

The scaling seems to work very well in any case! The virus hasn't robbed you of all capabilities yet. :)
I only found the unrelated #415.
And I wonder if we should try to find another keaboard shortcut for scaling. The modifier keys gets a bit ridiculous. :)

As with the original canvas scaling, the usage is a bit strange though. Wherever you click first, the image jumps already scaled accordingly. It'd be nicer to have the first mouse down set the "point of origin", i.e. the 100% scale. Dragging the mouse from there scales the image until you mouse up. Then you can repeat that, i.e. click again to set the new "point of origin" and allow further scaling from there.
Even better would be a bounding box with handles, like the canvas crop, of course...

Also, filter options would be nice, too. Like "Repeat pixels", "Bi-linear", "Resample" etc.
But I suppose that's for further down the line.

Merge as is?

humdingerb:
Oh, I spoke a bit too soon...
There seems to be a regression:
* Open an image
* Create a rectangular selection
* ALT+T to translate
---> The artifacts are back. o_O

I think those were eliminated with 1ab6af4f9cf70d64b511f4cae306c1a4030ad57f

dsizzle:
Thanks, yes, it's my first round with corona - I've been very careful and gotten all my vaccines - but then I had dinner with a family member who didn't tell me they weren't feeling well until after I arrived...and then a few days later called me to say they had covid...grrr

anyway if this patch passed muster, what I'd like to do as an enhancement is add a scaling box like crop.  It works much better imo - scale only lets you drag down and to the right and you ought to be able to scale in any direction.  

I also was researching different types of sampling pre-covid and was hoping to add those options as well as a follow up here.  I *think* this current method is basically "bilinear"

interesting about the regression - of all things to break I wouldn't have expected that one.  I'll see if I can repro it over here in my isolation chamber...

dsizzle:
This branch was a bit behind because I've been working on it so long - I tried rebasing it and I don't see any artifacts.  Can you try again with the latest version here please?

humdingerb:
That was it! Works all nicely now. Merge!

dsizzle:
Yes, please merge.

humdingerb:
I thought I already did... Sorry. :)

PR #419, opened 2022-12-06
RotationManipulator: removed rotate90 and use normal RotationManipulator

### LINKED ISSUE #379: Canvas rotation with selection is inconsistent
Steps to reproduce:

1. Create a project with 3 layers, e.g. by putting 3 different images on each layer
2. Create a selection
    a) Canvas | Rotate + 90° --> The whole canvas is rotated, instead of just the selection on all layers.
    b) Canvas | Rotate... --> Here the selection on all layers are rotated, but only on the ~~current~~ bottom  layer the selected area is cut out and rotated. On the other layers a copy of the selected area is rotated.

dsizzle:
- added ability to run any manipulator without the GUI window enabled: note that it requires setting up any settings properly as you'd expect
- removed Rotate90Manipulator files and references
- re-rotate selection in ManipulateBitmap so that it works with headless mode and also on multiple layers

Fixes #379


humdingerb:
Works, everything's still rotating.

Issue #418, opened 2022-12-04
Edit menu consistency


dsizzle:
- Undo/Redo change to "Undo (Redo) not available" when there's nothing on the undo/redo stack.  They should just gray out instead of changing the menu item
- Cut and Copy have submenu items for "all layers" and "current layers", but Paste has "Paste as new layer" and "Paste as new project" as explicit menu items. Why not make them submenu items under "Paste"?

humdingerb:
> Why not make them submenu items under "Paste"?

Possible. Though the other submenu-action are all "Active layer | All layers", so will be the other actions, if it's decided to move them over from the Layers menu (rotate, translate, scale...). Or we'll go with the a bit hidden" SHIFT-changes-menu" road to eliminate these submenus.

Alternatively, the function could move to File under "New project..." as "New project from clipboard".

dsizzle:
> Alternatively, the function could move to File under "New project..." as "New project from clipboard".

not a bad idea...

humdingerb:
Can't it be done by simply overriding the window's MenusBegging() hook change those menus' label, shortcut, message if the SHIFT keys is held?

To make it a bit more discoverable, we could add a hint to the status bar help text "...( add SHIFT for all layers)."

humdingerb:
Pfftt, "MenusBegging()"... at least it's an existing word. The counterpart is "MenusEgging()" btw...

dsizzle:
I tried adding the SHIFT to change to All layers and didn't really like how it felt.  Maybe if it was live it would be more compelling, but having to hold SHIFT before the menu opens feels clunky.  With removal of the "Edit > Rotate+/-90" options the menu looks clean enough in my opinion.  I'm going to close this for now.

humdingerb:
I think you're right. Also, these commands are so ubiquitous that most users learn the "active Layer" shortcuts quickly by heart and that adding SHIFT always extends to "all layers".
BTW, I was wondering if it'd make sense to limit this to "all visible layers". What do you think?

dsizzle:
intuitively I think that makes sense.  I feel like other tools work that way - don't operate on anything you can't see?  

I feel like that's probably an easy change. I think the "all layers" operations loop through all the layers and could just skip any that are not visible.

PR #420, opened 2022-12-06
PaintWindow: move Layer/Canvas operations to Edit menu

### LINKED ISSUE #155: Layer menu: move action to Edit menu & change behaviour

### LINKED ISSUE #156: Edit menu: Add a "Scale..." action

dsizzle:
- each operation has two submenus, one for "All layers" and one for "Active layer"
- added menu items for Rotate +- 90 for "Active layer" which didn't exist before
- removed Canvas menu because the only item left was "Crop" so moved that to Edit as well
- removed "not available" from Undo/Redo menu items and make them start disabled.

Fixes #155 
Fixes #156 
Partially address #418


humdingerb:
Nice change! You appear to come well through your covid adventure so far! :) Hope you are/stay/get well.

One thing no so nice is the Scale shortcuts being CTRL+ALT+S (and +SHIFT) . Preferrably, they should fit into the other edit commands ALT+something. 'E' would still be available. At least there's an 'E' in 'Scale'...

The Crop shortcut CTRL+ALT+C, while also 'non-standard, I suppose it's a seldomly enough used action, compared with the other transformation actions.

dsizzle:
updated - I'm fine with ALT+E 

covid update: definitely getting better but I expect a couple more days before I'm back to normal.  worst part was the sore throat. but I'm stuck all by myself in a room so not much else to do...one can only watch so much TV!

humdingerb:
Good to hear. Dr. Humdinger prescribes light ArtPaint coding, stat! :)

Played around with this change a bit more, and it might be that the removal of the "Canvas" menu/actions was premature.
* Arguably, the +/-90° rotation is mostly used to rotate the canvas portrait/landscape. However, now it only rotates the active/all layers, not the canvas. Parts of the image will get cut off.
* Scaling the active layer lets you scale beyond the canvas, effectively cutting it off. Expected I'd say. Scaling all layers actually scales the canvas, nothing gets cut off. Inconsistent.

Maybe the "Canvas" menu should be kept. Move "Rotate +/-90°" and "Crop..." over from "Edit" and add a "Scale...", all applying to the canvas itself. A new action "Size..." can be added to change the canvas size without scaling the image pixels, possibly with a DPI setting etc.

dsizzle:
> Arguably, the +/-90° rotation is mostly used to rotate the canvas portrait/landscape. However, now it only rotates the active/all layers, not the canvas. Parts of the image will get cut off.

This is actually a bug with #419 - it has nothing to do with this change. However, the simplest fix could be to simply add back the old Rotate+-90 manipulators and put them under the Canvas menu again.

> Scaling the active layer lets you scale beyond the canvas, effectively cutting it off. Expected I'd say. Scaling all layers actually scales the canvas, nothing gets cut off. Inconsistent.

This is related to why I was proposing having operations in the Canvas menu ignore selections and actually operate on the whole canvas.  Seems like maybe we need both "all layers" and "canvas" modes.  I spent a lot of time making it work the way it does though. :sob:  

>  A new action "Size..." can be added to change the canvas size without scaling the image pixels

Crop already does exactly this - you can make the canvas larger or smaller without affecting the image pixels

dsizzle:
I put back the Canvas menu with the Rotate+-90 and Crop in it.  I don't really want to do the canvas scale within this same change.

humdingerb:
Alright, let's merge.

I do wonder though, if the +/-90° items in Edit are an often enough taken shortcut that they are warrented.
We could add +/-90° buttons, and possibly even a 180° button - we do have enough space in those transformer windows.

dsizzle:
I'm honestly in favor of taking them out - they are merely a convenience, and we already have both snapping AND the ability to just type in "90" in the box.  Even buttons seems a bit of overkill.

humdingerb:
Fine with me.

PR #422, opened 2022-12-07
ImageView: don't send mouse click through when panning.

### LINKED ISSUE #421: Bug: SPACE-panning with Intelligent scissors and wand
* Open an image
* Zoom in so you can pan
* Choose Magic wand or Intelligent scissors selection tool
* Hold SPACE and start panning

-> The magic wand gets triggered
-> The intelligent scissors are started, but there's no way to set a point or abort the tool. You'll have to kill ArtPaint in the end...

dsizzle:
Fixes #421 

... well, this prevents unwanted magic wand selections, and getting into a bad state with the intelligent scissors. 

However, note that it's not possible to pan *while* you are making an 'intelligent scissors' selection. The reason is because only one operation can "own" the mouse at a time, and so once you start a selection, you can't interrupt it to pan. It's like if you are drawing a selection rectangle and you try to pan, it won't work.

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -981,23 +981,24 @@ ImageView::MouseDown(BPoint view_point)
 
 			SetCursor();
 			release_sem(mouse_mutex);
-		}
-		if (gui_manipulator != NULL)
-			start_thread(MANIPULATOR_MOUSE_THREAD);
-		else if (fManipulator == NULL) {
-			if (buttons & B_SECONDARY_MOUSE_BUTTON) {
-				BMenuItem* item =
-					ToolManager::Instance().
-						ToolPopUpMenu()->Go(ConvertToScreen(view_point));
-				if (item != NULL) {
-					ToolManager::Instance().ChangeTool(item->Command());
-					SetCursor();	// If the tool changes, the cursor should change too.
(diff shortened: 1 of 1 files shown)


humdingerb:
> However, note that it's not possible to pan while you are making an 'intelligent scissors' selection. 

That's very unfortunate. When using the scissors, you often zoom in real close to see it's doing a good job. That means you'll probably have to pan around often.
I'm sure you've thought about it a lot already. Any chance you'll find a way to scissor and pan? :)

dsizzle:
well we should definitely merge this no matter what because the scissors state is a bad bug.

regarding scissors and pan, I know it's probably something that needs to be fixed in the long run, but it might require major surgery.  

Honestly, though, in my past experience with Photoshop, I believe that I had a different workflow - I believe I used the smart scissors to get 90% of the selection, then went in and used freehand selection or selection painting to fix the mistakes.  I don't think I ever tried to get it perfect with just the scissors.

humdingerb:
You know, you may be right and I have misremembered creating a bezier shape around the contour of an object in WonderBrush...

Another thing I just noticed: Mabe auto-scrolling while using any tool would be nice. If you come near the window border, auto-scroll in that direction, so you can reach the canvas beyond and see where you're going.

Anyhoo, merge!

PR #423, opened 2022-12-08
Scale: added back old scaling as new manipulator ScaleCanvasManipulator

dsizzle:
- re-added Scale to Canvas menu, and it only scales the entire canvas
- also made Scale->All layers not resize the canvas

I don't think there was a ticket for this; it was just a followup from the recent menu re-arranging.


humdingerb:
Good. As canvas Crop... already uses the shortcut CTRL+ALT+C, should canvas Scale.. use CTRL+ALT+E?
(We have to be careful with CTRL+ALT because it's a system window managing shortcut. Z to zoom, H to hide, cursor keys to switch workspace etc.)

humdingerb:
I'm honestly not sure...
On the one hand, using CTRL+ALT is OK as long as we don't collide with system shortcuts, and keeping the label Scale and shortcut 'E' the same as in the Edit menu is at least consistent.
On the other hand, CTRL+ALT is quite unusual to use in an app. But a graphics just has so many commands. And will get rather worse over time, I suppose. Until we save on commands with one big Transform... action that combines translate, rotate,scale and flip.
On the third hand, does every action need a shortcut? One could argue that the Canvas operations are not used that often. The Layer commands are probably used more often (at least adding and merging down, I imagine) and they don't have shortcuts yet.

You decide. :)

dsizzle:
I say no shortcut.  More oftenly-used ones like "Layer->Add" still don't have shortcuts.

humdingerb:
Go for it.

humdingerb:
I thought you may want to remove the "Canvas |  Crop..." shortcut as well as it's also a CTRL+ALT shortcut and probably not used often.

dsizzle:
good call - removed the Crop shortcut.

humdingerb:
And we merge.

PR #426, opened 2022-12-08
FilePanel: rearrange save dialog so it behaves better with large font sizes

### LINKED ISSUE #217: Save image panel: make font sensitive
Currently, the "Save format" pop-up menu screws up the layout with larger font sizes. The rest of the widgets  don't leave it enough space. The menu bar is also affected:

save_fontsize-18

It'll also be nicer to have the "Settings..." button to the right of the "Save format" pop-up menu.
Plus, move the file name text box above the format settings. It being just one regular rectangle,  looks nicer when positioned directly under the equally rectangular file list box.

dsizzle:
- moved format box down and put filename box back in line with buttons
- made filename box resize based on placement of settings button

Fixes #217


humdingerb:
Weird that Haiku's file panel doesn't use layout management yet...

PR #427, opened 2022-12-09
ImageView: draw manipulator GUIs outside the bounds of the image

dsizzle:
...as the title says, I believe I have manipulator GUIs drawing outside canvas - Crop is the only one I think that really does anything at the moment.  Note that it's not *selections* but just manipulators.  

Selections might be really hard to do, and I'm not quite sure it's useful anyway?  

Anyway please give this a try when you have a chance and let me know if you see any problems:
crazycrop

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -262,6 +262,15 @@ ImageView::DrawManipulatorGUI(bool blit_image)
 			}
 		}
 
+		BView* parent = this->Parent();
+		BRect imgFrame = ConvertToParent(Bounds());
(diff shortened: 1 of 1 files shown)


humdingerb:
> Crop is the only one I think that really does anything at the moment.

Planned for canvas scale, too, in the future?

> Selections might be really hard to do, and I'm not quite sure it's useful anyway?

It's a bit awkward ATM when you e.g. Translate your selection outside the canvas, it becomes invisible. And generally, I'm a fan of a bounding box with grabbers for scaling, rotating, (future shearing and free transformation). One step at a time.

humdingerb:
I had a sort of idea that may be related to this, see #428.

PR #430, opened 2022-12-10
ImageView: Press ESC to cancel manipulators.

dsizzle:
Part of #282


humdingerb:
Very nice! I'm a major user of "ESC = Dismiss". \o/
Q: Would it be equally easy to have "RETURN = trigger active button (default being OK)" for those manipulator windows? Once you use the mouse, the manipulator window loses focus and the RETURN doesn't reach it.

Merging in any case.

dsizzle:
I don't think it would be equally easy but might not be too hard.  lemme add that to #282

dsizzle:
oh also I really wish ESC worked for intelligent scissors too...

humdingerb:
:+1

PR #431, opened 2022-12-10
Rename canvas "Scale..." to "Resize...", update en.catkeys

humdingerb:
Now that the canvas menu items have no shortcuts, there's no mnemonic need to share the menu label "Scale" and shortcut 'E' between Edit and Canvas menu.

To avoid confusion (e.g. in the docs or when discussing among users) it's better to go with another label for the canvas
scaling: "Resize"


humdingerb:
OK with you, @dsizzle  ?

dsizzle:
Totally ok with me!

PR #433, opened 2022-12-10
FontManipulator: make anti-alias setting "live"

### LINKED ISSUE #211: Text tool: Apply anti-aliasing setting changes live
Other than the size/rotate/shear sliders, changing the setting of the anti-aliasing checkbox isn't seen live on the canvas. You need to either move the text a bit in the canvas or use the size/rotate/shear sliders to force an update.

dsizzle:
- when comparing settings for font changes, check if font Flags() are the same - even if the font itself is the same, Flags() changes need to be reacted to.

Fixes #211


humdingerb:
Confirmedly fixed. Thanks!

PR #435, opened 2022-12-11
ScaleCanvasManipulator: make canvas Resize work like Crop

### LINKED ISSUE #416: Resize manipulator should work like Crop
The Resize manipulator currently only scales downward and to the right, but it should scale in all directions similar to how the Crop box works.  It should also have a box so the user can see what they're grabbing - currently the scale can jump to where the first click is.

### LINKED ISSUE #434: Crop manipulator outside of canvas doesn't go away on Cancel
expand the Crop manipulator outside the bounds of the canvas, and then hit Cancel - the crop box stays (the background doesn't redraw)

dsizzle:
- draws a box with 'grabbers' around the canvas
- allows scaling in any direction
- SHIFT locks aspect
- fixed minor typo in CropManip
- fixed redrawing of background when Cancel is selected

Fixes #416 
Fixes #434

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -917,6 +917,7 @@ ImageView::MessageReceived(BMessage* message)
 					fManipulator = NULL;
 					the_image->Render();
 					manipulated_layers = HS_MANIPULATE_NO_LAYER;
+					Parent()->Draw(Parent()->Bounds());
 					Invalidate();
 					start_thread(MANIPULATOR_FINISHER_THREAD);
 				}
--- a/artpaint/viewmanipulators/CropManipulator.cpp
+++ b/artpaint/viewmanipulators/CropManipulator.cpp
@@ -244,7 +244,7 @@ CropManipulator::MouseDown(BPoint point, uint32, BView*, bool first)
 		settings->left = settings->right - 1;
 
 	if (settings->top >= settings->bottom)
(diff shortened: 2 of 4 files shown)


humdingerb:
Nice change!

I'm wondering about the "Lock" checkbox. It's always active when opening the Resize... panel. If you don't want an aspect-locked resize, you start your resize with the mouse, then realize that it's not doing what you want, have to spot the Lock checkbox and move the mouse over to uncheck it.
OTOH, now that SHIFT-lock works, maybe the checkbox should be off by default. The users will learn to hold SHIFT when resizing with the mouse to aspect-lock, and when manually entering dmensions, they are near the checkbox already and change the tick if needed.
Or save the state of "Lock" (at least for the session)?
I expect that most of the time people want to lock aspect when resizing the canvas, but I guess sooner or later the "Edit | Scale..." will stand before the same question. And I expect that other when resizing the canvas, people will most often only use the mouse for scaling a selection. In that case it's much easier to use SHIFT instead of toggling a checkbox.
I dunno... your call.

While we're in the Canvas menu: I think the separator line in the menu can be removed now.

Let's merge.

dsizzle:
These are the great philosophical questions of our time... does it make sense to save a setting for an operation you are likely to only do once or twice per image?  

Overall, though, I think that every UI element should have its state saved.  I don't think that checkbox is even currently part of the "settings" object.  I ran into this trying to implement the 'Revert' button and didn't feel like dealing with it at the moment.

Issue #352, opened 2022-10-28
Wheel-zoom: hit major percentages


humdingerb:
Currently when mouse wheel zooming, the zoom factor goes to arbitrary percentages. BTW, is displaying decimal points for the factor necessary. Feels needlessly precise...

It'd be nice if the zoom factor hits the 50%, 100%, 150%, 200% etc. zoom factors.

dsizzle:
The little popup slider for Zoom also allows arbitrary decimal percentages.  Maybe that should also be less precise?  The slider feels fine but it's true that the zoom numbers aren't necessarily meaningful.

humdingerb:
To be onest, I never much used the slider. I rather clicked a few times on the x/- buttons.
Also, now that we have excellent mouse wheel zooming - even honing in on the mouse poiinter - I don't see much use in the slider.
I'd be OK to remove it, but it doesn't get in the way either.

Coming to think of it: instead of the slider, a list of percentages wold be more useful when "Zoom" is clicked: maybe 25%, 50%, 100%, 200%, 300%, 400%, 600%

humdingerb:
Fixed with #436

PR #436, opened 2022-12-12
ImageView: improve mouse zoom

dsizzle:
- below 100%, zoom tries to hit exact values
- removed decimal from Zoom as it's unnecessary
- fixed mouse zoom so it'll actually hit 1600% now
- fixed bug where +/- zoom buttons were not reflective of current zoom (i.e. if the user mouse-zoomed to 150% and then hit the + button, the zoom wouldn't go to 200%, it would go to whatever the last zoom level was before mouse-zooming)

Addresses #352, maybe fixes it? You can always hit the exact % above 100 by pressing +/- to get to the next nearest zoom level


humdingerb:
Good! I think that closes #352 indeed.

PR #438, opened 2022-12-16
GlobalSettingsWindow: restore lost change that makes window larger

### LINKED ISSUE #437: Preferences window size for languages other than English
Hi there. Sometimes captions get cut in the Preferences window when using languages other than English (w/longer sentences). This also causes some controls to shift to the right and partially hide.

As it seems this window isn't dynamically-sized and there is plenty of room, is it possible to enlarge it a bit to accommodate these edge cases?

ap1 ap2

dsizzle:
- fixes non-English settings window being cutoff

Fixes #437


humdingerb:
Confirmed, fixed. Again... :}

PR #439, opened 2022-12-16
ScaleManipulator: Added selection box with handles for Scale

dsizzle:
ok, I made scaling of selections work with the bounding box with handles.

### DIFF
--- a/artpaint/application/Selection.cpp
+++ b/artpaint/application/Selection.cpp
@@ -419,7 +419,7 @@ Selection::Erode()
 		1	1	1
 
 		All 1s must be contained in the selection for every X, otherwise X is
-		not included in the new selection. The image edges are always eroded 
+		not included in the new selection. The image edges are always eroded
 		because they cannot satisfy this condition.
 	*/
 	selection_bounds = GetBoundingRect();
@@ -437,12 +437,12 @@ Selection::Erode()
 		int32 right = selection_bounds.right;
 		int32 top = selection_bounds.top;
 		int32 bottom = selection_bounds.bottom;
-		
+
 		for (int32 y = top; y <= bottom; ++y) {
 			for (int32 x = left; x <= right; ++x) {
 				if (ContainsPoint(x, y)) {
 					// edge pixels get eroded
-					if (x == 0 || y == 0 || 
+					if (x == 0 || y == 0 ||
 						x == width || y == height) {
 						*(new_bits + y * new_bpr + x) = 0x00;
 					} else {
@@ -456,8 +456,8 @@ Selection::Erode()
 								}
 							}
 						}
-						
-						if (all == false) 
+
+						if (all == false)
 							*(new_bits + y * new_bpr + x) = 0x00;
 					}
 				}
@@ -609,9 +609,6 @@ Selection::ScaleTo(BPoint origin, float x_scale, float y_scale)
 		HSPolygon* p = selection_data->ReturnSelectionAt(i);
 		p->ScaleBy(origin, dx, dy);
 	}
-
-	selection_bounds.right *= dx;
-	selection_bounds.bottom *= dy;
 
 	needs_recalculating = TRUE;
 	release_sem(selection_mutex);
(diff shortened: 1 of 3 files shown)


dsizzle:
Don’t merge this yet, I found some bad behavior if you have a selection and move it so that the upper left corner is at the bottom right of the canvas. It kinda gets stuck there.

humdingerb:
Otherwise, nice change!

Now that you can "Translate" with the "Scale" action, how are your feelings on consolidating into one action. Would make most sense (and simplifies naming the thang to a general "Transform") if "Rotate" could be melted in there, too.

Also: is it possible to have ESC only resetting to initial conditions if the user has moved/scaled/rotated the selection? If there's no change, abort the action and close the window.

humdingerb:
> if you have a selection and move it so that the upper left corner is at the bottom right of the canvas. It kinda gets stuck there.

Or, another nice effect: try moving the bounding box out of the canvas at the top left corner.
I suppose it's easy to replicate, but here's a screencapture: http://0x0.st/onl8.mpg [7.4 MiB]

dsizzle:
> how are your feelings on consolidating into one action

I think it makes sense but the thought of mixing in Rotation fills me with great fear… scale has been quite challenging to get right in all the possible cases. There is commented-out code for a Free Transform manipulator but I haven’t looked at it. I assume since it’s commented out the original author couldn’t get it working.

> Also: is it possible to have ESC only resetting to initial conditions if the user has moved/scaled/rotated the selection? If there's no change, abort the action and close the window.

this doesn’t quite feel right to me. Kind of like an ESC-to-undo. I’ll keep it in mind.

humdingerb:
> I assume since it’s commented out the original author couldn’t get it working.

I have complete confidence in your code fu!

> this doesn’t quite feel right to me. Kind of like an ESC-to-undo. I’ll keep it in mind.

Mull it over. It's more a ESC-to-revert, as there's nothing yet to undo because the action wasn't applied yet. Worst case, people have to hit ESC twice, once to revert the move/scale/rotate to reset all to the initial condition, once more to abort the action and close the manipulator window.

dsizzle:
ok, I think I fixed the bad behavior.  I think you can still get a bit stuck if you mess around with going completely outside the upper left corner but you can always hit ESC to recover (or Cancel).  I'm not sure why someone would be doing that anyway.  

also the Scale "translate" isn't quite the same as the Translate "translate" - it gets cranky when you go outside the bounds. I think with Translate you can go far outside the canvas and it's fine. 

However, I did notice that if you create a selection out of a couple of boxes that aren't touching, and then you scale it so that one of the boxes is outside the canvas, then use Alt+T to Translate it, you get some artifacts.  

I think as we've discussed in various tickets, out-of-canvas behavior is not really ideal overall in ArtPaint. 

But I think this change is good enough at the moment.

dsizzle:
oh also I have run into about 2 crashes along the way but whenever I've tried to reproduce them I cannot.  If you happen to, please let me know. Maybe it was due to my heavy development but I'd love to shake them out if they still exist in this patch.

humdingerb:
A crash easy to reproduce: Invoke any manipulator (rotate/translate/scale) and while using the mouse to do the manipulatin', press ESC. Boom: 
[ArtPaint-2706-debug-18-12-2022-05-30-50.report.txt](https://github.com/HaikuArchives/ArtPaint/files/10253251/ArtPaint-2706-debug-18-12-2022-05-30-50.report.txt)

Something else. Difficult to describe, so I'll just upload the screencapture: http://0x0.st/o5rz.mpg [5 MiB]

Tell me if you prefer to have this PR merged as is and have another PR with fixes.

dsizzle:
well, that crash has nothing to do with this PR at all, so that should be a separate issue.  

I'll look into that other behavior. Definitely not ideal, although I hope users don't do that just to be difficult. :laughing:

humdingerb:
Opened a new ticket for the crash: #440 

> although I hope users don't do that just to be difficult. \U0001f606

Typical for these users. If there's a bug they bumble into it sooner or later... :)
Besides, even improbably triggered bugs may have other side-effects.

dsizzle:
ok, pretty sure I fixed that issue. Let me know if you run into anything else!

humdingerb:
Now all is well it seems. Nice change!

PR #442, opened 2022-12-19
ScaleCanvasManipulator: greatly simplified resizing code

dsizzle:
in preparation for/while researching different resizing algorithms, I found that I could simplify the current resizing code immensely. will try to apply the same change to the normal scale manipulator.

seems to work as well as before, but way less code.

### DIFF
--- a/artpaint/application/PixelOperations.h
+++ b/artpaint/application/PixelOperations.h
@@ -748,4 +748,18 @@ inline bool compare_2_pixels_with_variance(uint32 p1, uint32 p2, uint32 var)
 			((p1&0xFF) <= ((p2&0xFF)+var) && ((p1&0xFF)+var) >= (p2&0xFF));
 }
 
+
+inline uint32 lerp(uint32 p1, uint32 p2, float t)
+{
+	union color_conversion one, two, result;
+
+	one.word = p1;
+	two.word = p2;
+
+	for (int i = 0; i < 4; ++i)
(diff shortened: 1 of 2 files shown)


humdingerb:
As a test, I've saved a stupid picture of me (sweating after a bike ride) and saved as BMP. I then resized  from a width of 2988  to 800 pixels with master ArtPaint and this PR build. Put both in a layer and saved as an ArtPaint project: http://0x0.st/o5Kn.artpaint [8.7 MiB]

Switching between those layers, it appears that the new resizing moves the image 1 pixel down and right. Also, if you zoom to 1600% and look at the wall plaster, it appears the old algo has ever so slightly more contrast or sharpness.
It probably doesn't matter at all, but I just wanted to mention it.

Tell me if you want to merge as is.

dsizzle:
I personally can’t see the difference in the wall plaster, which is surprising because I used to play a really dumb game called “I Love Hue” that was all about differentiating subtle color differences. Ah well.

The 1 pixel shift doesn’t bother me. This algorithm is actually bilinear for both enlarge and reduce, whereas the old algorithm is bilinear-ish for enlarge and some sort of weighted average for reduce. I’ll pay the 1 pixel price to use the same code for both.

please merge if there are no other issues, thanks!

humdingerb:
> This algorithm is actually bilinear for both enlarge and reduce, whereas the old algorithm is bilinear-ish for enlarge and some sort of weighted average for reduce.

That may be the reason my eagle eyes see a slight difference.
Anyhoo, yay for grabbered bounding box scaling!

PR #445, opened 2022-12-24
Scale: add different scaling methods

### LINKED ISSUE #417: Scaling should have different resampling modes
Currently Scale only supports one type of resampling.  Many other tools support multiple methods, such as:

- nearest-neighbor
- bilinear
- bicubic
- Mitchell  
- Lanczos

Add a dropdown in the Scale dialog to support these (or some subset/superset of these).

dsizzle:
- added nearest neighbor and bicubic types to the existing bilinear
- also refactored out common code between Scale and ScaleManipulator to make it easier to make common changes like this

Fixes #417

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,8 @@ APP_MIME_SIG= 'application/x-vnd.artpaint'
 #	if two source files with the same name (source.c or source.cpp)
 #	are included from different directories.  Also note that spaces
 #	in folder names do not work well with this makefile.
-SRCS= artpaint/Utilities/BitmapUtilities.cpp artpaint/application/FilePanels.cpp artpaint/application/FloaterManager.cpp \
+SRCS= artpaint/Utilities/BitmapUtilities.cpp artpaint/Utilities/ScaleUtilities.cpp \
+artpaint/application/FilePanels.cpp artpaint/application/FloaterManager.cpp \
 artpaint/application/HSPolygon.cpp artpaint/application/IntelligentPathFinder.cpp artpaint/application/MatrixView.cpp \
 artpaint/application/MessageFilters.cpp artpaint/application/PaintApplication.cpp artpaint/application/ProjectFileFunctions.cpp \
 artpaint/application/RandomNumberGenerator.cpp artpaint/application/RefFilters.cpp artpaint/application/ResourceServer.cpp \
--- a/artpaint/Utilities/ScaleUtilities.cpp
+++ b/artpaint/Utilities/ScaleUtilities.cpp
@@ -1,0 +1,322 @@
+/*
+ * Copyright 2022, Dale Cieslak
+ * Distributed under the terms of the MIT License.
+ *
+ * Authors:
+ *		Dale Cieslak <dcieslak@yahoo.com>
+ *
+ */
+
+#include "ScaleUtilities.h"
+
+
+void
+ScaleUtilities::ScaleHorizontally(float width, float height, BPoint offset,
+	BBitmap* source, BBitmap* target, float ratio,
+	interpolation_type method)
+{
+	uint32* target_bits = (uint32*)target->Bits();
+	int32 target_bpr = target->BytesPerRow() / 4;
+	uint32* source_bits = (uint32*)source->Bits();
+	int32 source_bpr = source->BytesPerRow() / 4;
+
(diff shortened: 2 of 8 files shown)


humdingerb:
You forgot to add "artpaint/Utilities/ScaleUtilities.cpp".
Too much eggnog? :)

dsizzle:
oops!  All the files should be there now.  

> Too much eggnog? :)

Close - I was kind of hurrying to finish this because we are hopping on an airplane tomorrow to spend 11 days in the Caribbean, and in the past 2 days we've had two Christmas celebrations, one with 30 people on my wife's side, and one with 11 people on my side of the family. Today we need to pack our suitcases for the trip.  So, I'm busy with a lot of holiday-related activity!  

I was kind of hoping to get a release together soon but haven't had a chance to assess whether there's more stuff i wanted to try to get in.  I'll be gone from 12/26 through 1/7 so I hope there aren't a bunch of bugs in this code because they'll have to wait! :palm_tree: 

Hope you are having a wonderful Christmas or Hanukkah or whatever holiday is your favorite!  If I remember my high-school German: Froehliche Weihnachten und ein gutes neues Jahre!

dsizzle:
> I hope there aren't a bunch of bugs in this code because they'll have to wait!

...and I just found one - if you translate the scale box it doesn't keep the selection when you hit Ok.  I know that worked fine before this change.

dsizzle:
Ok, I think I fixed that issue at least.  Let me know if you see any other issues.

I don't think it's related to this change, but seems like the scaled selections are 1-2 pixels too small in each direction. I think it was like that before perhaps?

humdingerb:
> hopping on an airplane tomorrow to spend 11 days in the Caribbean, and in the past 2 days we've had two Christmas celebrations, one with 30 people on my wife's side, and one with 11 people on my side of the family. 

I'm envious. However, after those two xmas parties, I suspect you're very much in need of this holiday... :)
Hope to see you fully recovered in the new year! Your higshchool German was almost perfect, the same to you and the Missus.

WRT this PR, I guess the 1-2 pixel size diff is barely noticable, because I didn't.
Those are some exotic algorithms in the list of filters. Do the results really differ that much?

It'd be nice if the the preview were live, then one could compare filters and pick the one with the best result for the current situation. 
Also, the filter setting should be saved and when that is implemented, the Lock setting could just as well be too.
I'll merge as is and add the above as new issues.

humdingerb:
Merged a bit too quickly. :)
I only tested with scaling selections, resizing the whole canvas crashes, see #448.

dsizzle:
> Those are some exotic algorithms in the list of filters. Do the results really differ that much?

Not to my eyes - but with image resizing there can be subtle differences depending on the image.  Most of those are just different parameters to the same method. We can always trim it down if it feels unnecessary.  

"Bicubic" = Adobe Photoshop's version of "Bicubic" method
"Bicubic (Catmull-Rom)" = GIMP's version of "Bicubic" method
"Bicubic (B-Spline)" = Paint.NET's version of "Bicubic" method
"Mitchell-Netravali" = also a type of "Bicubic" but called simply "Mitchell" in ImageMagick

One that seems popular that I wanted to add is Lanczos, but I couldn't figure out the math yet. 

> It'd be nice if the the preview were live, then one could compare filters and pick the one with the best result for the current situation.

I thought the same; maybe some sort of Preview checkbox or button would be nice.  Or a small view of the results. I think Photoshop has little preview windows on the dialog for some operations.  Doing a live preview on the whole image may be too slow.

>  resizing the whole canvas crashes,

augh, this is due to my 1-pixel tweaking I'm sure. I doubt I can fix that before my vacation. :(

humdingerb:
OK, I heard/seen "Lanczos" before.

Let's see how fast the resizing is in the end. At least currently a huge 6000x4000px image resizes in 0.5s. That's with old standard filtering (bilienar?).
This apparently gets multiplied by the number of layers. Maybe there could be one thread per layer for resizing? Or is multithreaded scaling possible? That would be awesome, but many graphics related action appear to be difficult to parallelize...

In any case, maybe it's fast enough on modern hardware to not need a tiny preview window. Then there could be a checkbox "Preview" that can be de-activated when it takes too long.

> augh, this is due to my 1-pixel tweaking I'm sure. I doubt I can fix that before my vacation. :(

Don't sweat it, I thought you were in the air already. Have a nice trip! :)

PR #450, opened 2023-01-09
Crop: subtract 1 for final width/height since we added 1 to start

### LINKED ISSUE #425: Crop grows canvas by 1px
* Create a 100x100 canvas
* Open Canvas|Crop...
* Click OK 

-> Canvas grows to 101x101
Any crop adds 1 pixel to width and height of the canvas.

dsizzle:
- also return "original" bitmap instead of NULL when crop size doesn't change. Seems to prevent occasional crash.

Fixes #425


humdingerb:
Confirmed fixed. Thanks!

PR #451, opened 2023-01-10
CropManipulator: use ScaleUtilities::MoveGrabbers to save some code

dsizzle:
- removed duplicate code for moving grabbers that is shared between scale manipulators and crop

no functional change; crop should behave exactly as before.

I should probably share the drawing code between scale and crop too one of these days...


humdingerb:
Seems to work as well as before. 👍

Issue #176, opened 2022-06-01
Replace Transparency tool with options for Eraser


dsizzle:
The Transparency tool is hard to understand.  The Eraser tool would be more intuitive with some options:

- "Pressure" or "Strength" or something to say how transparent to erase
- "Hardness" like the Brush tool for soft edge erasing

humdingerb:
After looking at the transparency tool for the user docs, I'm not so sure anymore it should be integrated with the eraser. While the eraser could do the "make-slightly-transparent" part with the above mentioned "Pressure/Strength", it cannot do the reverse: make something less transparent. A sort of "anti-eraser"...

I'm coming around to how the transparency tool is working... :)
From the user doc:

> This tool changes the transparency of the touched pixels by applying the "alpha" value of the current foreground color.
> 
> Say you have one image in layer 1 and another in layer 2 above it. If you set the transparency (=alpha) of the foreground color to 50% (the actual color doesn't matter), and start painting in layer 2, the touched pixels become max. 50% transparent and the image of layer 1 shines through.
> 
> If you then set a color with alpha 100%, i.e. fully opaque, you can restore the original image when painting over the same areas you reduced the alpha values before.
> 
> You can set the "Size" of your transparency brush.
> "Speed" controls how fast the alpha value is applied: the lower the speed, the longer it takes for the pixels to become transparent/opaque.

However, it may be more obvious to forget about the alpha of the current foreground colour and just add another slider "Transparency" from 0% to 100%...

PeteCA:
On Tue, Jun 07, 2022 at 09:05:50AM -0700, humdinger wrote:
> After looking at the transparency tool for the user docs, I'm not so sure anymore it should be integrated with the eraser.

I think combining the two tools would *not* be a good idea.  Your analysis
explains why!

dsizzle:
I still think it’s a weird, unintuitive tool and this use case would be better served via a mask. Basing the transparency on the opacity of the currently selected foreground color is imprecise and confusing - especially given the way the “speed” part works.

I agree that ditching the Transparency tool sounds like we’d be losing functionality so that’s not a good idea.

Having said that, I think if we add functionality to Eraser, and make Transparency more intuitive, the tools will be quite close. Maybe Eraser needs an “unerase” mode as @humdingerb suggests?

humdingerb:
> Basing the transparency on the opacity of the currently selected foreground color is imprecise and confusing - especially given the way the “speed” part works.

[...]

PR #453, opened 2023-01-11
EraserTool: Anti-alias edge of eraser

dsizzle:
This is something that's been bugging me for a while - the edge of the erased area is not anti-aliased.  This isn't just a personal preference, it's a bug fix to the compositing - now it'll work either way, if we decide to add a checkbox for antialiasing at some point.  This fix is also needed for the Eraser portion of #176, to support soft-edged erasers.


humdingerb:
Good change!

PR #454, opened 2023-01-12
ColorPalette: support drag/drop of text values

### LINKED ISSUE #257: Accept dropped colours

dsizzle:
i.e.:
- if you drag a color to a text editor, you get a hex value
- if you drag a hex value onto the color palette, it should set the color

Fixes #257 

... I think this works as requested now

### DIFF
--- a/artpaint/controls/ColorPalette.cpp
+++ b/artpaint/controls/ColorPalette.cpp
@@ -597,6 +597,22 @@ void ColorPaletteWindow::MessageReceived(BMessage *message)
 		}
 	} break;
 
+	case B_MIME_DATA: {
+		if (message->WasDropped()) {
+			const char* colorr;
+			ssize_t color_size;
+			if (message->FindData("text/plain", B_MIME_TYPE,
+				(const void**)&colorr, &color_size) == B_OK) {
+				char colorStr[color_size+1];
+				for (int i = 0; i < color_size; ++i)
+					colorStr[i] = colorr[i];
+				colorStr[color_size] = '\0';
+				hexColorField->SetText(colorStr);
+				PostMessage(HEX_COLOR_EDITED);
+			}
+		}
+	}	break;
+
 	case HEX_COLOR_EDITED: {
 		BString hexColor = hexColorField->Text();
 		hexColor.ReplaceAll("#", "");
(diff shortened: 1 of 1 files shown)


humdingerb:
Nice! It works with StyledEdit and any BTextControl - even when editing a file name a d'n'd the selected label.
It doesn't with Pe or Koder. Apparently they don't set the text/plain MIME type...

One thing that could be added: In HTML/CSS files you often see #FFF instead of #ffffff or #ADA instead of #aaddaa. Maybe those could be added as well?

humdingerb:
Hey, wanna get really fancy? Accept "css color names" (https://www.w3docs.com/learn-css/css-color-names.html) as d'n'd text or inserted in the hex-value field. New ticket or would you rather not?

dsizzle:
> One thing that could be added: In HTML/CSS files you often see #FFF instead of #ffffff or #ADA instead of #aaddaa. Maybe those could be added as well?

doesn’t that work already? I’m not at a place where I can try it but you should already be able to type 3, 4, 6, or 8 character colors (the 4 and 8 are with alpha) 

> Accept "css color names"

I’d rather not. That feels like a feature that would be neat but almost never used.

humdingerb:
> doesn’t that work already?

Sorry, it actually does work. I may have dropped the text on the hex-value box before. Doesn't work there,but everywhere else.

> That feels like a feature that would be neat but almost never used.
Probably correct. :)

Let's merge this.

PeteCA:
On Thu, Jan 12, 2023 at 07:38:49AM -0800, Dale Cieslak wrote:
> 
> > Accept "css color names"
> 
> I’d rather not. That feels like a feature that would be neat but almost never used.
> 
I'll just chip in to agree with Dale here.  I very seldom have a use for those.
Sounds like "Unnecessary Feature Creep"...

humdingerb:
Until you have to work with a html/css file using those named colours. Then it would come very handy to use them directy via drag&drop in ArtPaint instead of websearching a colour table and finding the corresponding values to use in your illustration... :)

And at least it would be "invisible" feature creep from the users POV.

PR #455, opened 2023-01-13
ColorPalette: accept B_SIMPLE_DATA as well as B_MIME_DATA message

dsizzle:
added support for Pe too for #257

### DIFF
--- a/artpaint/controls/ColorPalette.cpp
+++ b/artpaint/controls/ColorPalette.cpp
@@ -58,6 +58,65 @@ BList* ColorPaletteWindow::master_window_list = new BList();
 BList* ColorPaletteWindow::palette_window_clients = new BList();
 
 
+status_t
+HexStringToUInt32(BString hexColor, uint32& color_word)
+{
+	hexColor.ReplaceAll("#", "");
+	hexColor.ToLower();
+	union color_conversion color;
+	status_t valid_color = B_ERROR;
+
+	if (hexColor.Length() == 3 || hexColor.Length() == 4) {
+		BString byteStr;
+
+		for (int i = 0; i < 3; i++) {
+			char digit = hexColor.ByteAt(i);
+			uint32 byte = 0;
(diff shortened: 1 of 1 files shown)


humdingerb:
Cool, now works wth Pe and Koder, too!

Does dropping a hex-value have to be handled differently than dropping an actual colour (e.g. from Colors!)?
You can drop a colour directly onto any swatch of the palette, but a hex-value drop will always change the currently selected swatch.
Conversely, you can drop a colour only on a swatch, dropping anywhere in the window does nothing. It could change the current swatch in this case, as dropping a hex-value does already.

dsizzle:
I can make them consistent, should be easy.

dsizzle:
ok, the color d&d works on the entire window like the hex values.

humdingerb:
Sorry, still doesn't work here. I d&d a colour from the Colors! app and only when dropping on the big or a small swatch, it changes colour. Dropping anywhere else doesn't work.

dsizzle:
Ok, this is one instance where I feel reasonably confident in asking if you can please try pulling the code and rebuilding… 😄  I removed the code from the big swatch and put it on the window, so the big swatch doesn’t even handle the d&d at all with this change.

humdingerb:
Damn. I apparently didn't fetch your last changes to the branch...
Sorry about that.
Would it be easily possible to also drop hex-values on a specific swatch to change its colour? That's the only diff left between d&d colours and hex-values.

dsizzle:
Should be no problem. I need to update this anyway since I accidentally left some commented-out code in there

dsizzle:
ok, I think it all works now.  It was more work than I expected and so please check if I broke anything...

humdingerb:
On the first drop of a hex-value (#FFF) onto a swatch, the alpha was set to 0, so #ffffff00. But every following drop worked perfectly. Couldn't reproduce no matter what I tried...
All other d&d action works fine.
I think it's good. Let's merge!

Issue #304, opened 2022-10-14
Colour picker: Instantly show info window


humdingerb:
Currently, if you've select the Color Picker tool (or only temporarily switch to it when holding ALT), you have to click and move the mouse pointer for the RGB-info window to appear. With the sometimes unwanted effect that doing so will change your foreground colour on mouse-up...

I'd be nicer if the info window would just follow your mouse cursor as long as the Color Picker tool is active. Only pick a colour when you click the mouse (left: foreground, right: background, once we get rid of the tool context menu, see #266).

PR #457, opened 2023-01-26
ColorSelector: Draw a box showing the size of the color sample

### LINKED ISSUE #325: Color picker: Show sample size with mouse pointer
It'd be very useful to see the area sampled with the color picker tool. To be nicely visible, I'd say a black&white dashed frame following the mouse pointer.

BTW, I do understand correctly that the set "Size" - 1 to 10 - is the side length of the sample square that the mousepointer sits in the middle of?

dsizzle:
Fixes #325 

I'm sure it would be desirable to show the sample square even when the mouse button isn't down, but that isn't the way ArtPaint works right now.  Same problem with #304 - "tools" in ArtPaint only operate at all when you press the mouse.

of course with code all things are possible... ;D

### DIFF
--- a/artpaint/tools/ColorSelectorTool.cpp
+++ b/artpaint/tools/ColorSelectorTool.cpp
@@ -308,50 +308,62 @@ ColorSelectorTool::UseTool(ImageView *view, uint32 buttons, BPoint point,
 		old_color = color - 1;
 		bool select_foreground = (buttons & B_PRIMARY_MOUSE_BUTTON) != 0x00;
 
-		// for the quick calculation of square-roots
-		int num_sqrt = 500;
-		float sqrt_table[num_sqrt];
-		for (int32 i = 0; i < num_sqrt; i++)
-			sqrt_table[i] = sqrt(i);
+		int32 half_size = fToolSettings.size / 2.;
+		BRect rc = BRect(point.x, point.y,
+			point.x + fToolSettings.size, point.y + fToolSettings.size);
+		rc.OffsetBySelf(-half_size, -half_size);
 
-		float half_size = fToolSettings.size/2;
-		BRect rc = BRect(point.x - half_size, point.y - half_size,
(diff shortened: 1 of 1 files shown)


humdingerb:
Nice!
There seems to be an off-by-1 error, or better 2 off-by-0.5 errors...:

size4

This is with a "size" of 4. The sampling seems to take an additional half-pixel all around.

dsizzle:
Hmm, also looking at the 4px example maybe it’s not 0.5 all around, but rather it just needs to be shifted down and right by 0.5 ? The pointer doesn’t look centered.

humdingerb:
That is just a result of being zoomed in so much that you can move the mouse pointer around within the area of one pixel and the samplebox only jumping to the next pixel-border when you cross over to the next pixel. If you know what I mean... video: http://0x0.st/oFaS.mpg

dsizzle:
this was a bit harder than I thought - can you please give it a try?  I think it's working.

humdingerb:
Almost. Works nicely with all sizes >1. Maybe a 1-pixel sample box needs special casing (?), as the box picks the pixel one pixel to the top and left of the one under the mouse pointer: 
screenshot1

dsizzle:
Odd, 1 pixel worked fine for me. I guess I’ll try it again tomorrow.

dsizzle:
Ok, it should be fixed now.  I also changed the sampling. It was doing some weird square root distance thing and wasn't getting all the pixels. I changed it to just average all the pixels in the square.

humdingerb:
Yep. Works perfectly now, thanks!

dsizzle:
It occurred to me that perhaps the old method was trying to do a circular sample rather than square - but imo not sure if that’s useful or intuitive in a pixel drawing app?

humdingerb:
Yeah, I think going with a square sampling box is a good decision, because we end up with whole pixels being in/outside the sampling.

PR #458, opened 2023-02-03
Manipulators: when pressing enter, execute settings in window

dsizzle:
- on enter, read values of controls

Addresses #282 

I only did the main operations, no add-ons

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -939,6 +939,7 @@ ImageView::MessageReceived(BMessage* message)
 				if (WindowGUIManipulator* windowGuiManipulator =
 					dynamic_cast<WindowGUIManipulator*> (guiManipulator)) {
 					(void)windowGuiManipulator;	// suppress warning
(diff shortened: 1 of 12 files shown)


humdingerb:
Oh, I thought I could manipulate with the mouse and hit enter to apply when reading the commit message... :)

Anyway, there's something strange going on:
* Open an image.
* ALT+R to rotate.
* Rotate with mouse, say ca. -45°.
* Click into manipulator window and hit ENTER or click OK.

--> The mouse-set angle should be used, the manipulator window just be dismissed. Instead the image is rotated further.

dsizzle:
Is it just rotate?

humdingerb:
Yep, seems just to occur when rotating.

dsizzle:
Cool, thank you for testing. I thought my overall idea was sound, but after I saw your comment I remembered that I’d seen similar behavior when I did something else with rotate- snapping, perhaps?

dsizzle:
Ok, rotate seems to be fixed.

humdingerb:
Yep, works now, thanks!

PR #469, opened 2023-02-14
PaintWindow: gray out recent items that don't exist on disk

### LINKED ISSUE #461: Grey out/remove/error Recent Images that are invalid
1. open an image in ArtPaint
2. quit ArtPaint
3. move image to another folder
4. open ArtPaint, try to open it from Recent Images menu

...nothing happens. No error or any indication of a problem

One of the following should happen:
1. ArtPaint should show an error
2. images in Recent Images that can't be located should be grayed out
3. images in Recent Images that can't be located should not be shown in the menu at all

I lean towards option 2 because a user might've moved it inadvertently and it's friendlier to notice that rather than just being missing, but I'm open to opinions.

dsizzle:
Fixes #461


humdingerb:
Good!

PR #472, opened 2023-02-14
ImageView: accept Enter key when manipulators aren't focused

dsizzle:
Now ENTER should work to accept manipulators even when the window isn't focused, just like ESC does for cancelling.

Another piece of #282


humdingerb:
Fantastic!
Just a question: would it be possible to treat a double-click on the selection (or canvas if there's none) the same as a manipulator-confirming ENTER?
Merging this in any case.

PR #474, opened 2023-02-18
LayerView: identify layer by layout position instead of screen coordinates

### LINKED ISSUE #473: Investigate layer reordering
While trying to exercise the latest code by creating an actual image, I ran into odd behavior with layer reordering. Behaviors I witnessed were:

- Layers that I can’t reorder at all
- Layers moving up when I tried to move them down
- Layers reappearing in the wrong place after I moved them

interestingly the behavior seemed to persist even after restarting ArtPaint (at least with the immovable layer) so perhaps there’s something corrupted in the layer data.

Will investigate more and see if there’s an easy repro case

dsizzle:
when moving layers, the layer we want to swap with is either exactly 1 position above or below the layer we're moving, so don't try to find the layer via screen coordinates - just look at the layer above or below in the layout.

Fixes #473 

Layer reordering seems to work nicely with this fix. I do like how QuickLaunch works, but it uses features of BListView and reworking the layer window to use BListView is not ideal right this minute. The BListItems seem to be explicitly drawn rather than layouts, and so that would be quite different than the current LayerViews.  It might be nice to figure out a way to give some feedback while dragging but for now this fixes the egregious bugs in layer dragging as far as I can tell.

### DIFF
--- a/artpaint/layers/LayerView.cpp
+++ b/artpaint/layers/LayerView.cpp
@@ -239,20 +239,14 @@ LayerView::ReorderViews()
 		if (parent_view != NULL) {
 			while (buttons) {
 				the_window->Lock();
+				BPoint thisPos = layout->GetViewPosition(this);
 				frame = this->ConvertToParent(Bounds());
-				frame.OffsetBy(0, -Bounds().Height());
-				if (frame.Contains(location) == TRUE) {
-					exchanged_view = the_window->FindView(
-							the_window->ConvertFromScreen(
-								parent_view->ConvertToScreen(location)));
+				frame.OffsetBy(0, -(Bounds().Height() + 5));
+				if (frame.Contains(location) == TRUE && thisPos.y > 1) {
+					exchanged_view = layout->ItemAt(0, thisPos.y - 1)->View();
+
 					if (exchanged_view != NULL) {
 		 				int32 exchangedIndex = layout->IndexOfView(exchanged_view);
-		 				// make sure the selected view is one of the layer views,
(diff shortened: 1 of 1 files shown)


humdingerb:
Seems to work fine, until you grab a layer and drag it way up over the top of the list of layers, toward the "Mode" pop-up.
Kaboom: 
[ArtPaint-4547-debug-19-02-2023-05-29-32.report.txt](https://github.com/HaikuArchives/ArtPaint/files/10776031/ArtPaint-4547-debug-19-02-2023-05-29-32.report.txt)

dsizzle:
Well why in the world would you do that? 😄

dsizzle:
should be fixed now.  It was silly to not check if it was going off the end of the list.  doesn't seem to have the same problem at the bottom of the list for some reason.

humdingerb:
Crash is fixed. There's something else though, much as I hate to ruin your day...
Not sure if that was also the issue before the crash-fix.

Create a load of layers (say 10) and make the Layers window so small that not all layer items fit in the scroll view. Now try moving a layer from the middle of the stack to the top or bottom of the list (outside the visible part). The scroll view doesn't scroll with the dragged layer, so you'll move it 'blindly'. If you scroll manually after dropping it, you'll find it further up/down the list, possibly even at the right position.

dsizzle:
That doesn’t ruin my day because I’m sure it didn’t work properly before this change. It’d be good to fix but I think this change can be merged without it.

humdingerb:
So be it!

PR #475, opened 2023-02-20
LayerWindow: when layers are moved out of frame, scroll view

dsizzle:
How about this?  It seemed to work ok with my testing, but you are better at breaking stuff than me. :smile:


humdingerb:
It does the job, but isn't yet super nice. Look at this capture:


https://user-images.githubusercontent.com/6107158/220033365-e75f9fb0-45bf-413f-a43d-eda18b4a57b6.mp4

I save myself trying to describe it. Just look at it. :)

It'd be perfect, if while you're moving the layer down, there'd always be the layer below fully visible. (Same for moving up, ofc). Thereby you don't drag 'blindly':

no-yes

Not sure I make myself clear...

humdingerb:
That screencapture may not have made it... Try here: http://0x0.st/HsTY.mpg
 (temporary storage)
And in case you could use a test project file with 16 layers: 
[16+1_Layers.zip](https://github.com/HaikuArchives/ArtPaint/files/10780515/16%2B1_Layers.zip)

[BTW, it'd be nice it ArtPaint would use some compression for its project file format. That would shrink almost empty layers, like here the file is 17 MiB compared to the zipped 155 KiB.]

dsizzle:
Wow, that is pretty terrible.  It doesn't work that way with my test file at all - the dragged layer is always fully visible.  

anyway I was trying to do it as you illustrated, but it wasn't working. I'll give it another try because I agree that would be a lot better.

dsizzle:
regarding the file size, to be honest it's not something I have paid attention to at all, thanks for the idea. It sounds like it'd be nice to compress on save and decompress on load.

humdingerb:
Yeah, I bet it's a head-scratcher... :}
When testing stuff, keep in mind that font-size may play a role, too, and check with smaller and larger sizes.

dsizzle:
Having spend hours on this, I don't think it's solvable in the way you are envisioning. I shall attempt to illustrate. 

So if you drag Layer 14 downwards, like so:
laya

then the list scrolls up one entry. But then the problem is the mouse pointer is still in the same place, but layer 14 is where the red box is here:

laya02

so the list scrolls under the mouse, and thus triggers a second scroll.  In order to stop scrolling, you have to release the mouse button quickly. It requires a very light touch to only scroll one spot, and if you aren't quick, you'll scroll multiple spots or even very rapidly all the way down to the bottom.  

Does this make sense?  The scrolling can't move the mouse upwards/downwards (nor should it) so the mouse stays in the scrolly zone. If it were a keypress, like up/down arrow, that would be easy to scroll. But the dragging is moving the mouse pointer.

humdingerb:
Understood. I think that'll only work once there's the "Quicklaunch-like" layer dragging, i.e. with an indicator line where the layer will be inserted when dropped. Then the list of layers can be scrolled underneath the dragged layer and you can "overshoot" your desired drop-point and move the mouse-pointer back a bit to move the indicator line back a bit.

It can the work as does dragging a file in a crowded Tracker folder (only easier because of the 'indicator line' (which would be a nice racker addition)). Also, that Tracker scrolling is a PITA, because that "initiate-scroll-zone" seems to be only one or two pixels high...

Capture:  http://0x0.st/HsQJ.mp4

humdingerb:
> It can the work as does dragging a file in a crowded Tracker folder (only easier because of the 'indicator line' (which would be a nice Tracker addition)). Also, that Tracker scrolling is a PITA, because that "initiate-scroll-zone" seems to be only one or two pixels high...

I havetp take that back. First, an insert-indicator-line doesn't make much sense for Tracker, as the only use of dragging a file is to move/copy it into some folder, not putting it at some other place (as the list is sorted after an attribute (name, size, etc.).
Also, Tracker auto-scrolling isn't actually that much a PITA. The scrolling starts a fraction of a second after the mousepointer leaves the list view and hovers over the column headers (top) or the hor. scrollbar or info view (bottom).
That just as an aside... carry on.

dsizzle:
I gave it another try but if this doesn't work well enough then I'm inclined to just abandon this until I have time to explore the BListView solution.

humdingerb:
I think it's a slight improvement over the current master version. It does scroll way to fast, so you often end up at the top/bottom before you know it. But it's better than moving the layer blindly when the window refuses to scroll at all when moving up...
You decide, I merge.

dsizzle:
Yeah, go ahead and merge it, it’s still better than before.

I wonder if next version would benefit from a “move layer up “ and “move layer down” command in the layer menu.

but next version - don’t want to change any strings at this point.

humdingerb:
> I wonder if next version would benefit from a “move layer up “ and “move layer down” command in the layer menu.

Yes, or add icon buttons for duplicate, add, remove, merge, move up/down.
Having to use menu items for those few available commands is a bit tiresome.

Issue #403, opened 2022-11-19
Forward command messages to PaintWindow


dsizzle:
Currently the user can attempt to issue commands but nothing happens because the canvas is inactive.  Is it possible to forward menu commands from other windows to the last active canvas?

PR #478, opened 2023-02-23
LayerWindow: forward keypresses to canvas

dsizzle:
if key is pressed in layer window, activate image window and post the message to it.

Maybe fixes part of #403 ?  Only from the Layer window so far.

### DIFF
--- a/artpaint/layers/LayerWindow.cpp
+++ b/artpaint/layers/LayerWindow.cpp
@@ -332,6 +332,17 @@ LayerWindow::MessageReceived(BMessage *message)
 						image_window->PostMessage(message, image_view);
 				}
 			} break;
+		case B_KEY_DOWN:
+		case B_UNMAPPED_KEY_DOWN:
+			if (active_layer != NULL && active_layer->IsActive()) {
+				BView *image_view = (BView*)active_layer->GetImageView();
+				BWindow *image_window = image_view->Window();
+
+				if (image_window && active_layer->IsActive()) {
+					image_window->Activate();
+					image_window->PostMessage(message);
+				}
+			} break;
 		default:
 			BWindow::MessageReceived(message);
 			break;


humdingerb:
That does work for ALT keycombos, but it doesn't for single-letter-combos for changing the tool, e.g. 'b' for the brush tool. Could those non-modifier key presses be sent to the tool manager instead?

If we ever want menu shortcuts for the Layers window, can those be intercepted before activating and sending the message to the main window?

dsizzle:
> Could those non-modifier key presses be sent to the tool manager instead?

not sure about directly. They might need to go thru the image window and get passed along. I think the alt key is really what’s triggering this at the moment so it needs more investigation for a full solution.

> If we ever want menu shortcuts for the Layers window, can those be intercepted before activating and sending the message to the main window?

not sure here either, but if we can send keypresses to the main window successfully , do we need a menu in the layer window anymore?

humdingerb:
> not sure here either, but if we can send keypresses to the main window successfully , do we need a menu in the layer window anymore?

I don't see the direct connection between sending keypresses to main window and having a menu in the Layers window. We already have the same menu in the Layers window and main window. I think the reason is that if you're already mousing around on the Layers menu (making un/visible, reordering, renaming, setting modes and alpha), it'd be annoying having to move the mouse to the main window's Layer menu and back.

This could of course be avoided by adding icon buttons to the Layers window. Saves even a click by not having to open the menu... I do mention those a lot, don't I. :)

Let's merge this then. It's nice having commands forwarded!

dsizzle:
ok, I was digging into why the tool items don't get selected, and I remembered why - the letter keys are stopped on purpose, because they fire even when you are typing in the layer name boxes.  I need to see if I can detect if the user is typing a layer name before allowing those keys to update the tool.  Otherwise it's crazy when the tools change as you are typing.

humdingerb:
Maybe just do it on April fool's... :D

Issue #363, opened 2022-11-02
Respect alpha in selections


dsizzle:
If a selection contains an alpha value, use that when drawing into the selection. For example, an antialiased edge should stay antialiased. Or, if a brush stroke was made into a selection, its alpha edges should be respected as part of the selection. 

This will allow feathered or blurred edges of selections, as well as a step towards being able to paint selections.

PR #480, opened 2023-02-26
BitmapUtilities: make ConvertToMask only output exactly 1 or 0 alpha

### LINKED ISSUE #479: Invert selection bug
Not working after select-> all non selected ?

need to test more

dsizzle:
This fixes #479, but this change should be reverted when #363 is fixed.

I figured it out. It's not a bug, it's a feature!  The "select all non-transparent" uses the layer alpha to cut out the pieces that are transparent.  In my case here, the circle is not fully opaque, and so the resulting selection map is not fully opaque. We decide if something is selected or not based on whether it's exactly 0 or not. 

we need #363 to handle this properly and so for now/this release it's safest to just make "select all non-transparent" just take any non-opaque pixel.


humdingerb:
Ah, I didn't try with not fully opaque objects when I tried to reproduce earlier. Now I tried and can confirm that this change fixes it. Congrats!

PR #483, opened 2023-03-01
FillTool: refactor Check*Spans functions into one CheckSpans function

dsizzle:
- also did a bit of style cleanup, but not the whole file.

this is just a code refactor and shouldn't change anything - fill should work the same as before, no better or worse


dsizzle:
> Less undergrowth for these gradient bug(ger)s to hide.

exactly!

PR #485, opened 2023-03-02
Gradient fix

### LINKED ISSUE #353: Fill tool: Flood + gradient = chaos
If you set the fill tool to "Flood fill" and "Enable gradient" you an get interesting effect:

flood-gradient

Selections are ignored, it always messes up the whole layer.

dsizzle:
Please merge #484 first, this patch depends on it.

But, I believe this fixes the gradient chaos.  I can reproduce it pretty easily before this change, and afterward I can't seem to.  

Fixes #353

### DIFF
--- a/artpaint/tools/FillTool.cpp
+++ b/artpaint/tools/FillTool.cpp
@@ -519,6 +519,7 @@ FillTool::MakeBinaryMap(BitmapDrawer* drawer, int32 min_x, int32 max_x,
 	// This function makes a binary bitmap that has ones where the
 	// color of original bitmap is same as old_color, and zeroes elsewhere.
 	BBitmap* binary_map = new BBitmap(BRect(min_x, min_y, max_x, max_y), B_GRAY1);
+	memset(binary_map->Bits(), 0x00, binary_map->BitsLength());
 
 	uchar* binary_bits = (uchar*)binary_map->Bits();
 	int32 binary_bpr = binary_map->BytesPerRow();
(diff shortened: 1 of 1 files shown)


humdingerb:
Yay!  Seems to work perfectly now! Well, as perfect as the slightly weird gradient application allows... :)
(You noticed the merge conflict?)

BTW, I think we should remove the "Enable preview" option. Who wants to apply a gradient blindly?

dsizzle:
Darn. I thought merging the other pr first would avoid a merge conflict. I’ll fix it this evening. 

Re: enable preview, I’m also not sure what the flood fill option does. I can’t tell any difference. I thought one mode would be restricted only by selection and the other would be restricted by other pixels as well.

ie, if you draw a rectangle, and then click on the rectangle to fill, one mode fills the whole canvas and the other just fills just the rectangle. But I believe both flood the whole canvas at the moment (assuming no selection)

humdingerb:
There's a difference. From the documentation:

> Flood fill only colors objects that are connected, otherwise all similar pixels on the current layer get changed.

As an example, open the Walter logo and fill the "W" with Flood fill, then fill the "a" without Flood fill. Useful.

dsizzle:
ok, fixed the merge conflict.

I also played with the different settings - I had Tolerance set to 100 so both options were filling the whole canvas!

I don't think I've ever used a fill that works this way, but I must say once I understand it, it's pretty darn cool how it works.  

Like the flood + tolerance is like a fill version of the magic wand selection.
And the non-flood + tolerance is like Photoshop's "Replace Color" tool.  The fact that you can do this with basically one click is *really* fancy:

0waltgrad

humdingerb:
> Like the flood + tolerance is like a fill version of the magic wand selection.
And the non-flood + tolerance is like Photoshop's "Replace Color" tool. The fact that you can do this with basically one click is really fancy:

It is rather nice! :)
I think I'll put that as a further explanation into the documentation.

Issue #266, opened 2022-09-21
Remove context menu in canvas, draw bg colour instead


humdingerb:
Do many people use the context menu of the canvas to change tools? I kinda doubt it. The Tools window doesn't take much space and is probably always visible, if not even floating near where you paint. So mouse miles shouldn't be a big issue.

It makes more sense to use the right mouse button to use the selected tool with the background colour. That'd fit in with the usage of the RMB elsewhere in the app.

vansh-2266:
I want to take this issue can you please explain me this? @humdingerb

humdingerb:
Right now, if you right-click into the canvas, a context menu appears that lets you choose another tool. This ticket suggests to instead draw with the currently set backgroud colour (as opposed to the foreground colour when using the left mouse button).

Of course only for the tools where that makes sense.
Maybe for other tools different behaviour would be appropriate:
* Eraser: temporarily reverse the "Replace pixels with" setting (Background color <-> Transparency).
* Selection: temporarily reverse the "Behavior" setting (Add adrea <-> Subtract area).
* Color picker: used with the right mouse button sets the background colour.

dsizzle:
>  I propose moving the selection itself (not translating the seletcted pixels) with the right mouse button.

but what about rotating and scaling the selection? Do you think that is less common and so not having a shorthand method is acceptable?  I wouldn’t want to try to accomplish those via modifiers to the right click.

dsizzle:
> Eraser: temporarily reverse the "Replace pixels with" setting (Background color <-> Transparency).

the reason I didn’t do this one yet is that it felt a little off. Would anyone ever want to “erase” with the fg color if right click erases with the bg color?  And then that’s basically just painting.

humdingerb:
> but what about rotating and scaling the selection? 

I think we need all manipulation actions - Rotate, Translate, Scale - for the selections themselves, too. I see three possibilities:
1. Create Rotate, Translate, Scale menu items in the "Selections" menu, mirroring what's under "Edit"
2. Add a checkbox "Selection only" in all manipulator windows.
3. Add another submenu-item "Selection only" where there are now "All | Active layer" submenus of Rotate, Translate, Scale in the "Edit" menu.

> Do you think that is less common and so not having a shorthand method is acceptable?

At least for me, yes. I often use a selection to delete part of an image, then move it to the next part of the image, delete, and so on.
So a shortcut for moving a selection would be very welcome.

dsizzle:
[...]

PR #486, opened 2023-03-03
ImageView: remove context menu for right-click

dsizzle:
In all the color-based drawing tools, right-click draws with the bg color.

No change to Blur, Transparency, Eraser, Color Selector, or Selection Tool in this commit.

This is part of #266.


humdingerb:
I updated #266 with a proposal for the right-click with the Selection tool.

PR #488, opened 2023-03-04
Selection: added ability to translate, rotate, and scale selections

### LINKED ISSUE #432: Transform selections
Sometimes it's nice to take a selection and move / rotate / scale it - just the selection, NOT the image.  For example if you make a selection that is slightly off, you might want to just resize it or move it instead of undoing and redoing.  ArtPaint already has the tools to scale, rotate, and translate a selection under the hood, so this would just expose those functions to the user.

...tbh I can't think of obvious use cases that can't be accomplished other ways, but Photoshop has this ability, and as I mention, the functionality is already there.

dsizzle:
Fixes #432


humdingerb:
Nice!
Not sure about this, but do we want to have ReturnName() return e.g. "Rotate selection" if transform_selection_only == true?
I suppose, "ReturnHelpString()" should stay as is, to save space..

dsizzle:
> do we want to have ReturnName() return e.g. "Rotate selection" if transform_selection_only == true?

Yes, I think this is a good idea because it changes the window title for the manipulator. 

However, there's a bigger problem that I keep ignoring until I run into it: the way ArtPaint's undo system works is that it records pixel differences in layers. Since there are none here, you can't undo any of these. It's the same problem with the Layer window changes (reordering, changing blend modes, renaming).

humdingerb:
> I guess leave this in limbo until I figure that out?

You mean the whole PR?
I dunno, same as with the layer nundoablity, I think modifying selections is more useful than a missing undo is bad.

dsizzle:
ok, I added the string changes for ReturnName().  Feel free to merge if you think this is fine without undo.  But maybe I can try to do the undo stuff before v2.6 release.

dsizzle:
actually scratch that, I just added undo for this - it occurred to me that selection changes work fine, and this was just a selection change, so I hacked in some code if the manipulator name is "Rotate selection", "Translate selection", or "Scale selection" (localized of course) and it seems to work fine.

humdingerb:
Yes. If they aren't used in the menus - the only place a "..." indicating a window will open when invoked - they can be eliminated.

PR #491, opened 2023-03-09
Transform selections: remove ellipsis from names, disable/enable menu items

dsizzle:
Removed ellipsis from the ReturnName() functions for the manipulators.

When there's no selection, disable the menu items!  Forgot about this in the last PR.


humdingerb:
Thanks

Issue #240, opened 2022-08-22
Add all maipulations done in Layers window to undo-history


humdingerb:
All actions done in the Layers window should be undoable:
* Re-arranging via dragging
* Setting alpha
* Setting blending mode
* Changing visibility via checkbox
* Possibly also renaming a layer (that one is a bit more meta, as it doesn't change the canvas)

dsizzle:
So the current issue is that the Undo system stores changes to the layer bitmaps only.  All of these changes don't actually change any pixels, so the undo system needs to be able to handle metadata as well as bitmap changes.  

Fun side effect; if you rename a layer, delete it, and then undo, the layer will return with the default "Layer x" name (because the *bitmap* is saved and then the layer is recreated)

PR #492, opened 2023-03-19
LayerWindow: add ability to undo/redo layer 'meta-data' changes

dsizzle:
- supports undo/redo of visibility, blend mode, transparency, and layer name.

Partially addresses #240 

The remaining piece from the ticket is layer re-arranging.  However, another area that should be improved is deleting a layer.  Currently all layer settings are lost when you delete and then undo.

### DIFF
--- a/artpaint/application/UndoEvent.cpp
+++ b/artpaint/application/UndoEvent.cpp
@@ -30,6 +30,7 @@ UndoEvent::UndoEvent(const BString& name, const BBitmap*)
 	previous_event = NULL;
 
 	selection_data = NULL;
+	layer_data = NULL;
 }
 
 
@@ -44,6 +45,8 @@ UndoEvent::~UndoEvent()
 	delete[] actions;
 
 	delete selection_data;
+
+	delete layer_data;
 }
 
 
@@ -101,4 +104,16 @@ void UndoEvent::SetSelectionData(const SelectionData *s)
 {
 	delete selection_data;
 	selection_data = new SelectionData(s);
+}
+
+
+void
+UndoEvent::SetLayerData(Layer* src_layer)
+{
+	layer_data = new Layer(BRect(0, 0, 1, 1), src_layer->Id(), NULL);
+
+	layer_data->SetName(src_layer->ReturnLayerName());
+	layer_data->SetVisibility(src_layer->IsVisible());
+	layer_data->SetTransparency(src_layer->GetOldTransparency());
+	layer_data->SetBlendMode(src_layer->GetBlendMode());
 }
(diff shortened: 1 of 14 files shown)


dsizzle:
actually I got layer data working for undo, but not redo.  that is to say:

- delete a layer
- undo delete layer -> layer name, blend mode etc. is restored
- redo delete layer
- undo delete layer -> layer data gone

but it's better than before.

humdingerb:
Nice! I gave it a little test run. Here's one thing I found:

* Open ducky and Walter_the_OS
* Copy Walter as a new layer on top of ducky
* Change Walter's layer to "Color burn"
* Change Walter layer's alpha to 30%
* Delete Walter layer
* Undo "Delete layer"
-->The alpha is back at 100%. And the alpha slider has no effect, until you click the Walter (or any other layer, and back to the Walter layer) in the Layers window.

dsizzle:
I figured out the problem, not sure of the fix yet.  A difference with this change is that when you un-delete a layer, it gets selected.  Previously undoing a layer deletion kept the current selection (which I think is wrong). Seems like some percentage of the time, the layer isn't really selected until you click on it, so the UI doesn't work.  Hmm....

dsizzle:
ok. I *think* it's fixed by sort of forcing an update of the Layer window.

humdingerb:
Hmm... doing the same as described above, still the transparency isn't restored after undoing the "Delete layer".

dsizzle:
> doing the same as described above, still the transparency isn't restored after undoing the "Delete layer".

but does everything else work?

humdingerb:
Not completely. If I do this:
* duplicate layer
* change mode
* change alpha
* delete layer

then undo one by one:
* layer gets undeleted
  --> however: alpha = 100%
* After that, undo works as expected

then re-do everything.
then undo one by one again:
* layer gets undeleted
 --> however: mode= normal, alpha 100%
* next undo says: "Undo transparency"
--> however: not alpha gets changed, but mode
* After that, undo works as expected

It's a bit complicated... best try yourself, do these steps from above:
* duplicate layer
* change mode
* change alpha
* delete layer
then undo and redo and undo again. first undo wors (aside from transparency), second undo is different than the first one. Complicated.

dsizzle:
So the transparency part is a problem. But I mentioned above that undo/redo/undo/redo doesn’t work. But undo/redo should.

dsizzle:
Ok, I fixed the transparency for undoing a deleted layer.  But still, undo/redo/undo/redo doesn't work, as before.  

I had forgotten I needed to do a bit of a hack for transparency because it's a live updating slider.  So I only want an Undo event when you stop sliding, but I need to update the layer transparency every time the slider moves.  That means I need to keep track of what the original transparency was before the user started messing with it.  I think it's good now.

PR #494, opened 2023-03-19
ScaleManipulator: register undo when selection/layer is translated

### LINKED ISSUE #493: Using Scale to just move doesn't register in undo buffer
You can use Scale to keep the size at 100%, but move the pixels. That apparently doesn't produce an entry in the undo buffer. Should it say "Undo Scale" or can it recognize the "scale" was actually just a "translate"?

dsizzle:
Fixes #493 

I didn't update canvas resize, because a pure translate in that case really is nothing.

### DIFF
--- a/artpaint/viewmanipulators/ScaleManipulator.cpp
+++ b/artpaint/viewmanipulators/ScaleManipulator.cpp
@@ -168,8 +168,13 @@ BBitmap* ScaleManipulator::ManipulateBitmap(ManipulatorSettings *set,
 	// Remember that the returned bitmap must accept views
 	// First scale the width.
 	// If the new size is the same as old return the original
-	if ((new_width == starting_width) && (new_height == starting_height))
+	if (new_width == starting_width && new_height == starting_height &&
+		new_settings->left == original_left &&
+		new_settings->top == original_top)
 		return NULL;
+
+	if (new_width == starting_width && new_height == starting_height)
+		return preview_bitmap;
 
 	BMessage progress_message = BMessage(B_UPDATE_STATUS_BAR);
 	progress_message.AddFloat("delta",0.0);


humdingerb:
Something's wrong. One time it crashed:
[ArtPaint-15093-debug-19-03-2023-19-02-11.report.txt](https://github.com/HaikuArchives/ArtPaint/files/11012040/ArtPaint-15093-debug-19-03-2023-19-02-11.report.txt)

Plus:
* Move something with Scale  to the left. RETURN.
     -> Undo resets the position alright, but adds some part of the image at the top
* Move something with Scale  to the right. RETURN.
    -> Part of the image is repeated on the left side.

dsizzle:
is this better?

humdingerb:
I'm afraid not...
See for example: 

https://user-images.githubusercontent.com/6107158/226387868-0e5a5f21-1e66-44b0-8399-95cca71d18fa.mp4

Hope that clip works...

Easy to reproduce:
* ALT+E
* Move down and left, partially outside the canvas
* ALT+Z

dsizzle:
The clip doesn't work.  I also can't reproduce:

ALT+E, move down and left, partially outside canvas:
01trans

Undo (or ALT+Z):
03trans

No issue:
02trans

Same thing with a selection.  Can you try uploading a different video?

humdingerb:
Terribly sorry, I seem to have not used the latest build when trying yesterday.
Or it only works on odd days... :}
Anyway, the fix is good. And we merge...

dsizzle:
hahaha, no worries at all.  It wouldn't be the first time I couldn't reproduce something.  Also sometimes these "easy" ones turn out to be quite difficult - so I don't judge.  Thanks!

PR #499, opened 2023-03-31
Brush: change Brush painting to not use the "diff" brushes

### LINKED ISSUE #498: Tweaking Brush setup can crash
As mentioned in PR #497.

Choose the Brush tool and start first moving the Aspect slider to something other than 0.0. The in/decrease the Size until it crashes. The report: 
[ArtPaint-2365-debug-29-03-2023-16-11-33.report.txt](https://github.com/HaikuArchives/ArtPaint/files/11103040/ArtPaint-2365-debug-29-03-2023-16-11-33.report.txt)
And here when compiling a debug version: 
[ArtPaint_debug-1544-debug-29-03-2023-16-09-36.report.txt](https://github.com/HaikuArchives/ArtPaint/files/11103046/ArtPaint_debug-1544-debug-29-03-2023-16-09-36.report.txt)

Having Slayer open, I see memory consumption going up. If you only move the Size slider, it tops at 26.8 MB and doesn't crash. Setting Aspect first to e.g. -3 tops the memory when sliding Size at 206.2 MB. The slider gets more unresponsive the larger the Size gets.

This is all with just changing the brush setup, no drawing is done.

dsizzle:
- also remove all the diff brush code
- fixed brush ratio calculation so that "size" is the long axis - this probably also contributed to the crashes because the brushes were HUGE - a "Size" of 500px meant a brush of 5000 x 500 pixels * 4 bytes, then times 9 extra "diff" brushes = almost 100 MB for a single brush - so moving the slider meant throwing around hundreds of MB of memory
- fixed bug where size slider ignored the sign of the ratio

Fixes #498


dsizzle:
please lmk if the painting behaves differently than before.  It seemed to work fine in my testing.

humdingerb:
I have compared current master with pr499:

master:
master

pr499:
pr499

The bigger objects are default brush (size 30, aspect 0), the small objects at the top-rigt and left bottom corners are size 60, aspect -2.
The aspect -2 lines appear smaller in pr499, maybe that's expected with these changes.
Weird though, that the blue and yellow colours are darker in pr499.
I mean, in the great scheme of things, it doesn't really matter I guess. Though, going forward we may want to avoid changing those aspects. If ArtPaint gains more regular users, they'll over time have projects they want to use as templates for brushes and colours. Those should not change in the future.

Otherwise a nice optimization! No massive delayes with large un-round brushes anymore. :)

dsizzle:
Thank you for testing. The color difference *does* matter greatly in this case because I was trying to keep it the same - so this means my math is still wrong.

dsizzle:
I guess I tested with the wrong colors.  This is a brush, airbrush, and freehand and they are close enough:
00blue2

but with a different shade of blue they are not:
00blue

these should all match, so I see the brush still needs work.

dsizzle:
hmm, i am seeing some crashes when exiting ArtPaint. Do you get those too?  Not sure if it's somehow related to this change but it seems unrelated.

dsizzle:
ok, I've narrowed it down somewhat. It seems to only happen when the Brushes window is open when you start ArtPaint. 

1. open the Brushes window then close ArtPaint
2. reopen ArtPaint
3. open a recent project or image
4. close project
5. close 'new canvas' window (to quit ArtPaint)

... crashes on exit.  If you open ArtPaint with the Brushes window closed, it seems to not crash.

dsizzle:
I think #500 is enough to fix the crash when tweaking brushes, so please merge that one until I figure this out. :(

dsizzle:
I think I may have figured this out... I think I can fix it tomorrow.

dsizzle:
Ok, I think the crashing is fixed.  Whew.  Please test again when you have a moment and then I think this change is good to go.

humdingerb:
I still get a crash when I have settings from master-ArtPaint around: 
[ArtPaint_debug-4981-debug-02-04-2023-06-10-34.report.txt](https://github.com/HaikuArchives/ArtPaint/files/11131188/ArtPaint_debug-4981-debug-02-04-2023-06-10-34.report.txt)

After deleting /boot/home/config/settings/ArtPaint, no more crashes.
Here is a zip of ArtPaint master settings for testing: 
[ArtPaint_master_settings.zip](https://github.com/HaikuArchives/ArtPaint/files/11131195/ArtPaint_master_settings.zip)

dsizzle:
Ah, I was afraid of that. I should just undo the settings part of it; it’s not really needed for this patch.

dsizzle:
Ok, I fixed it to not accept invalid brush settings.  It worked with the attached settings file.

PR #501, opened 2023-04-05
BrushStore: move brush editor from Brush tool settings to store window

### LINKED ISSUE #202: Show current brush in main window and use if for all suitable tools
Thinking more about #177.

What about this: Similar to the current colour, ArtPaint shows the current brush. 
Mockup:

show-current-brush

This brush (ellipse|rectangle/size/aspect/angle/hardness) will be used for suitable tools: brush, blur, transparency, eraser (and in future other tools like lighten/darken, smudge, stencil, whathaveyou).
This will remove having that configuration (ellipse|rectangle/size/aspect/angle/hardness) for all tools,  which currently only have a Size slider and could do with different brushes. (BTW, is there a better word for "Brushes"? Once it's used for different tools, something more generic would be nice. "Tool tips", "Tool tops"?)

A left-click opens what's currently the Brush tool setup.
A right-click opens the Brushes window with stored settings.

dsizzle:
- fix/wire up brush preview in status bar
- remove brush tool
- add option to use current brush in freehand tool

part of #202

EDIT: also realized I should gray out the "Size" slider when "Use current brush" is selected.

### DIFF
--- a/artpaint/paintwindow/PaintWindow.cpp
+++ b/artpaint/paintwindow/PaintWindow.cpp
@@ -365,6 +365,9 @@ PaintWindow::CreatePaintWindow(BBitmap* bitmap, const char* fileName,
 					paintWindow->AddImageView();
 				}
 			}
+
+			CurrentBrushView::SendMessageToAll(HS_BRUSH_CHANGED);
+
 			return paintWindow;
 		}
 	}
--- a/artpaint/paintwindow/StatusView.cpp
+++ b/artpaint/paintwindow/StatusView.cpp
@@ -13,6 +13,7 @@
 #include "StatusView.h"
 
 #include "BitmapUtilities.h"
+#include "BrushStoreWindow.h"
 #include "ColorPalette.h"
 #include "PaintApplication.h"
 #include "MessageConstants.h"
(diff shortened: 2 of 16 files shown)


humdingerb:
> also realized I should gray out the "Size" slider when "Use current brush" is selected.

Yes. Either that, or use both option as radio buttons instead of the checkmark.

One issue I found: If you "Use current brush" and select a saved brush, nothing seems to be drawn. Only if you "Reset brush", stuff gets drawn.

Possibly difficult to do, since it's a layouting detail, but could the brush preview frame in the status bar merge at the top and bottom with the status bar border? Otherwise it appears like a thick stroke:

screenshot1

Nice change! But Oh noes!, now we either need to come up with another tool or remove a second tool, so I can have two columns without a tool-gap... :)

dsizzle:
> One issue I found: If you "Use current brush" and select a saved brush, nothing seems to be drawn. Only if you "Reset brush", stuff gets drawn.

oops. You are correct.  In fact any change to the brush (not just 'reset') made drawing work.  Anyway, I fixed the issue and pushed it. 

>  but could the brush preview frame in the status bar merge at the top and bottom with the status bar border?

you and that magnifying glass again.  I think I can resize the preview - not sure if I can make it larger to blend with the upper/lower edges, but I can probably make it smaller so there's at least a little border?

humdingerb:
> you and that magnifying glass again.

Heh. One of my pet peeves. :) Fat lines kinda stand out to me even non-magnified. I just include a Mag-shot to see it clearer and to be able to count pixels.

> I think I can resize the preview - not sure if I can make it larger to blend with the upper/lower edges, but I can probably make it smaller so there's at least a little border?

Yeah, maybe aligned to the colour swatches beside it, so 3 pixels.

dsizzle:
on my screen it was only 2 pixels, so I made it match the color swatch.  There's a bit of a visual difference since the color swatch has a white border, but it does match size-wise now.

humdingerb:
Very good! Merge!

Issue #222, opened 2022-06-30
Brush tool needs a "pressure" setting


dsizzle:
Trying to use ArtPaint a bit just for fun and testing, and it gets tedious constantly adjusting the alpha of a color.  It'd be much nicer to not touch the color but instead just tweak the pressure (or whatever we call it) to adjust the opacity. That also allows you to change colors and keep the same opacity between colors.

I think this is similar with other tools; adjusting opacity is often done in the tool rather than the color.  This might also be a stepping stone for pen/tablet support anyway?

PR #506, opened 2023-04-08
EraserTool: added "use current brush" functionality

dsizzle:
- also updated FreeLine tool to disable size slider when "use current brush" is checked.

More for #202 

still to-do is straight line, blur, transparency

### DIFF
--- a/artpaint/tools/EraserTool.cpp
+++ b/artpaint/tools/EraserTool.cpp
@@ -13,17 +13,22 @@
 #include "EraserTool.h"
 
 #include "BitmapDrawer.h"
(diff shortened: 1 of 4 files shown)


dsizzle:
also, incidentally, I improved "erase to bg color" with alpha colors in this version. 

master:
01bg

this PR:
00bg

humdingerb:
Very nice! Finally erasers with a soft edge.
Giving it a quick spin, I was thinking about a missing "pressure" setting, which would be similar to Airbrush's "flow" setting. There's #222 tracking this for the brushes (and mentions touchpads for this, which would indeed be nice).

PR #507, opened 2023-04-16
StraightLineTool: added "use current brush" option

dsizzle:
also made line drawing "live" so you can see the actual line as you draw it with the appropriate color (or brush).  There can be a slight artifact sometimes with the "adjustable width" option

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -219,10 +219,16 @@ BitmapUtilities::ClearBitmap(BBitmap* bitmap, uint32 color, BRect* area)
 	int32 start_y = 0;
 
 	if (area) {
-		width = area->IntegerWidth()+1;
-		height = area->IntegerHeight()+1;
-		start_x = (int32)area->left;
-		start_y = (int32)area->top;
+		BRect tmpArea = *area & bitmap->Bounds();
+
+		start_x = tmpArea.left;
+		start_y = tmpArea.top;
+
+		if (start_x > width || start_y > height)
+			return;
+
+		width = min_c(width, tmpArea.IntegerWidth()+1);
+		height = min_c(height, tmpArea.IntegerHeight()+1);
 	}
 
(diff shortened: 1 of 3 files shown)


humdingerb:
> There can be a slight artifact sometimes with the "adjustable width" option

That's the line becoming a triangle sometimes, right?
Seems to happen when the the adjustable width crosses from ca. +1 pixel to -1 pixel, if you nkow what I mean. Possibly when it's 0? If you click at that stage, it actually draws a triangle.

I notice that the preview-line when using "Adjustable width" is drawn with the disabled slider's width. If that's rather large, say 30 pixels, and you MouseUp the first time to set the endpoint, the line width jumps back to 1pixel when you start to adjust the line width.
Wouldn't it be nicer to always do the preview with 1 or 2 pixel width when we're in "Adjustable width"-mode?

Most important though: Things are going haywire when the zoom level isn't 100%... :)

dsizzle:
This is why I didn't do straight line right away...it's different enough to be a pain.  

Also I haven't been feeling well for the last week...I prob shouldn't code when I'm ill. :laughing: 

Anyway, latest update works great unless you use the scroll wheel *while* adjusting the size.  :frowning_face:

humdingerb:
Feverish coding can have interesting results... :)

Yes, changing zoom while adjusting width wrecks havoc. Same with ALT+/-, of course.
As usual, you tell me when we're ready to merge (the first iteration). The "don't zoom while adjusting width" may be a limitation we could live with...

dsizzle:
btw, I discovered the master version doesn't play nice if you zoom while resizing either. :(

humdingerb:
See, nobody noticed. Well, that or people learned what happens when they zoom in the middle of line-drawing, but didn't report it... :)

dsizzle:
ok, I think I have defeated the artifacts and haywires(?) - please give it a try when you have a chance.

humdingerb:
Very nice, works perfectly now! Thanks!

dsizzle:
Part of #202

PR #508, opened 2023-04-22
StraightLine: fix position of rotated preview line when zooming

dsizzle:
until I can figure out #507, this should at least fix master's current broken resizing straight line when zooming.


humdingerb:
That works!

PR #509, opened 2023-04-29
Transparency Tool: use slider instead of alpha from current color

dsizzle:
- also changed "speed" to pressure and made it a percentage of the way to the chosen value.

addresses transparency part of #176

### DIFF
--- a/artpaint/tools/TransparencyTool.cpp
+++ b/artpaint/tools/TransparencyTool.cpp
@@ -39,11 +39,12 @@ TransparencyTool::TransparencyTool()
 		TRANSPARENCY_TOOL)
 {
 	// The pressure option controls the speed of transparency change.
-	fOptions = SIZE_OPTION | PRESSURE_OPTION;
-	fOptionsCount = 2;
+	fOptions = SIZE_OPTION | PRESSURE_OPTION | TRANSPARENCY_OPTION;
+	fOptionsCount = 3;
 
-	SetOption(SIZE_OPTION,1);
-	SetOption(PRESSURE_OPTION,1);
+	SetOption(SIZE_OPTION, 1);
+	SetOption(PRESSURE_OPTION, 1);
+	SetOption(TRANSPARENCY_OPTION, 1);
 }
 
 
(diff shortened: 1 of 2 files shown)


humdingerb:
Nice change. I see two noteworthy things.

1. Using the tool on transparent parts of an image turns them black.
* Put Ducky.png on layer 1
* Put Walter.png on layer 2
* Set tool to e.g. size:75 / transparency:60 / pressure:100
* On layer2, draw over the logo into the transparent parts of Walter and Ducky:



Drawing on the transparent pixels of Walter appears dark, on Ducky it's bright.
The lower your transparency slider setting, the starker the dark/bright.
Is that expected?

2. Low pressure never reaches 100% transparency
* Set transparency:100 and pressure:10
* Keep clicking on an area, it will never reach 100% transparency. Shouldn't it eventually?

dsizzle:
> Using the tool on transparent parts of an image turns them black.

the tool ONLY changes the alpha value and so it’ll always be some color. On Walter it must be that the transparent parts are RGBA = 0, 0, 0, 0 and on ducky they are 255, 255, 255, 0.  Not sure anything can be done about it because the tool needs to be able to work on fully transparent colors just like any other color. 

> Low pressure never reaches 100% transparency

This is a bug. I think I introduced this by trying to slow down the pressure.

dsizzle:
ok, the low-pressure-never-reaching-full-transparency-or-opacity has been fixed.

humdingerb:
> the tool ONLY changes the alpha value and so it’ll always be some color. On Walter it must be that the transparent parts are RGBA = 0, 0, 0, 0 and on ducky they are 255, 255, 255, 0. Not sure anything can be done about it because the tool needs to be able to work on fully transparent colors just like any other color.

I see, I haven't thought about it that way... That's problem of only having 0-255 available per channel, or a -1 could signal "no pixel". :)

Anyway, working now. Merge. Thanks!

dsizzle:
The problem with the idea of a “non pixel” is that then for any operation you also need to decide whether to affect them or not.  Also, how/when do you make a non pixel into a pixel and vice versa? It would complicate the code unnecessarily - the proper workflow is to use a selection to avoid affecting pixels you don’t want to affect.

humdingerb:
Deep philosophical questions... To be a pixel, or non-pixel, that is the question. Where's my skull again...? :)

Issue #177, opened 2022-06-01
Tools: Use stored brushes for more tools


humdingerb:
Since we can configure brushes and store them for later use, it'd be nice to re-use those brushes for other tools.
Candidates are: ~~freehand-line and straight-line, rectangle and ellipse (if they are outlines, not filled)~~, blur, eraser

A checkbox at the top, "Use current brush" with a tool tip "Stored in the Brushes window". When set, related sliders Size or Width need to be disabled.

To always work, we have to make sure a default brush is always available in the Brushes window. That'd be similar to color sets.

The main work is probably to have those tools actually use those stored brushes...

dsizzle:
As mentioned before, freehand line is equivalent to a brush stroke already. Making it use the brushes would be even more redundant. I vote to just eliminate freehand line.

For straight line, it might make more sense to add a modifier or mode to brush to draw straight lines.

PeteCA:
On Thu, Jun 02, 2022 at 12:26:47AM -0700, humdinger wrote:
> +1 on eliminating the freehand tool.
> 
> Moving straight lines into the brush tool, I'm not so sure about.
I think I agree with the above.  Straight lines are something one does a lot.
Probably justifies a separate tool.
> 
> Though one day, it would be nice to be able to draw with any shape (from a selected part of the image or predefined stencils etc.) and then using those brushes with different tools may be nice...

I've often missed the feature in Amiga's old Deluxepaint, where a brush could
be any image.  If you painted with it, it would essentially stamp multiple copies
along the path, depending on how fast you moved, but the effect could be
quite impressive.

humdingerb:
> I've often missed the feature in Amiga's old Deluxepaint[...]

I think I remember that, too. It's been a long time... :)

dsizzle:
Lol, I used DP III or IV on DOS (I never had an Amiga) back in the 90s. Great program.

I think Photoshop also lets you make a selection into a brush. The brushes in ArtPaint are just monochrome with alpha right now, but it could either be extended to use colored brushes, or a separate "stamp" kind of tool.

humdingerb:
> it could either be extended to use colored brushes, or a separate "stamp" kind of tool.

Yes, I use the stamp tool all the time (in Photoshop at work) to copy one part of an image a bit further down in the image. As I move the mouse, the source follows my mouse pointer. Harder to put into words than expected... :)

Anyway, this got me thinking about what direction you want to go with ArtPaint.
Please comment at #194.

[...]

PR #511, opened 2023-05-03
BlurTool: added "use current brush" option

dsizzle:
so, let me know what you think:
this successfully uses the *shape* of the brush, but not the softness/hardness of it.  Since it's blurring, it's all soft...but I can't figure out (so far) how to make the soft parts of a brush "less blurry" - does that matter in this case?  

if so I will continue to work on it. 

if not, I *think* this is the last tool that makes sense for #202 and #177 ?


dsizzle:
actually I discovered really buggy behavior switching back and forth between using the brush and not:

ugh-blur

so, don't merge this. :laughing:

humdingerb:
> this successfully uses the shape of the brush, but not the softness/hardness of it. Since it's blurring, it's all soft...but I can't figure out (so far) how to make the soft parts of a brush "less blurry" - does that matter in this case?

Probably not. The checkmark label could be "Use current brush (size & shape)" to be perfectly honest with the user. :)

dsizzle:
Ok, i think I fixed the crazy bad artifacts - please give it a try.

humdingerb:
Works when "using current brush". Yay!
Makes a mess when using the explicit Size-controlled tool tip. Oh noes!

dsizzle:
:disappointed: 

I thought I had fixed it because I switched back and forth many times with no problems.

dsizzle:
huh, seems like changing the size is what breaks it.  weird.  I must've broken the old code path.

dsizzle:
Ok, *now* I think this is really fixed. After poring over and over the code for hours and this being the "smoking gun" between working and not, I hope this is it. It's an integer-vs-floating point issue.

humdingerb:
The old integer vs. float conflict... why can't they just get along? :)
Seems your peace negotiations were successful, or at least an armistice was reached. Blur tool appears to be in full working order. March!^W Merge!

[I'm in quite a good mood for a Monday, it appears...]

PR #512, opened 2023-05-04
Eraser Tool: added "Pressure" slider to control opacity of erasure

### LINKED ISSUE #176: Replace Transparency tool with options for Eraser

dsizzle:
Blur tool was giving me fits so I took a break to do this.

Fixes #176 
 
(this was the last piece to close out that ticket)

### DIFF
--- a/artpaint/tools/EraserTool.cpp
+++ b/artpaint/tools/EraserTool.cpp
@@ -47,12 +47,13 @@ EraserTool::EraserTool()
 	: DrawingTool(B_TRANSLATE("Eraser tool"),
 		ERASER_TOOL)
 {
-	fOptions = SIZE_OPTION | MODE_OPTION | USE_BRUSH_OPTION;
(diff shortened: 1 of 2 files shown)


humdingerb:
I know that the "use current" checkbox is kinda linked to the "Size" slider, but it'd look neater to have the sliders together and have the checkbox below them. That the checkbox only influences the Size not the Pressure setting, should be clear by only the Size slider being en/disabled when clicking the checkbox.

dsizzle:
Yeah I was wondering about that but went for the logical grouping. Happy to make this change if it looks better.

dsizzle:
fixed!

humdingerb:
Very good!

PR #513, opened 2023-05-04
FreeLineTool: added pressure setting to control opacity

### LINKED ISSUE #222: Brush tool needs a "pressure" setting

dsizzle:
Fixes #222 

tablet support will be a later thing...

### DIFF
--- a/artpaint/tools/FreeLineTool.cpp
+++ b/artpaint/tools/FreeLineTool.cpp
@@ -47,11 +47,12 @@ FreeLineTool::FreeLineTool()
 	: DrawingTool(B_TRANSLATE("Freehand line tool"),
 		FREE_LINE_TOOL)
 {
-	fOptions = SIZE_OPTION | USE_BRUSH_OPTION;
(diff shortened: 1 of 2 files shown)


humdingerb:
Same suggestions as in #512 : 
> I know that the "use current" checkbox is kinda linked to the "Size" slider, but it'd look neater to have the sliders together and have the checkbox below them. That the checkbox only influences the Size not the Pressure setting, should be clear by only the Size slider being en/disabled when clicking the checkbox.

dsizzle:
fixed!

humdingerb:
Good!

PR #514, opened 2023-05-04
StraightLineTool: add pressure setting for opacity control

dsizzle:
...and one more with a Pressure option.  Not sure what else will need pressure...maybe Hairy Brush?

### DIFF
--- a/artpaint/tools/StraightLineTool.cpp
+++ b/artpaint/tools/StraightLineTool.cpp
@@ -52,13 +52,14 @@ StraightLineTool::StraightLineTool()
 		STRAIGHT_LINE_TOOL)
 {
 	fOptions = SIZE_OPTION | ANTI_ALIASING_LEVEL_OPTION | MODE_OPTION |
-		USE_BRUSH_OPTION;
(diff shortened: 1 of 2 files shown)


humdingerb:
Same suggestions as in #512 : 
> I know that the "use current" checkbox is kinda linked to the "Size" slider, but it'd look neater to have the sliders together and have the checkbox below them. That the checkbox only influences the Size not the Pressure setting, should be clear by only the Size slider being en/disabled when clicking the checkbox.

humdingerb:
> Not sure what else will need pressure...maybe Hairy Brush?

Yes. Most would probably link their tablet-pressure to the Size slider, but it's nice to have options.

dsizzle:
fixed

humdingerb:
Works well. I was wondering for a second, if the "Enable antialiasing" should also be available when using the "current brush". At least if it's a 100% hardness brush you end up with blocky edges.
But then I thought, if users wanted smoothness, they can adjust the hardness of the brush...
And so we merge.

PR #515, opened 2023-05-06
LayerWindow: adjust min width so that scrollbar isn't cut off.

dsizzle:
as the description says, minor tweak so scrollbar doesn't get cut off at the minimum window size.


humdingerb:
Fixed indeed, thanks!

PR #517, opened 2023-05-06
BrushStoreWindow: make blue selection outline thicker

### LINKED ISSUE #503: Brushes: make selected stored brush more visible
The fine blue frame isn't very visible.
Maybe do it similarly to the Tools window: light background for unselected, darker background for the selected brush.

dsizzle:
also fixed redrawing, and cleaned up the code a tiny bit

Fixes #503

### DIFF
--- a/artpaint/tools/Brush.cpp
+++ b/artpaint/tools/Brush.cpp
@@ -375,7 +375,6 @@ Brush::PreviewBrush(BBitmap *preview_bitmap)
 			color.bytes[1] = value;
 			color.bytes[2] = value;
 			color.bytes[3] = 0xFF;
-//			value = value << 24 | value << 16 | value << 8 | 0xFF;
 			*(bits + (top+y)*bpr + left+x) = color.word;
 		}
 	}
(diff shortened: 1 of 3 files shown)


humdingerb:
Have you tried other blue border width, too?

It went from 1-pixel (too small):

before

Right to 3-pixel:

pr

A bit too thick IMO, also cuts a bit into the red size-rule at the bottom-left of each brush square.

Would 2-pixel be acceptable? Mockup:

suggestion

Also, and you'll damn my magnifying retentiveness, the view borders don't meld into the window borders on the left, bottom, and the scrollbar to the right, so they appear too thick:

thick border

dsizzle:
Yeah, I tried 2 pixels and I think there is some floating point rounding or the previews are weird sizes, because it is slightly off - nothing I tried would get it to line up, whereas 3 “just works.”

and yes, I do wish you would misplace your Magnifying glass. 🤣

dsizzle:
Ok, I went with 1.5 pixels; it looks similar to the 2-pixel mockup, and I moved the red size bar up a little and shrank it from the right size so it centers better. I also added another pixel in-between brushes as they seemed too tightly-packed. Of course I fixed the offending 2-pixel borders on the scroll view as well.

humdingerb:
Very nice merging borders. The blue frame isn't pixel perfect, but I understand it's not easy and one doesn't really see it. :)

One thing, is the blue colour. I think it should be using the same colour constant as the Tools window (ToolButton.cpp): B_KEYBOARD_NAVIGATION_COLOR
However, that is actually a deprecated constant name and both should be changed to B_NAVIGATION_BASE_COLOR.

dsizzle:
ok, updated to use the standard ui color (at least for this window; will do the tool button in a different PR)

humdingerb:
Very good. Nice optical illusion BTW:  the blue appears darker in the Brushes window than in the Tools window. Magnify confirms it's the same 0,0,229 though. :)
Merge!

humdingerb:
Very good. Nice optical illusion BTW:  the blue appears darker in the Brushes window than in the Tools window. Magnify confirms it's the same 0,0,229 though. :)
Merge!

Issue #504, opened 2023-04-07
Brushes: remove selection indicator once brush settings change


humdingerb:
You click on a stored brush, then change size/apect/etc. That's doesn't change the selected brush, but creates a new "current brush" (which could be stored as a new brush).
The selection indicator of the stored brush should therefore be removed.

As collateral damage, the menu item "Delete selected brush" would then need to be disabled, too.

dsizzle:
pulled from comments on PR #518 - 

> "Reset brush" should remove any selection from the stored brushes (except if you have stored a brush with the "Reset" settings, then it should be selected).

dsizzle:
what if you change the settings and it matches an existing brush?  Should that be selected?  I wonder if checking all the brushes every time you change a setting could be slow, if you have a large number of brushes.

humdingerb:
Come to think of it, there's really not much use in having a stored bush selected if you hit upon identical settings.
I wonder though how best to deal with the user trying to store the brush, if there's already one with identical settings.
Say, there are 12 stored brushes, the user clicks "Store brush" and expects a 13th to appear. Instead brush 7 is selected. Is that intuitive or surprising?

PR #518, opened 2023-05-06
BrushStoreWindow: don't allow duplicate brush adding

### LINKED ISSUE #502: Don't store identical brushes
If there's already a brush stored with identical settings, don't store an additional copy.

Additionall, the existing brush could be moved to the end of the list of brushes; the user may expect newly stored brush to appear at the end of the list.
Or, just move the selection to the already existing brush. (BTW, the selection isn't very visible. I'll create another ticket...)

dsizzle:
also fixed selection of brush on deletion, as well as moving between brushes with left/right arrow keys.

fixes #502

### DIFF
--- a/artpaint/tools/Brush.cpp
+++ b/artpaint/tools/Brush.cpp
@@ -27,6 +27,19 @@
 #define PI M_PI
 
 
+bool Brush::compare_brushes(brush_info one, brush_info two)
+{
+	if (one.shape == two.shape &&
+		one.width == two.width &&
+		one.height == two.height &&
+		one.angle == two.angle &&
+		one.hardness == two.hardness)
+		return true;
+
+	return false;
+}
+
+
 Brush::Brush(brush_info &info)
 {
 	// First record the data for the brush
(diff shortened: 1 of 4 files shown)


humdingerb:
Nice touch with the selecting with cursor left/right!

Some things though:
1. Deleting the last remaining brush will crash: 
[ArtPaint-1424-debug-07-05-2023-06-25-29.report.txt](https://github.com/HaikuArchives/ArtPaint/files/11414082/ArtPaint-1424-debug-07-05-2023-06-25-29.report.txt)
2. Do this:
* Store a few brushes
* (You may need to select some other brush than the first, first)
* Select the first brush
* "Delete selected brush" until none (or one, beware above crash)
--> There's an outline left from a deleted brush-square:
frme-left
4. "Store brush" should immediately select that just stored brush.
3. "Reset brush" should remove any selection from the stored brushes (except if you have stored a brush with the "Reset" settings, then it should be selected).

dsizzle:
items 1-4 should be fixed.

> "Reset brush" should remove any selection from the stored brushes (except if you have stored a brush with the "Reset" settings, then it should be selected).

I'd rather do this as part of #504 , not this change.

humdingerb:
Nicely fixed!

PR #522, opened 2023-05-13
BrushEditor: make brush preview size match brush store pane

### LINKED ISSUE #521: Stored brushes panes are too large
If you create a brush like the below and store it, the created stored-brush-pane doesn't look like in the preview at the top-left. I guess that is because the stored-brush-pane is a bit larger than the preview pane. Also compare their red scale-lines. I think preview- and stored-panes should be identical.

brushes

dsizzle:
Okay, they should be the same size now. I even tried to verify with Magnify. 

I actually made the brush preview larger rather than decreasing the size of the stored brushes, unless you feel strongly the other way.

Fixes #521

### DIFF
--- a/artpaint/tools/BrushEditor.cpp
+++ b/artpaint/tools/BrushEditor.cpp
@@ -58,10 +58,10 @@ BrushEditor::BrushEditor(Brush* brush)
 	, fBrush(brush)
 	, fBrushInfo(fBrush->GetInfo())
 {
-	float height = BRUSH_PREVIEW_HEIGHT - kExtraEdge;
-	float width = BRUSH_PREVIEW_WIDTH - kExtraEdge;
+	float height = BRUSH_PREVIEW_HEIGHT + 2;
+	float width = BRUSH_PREVIEW_WIDTH + 2;
 
-	fBrushView = new BrushView(BRect(kExtraEdge, kExtraEdge, width, height),
+	fBrushView = new BrushView(BRect(0, 0, width, height),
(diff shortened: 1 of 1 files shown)


humdingerb:
OK for the slightly larger preview panel. Seems to have improved the ellipse/rectangle radio button layout, too.

On thing though, I made a short capture: https://0x0.st/HNBz.mpg
Plays back fine in Haiku...

Basically, set your brush to this:
screenshot2

and slowly move the angle slider. The scale will flip back and forth making the brush appear larger/smaller while it rotates.

dsizzle:
I’m not by my computer right now so I can’t try it - does that happen only after this change, or is it not related?

humdingerb:
No, you're right, it's on master, too. Seems to be a general issue we didn't test for before...
Merge this and create a new ticket?

dsizzle:
That would be my preference.

humdingerb:
Alright Will do.

dsizzle:
btw, I watched the video, and it's not really a bug as such - the thing is that when you rotate the brush, at certain angles it's wider/taller than the current preview.  The way the code works is that it scales down by multiples of 2 when the brush width or height exceeds the bounds of the preview square.  To "fix" it would require scaling down the preview smaller than it needs to be, just so that the extreme angles would always fit.  I'm not sure that's a positive.

humdingerb:
I'll add your comment to the newly created #523. Let's continue discussions there.

PR #526, opened 2023-05-21
Selection: Moore's neighborhood walk replacement for SimplifySelection

dsizzle:
- factored out rasterization so that it's more generic and could possibly be used elsewhere.
- moved PointContainer class to UtilityClasses
- added "GetPixel" functions to BitmapUtilities to get the value of a pixel in any bitmap (only supports GRAY8 and RGBA32 for now)

### DIFF
--- a/artpaint/Utilities/BitmapUtilities.cpp
+++ b/artpaint/Utilities/BitmapUtilities.cpp
@@ -288,3 +288,36 @@ BitmapUtilities::CheckerBitmap(BBitmap* bitmap,
 		bits += row_size;
 	}
 }
+
+
+uint32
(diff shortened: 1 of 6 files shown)


dsizzle:
it's not perfect but imho it's categorically better than master, unless I haven't found bugs. A few comparisons; sorry, the screenshots don't exactly match but you can see the difference:

master:
001b-sel

this PR:
001sel

master:
003b-sel

this PR (not perfect, but better):
003sel

master:
004b-sel

this PR:
002sel

humdingerb:
Nice!
Getting these things right isn't something one expects to be such a challenge, but they are fundamental. Good work!

Issue #182, opened 2022-06-02
Fill tool: Gradient drawing is awkward


humdingerb:
I'd expect that the first click into the canvas sets the left hand colour from the setup window, the second click will set the right hand colour. Between those points is the gradient.

Currently it seems the first click sets the right hand colour, and the second click - depending how far away from the first click - sets the right hand colour again, and some point after the second click is a compressed gradient towards the left-hand colour.
Weird. I say this is a bug, because no twisted mind could have come up with that... :)

Also, a connecting line from the startpoint when the mouse is moving to find the endpoint would be nice.

dsizzle:
I also don’t think the gradient handles alpha properly. I tried to do a solid color on one end and pure transparent on the other, and the pure transparent end came out black.

humdingerb:
Total opposite here. :) Solid blue at the left end, total transparent to the right. As described in this ticket - when clicking on the left side, move mouse to the right and let go - the blue appears at the right. What should be transparent is white, not black as it's for you... weird.

humdingerb:
All done! \o/

PR #528, opened 2023-05-25
FillTool: make gradient range track mouse (aka intuitive)

dsizzle:
- also added Radial gradient option

Alpha still doesn't work, but aside from alpha I think this is much better.  I also added Radial gradients.

...as an aside, I tried to make the "linear" and "radial" buttons as well as the "preview" checkbox disable when you turn off gradient fill, but for some reason it totally breaks the functionality?  

Anyway please try out the gradient drawing and lmk if it is more intuitive.

the related issue is #182

### DIFF
--- a/artpaint/tools/FillTool.cpp
+++ b/artpaint/tools/FillTool.cpp
@@ -13,6 +13,7 @@
 #include "FillTool.h"
 
 #include "BitmapDrawer.h"
+#include "BitmapUtilities.h"
 #include "ColorPalette.h"
 #include "ColorView.h"
 #include "Cursors.h"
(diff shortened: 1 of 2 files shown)


humdingerb:
It works much better than before!
However, it still applies the colours the wrong way round, I think:
* You have a gradient going from black to white
* You click in the canvas and move the mouse to the right
* The result is a gradient from white to black

dsizzle:
well, I flipped the colors.  Still no luck on alpha.

dsizzle:
baby steps. Alpha works, but only if you don't use Preview mode.

humdingerb:
Cool! You'll get there. :)
You say when to merge the status quo. It's already much nicer than before.

dsizzle:
Ok, I think it all works now - please give it a try!  If it works fine for you then please merge.

humdingerb:
Those were quite huge baby steps... :)
Appears to work perfectly! \o/

PR #530, opened 2023-05-27
FillTool: draw line from start point to end of gradient

dsizzle:
Added a line from the gradient start point to the current point as mentioned in #182 

With this I think we can close that ticket. :smile:


humdingerb:
Good!

PR #532, opened 2023-05-28
ImageView: unhide selection when selection is cleared

### LINKED ISSUE #490: Unhide selection border when ALT+D (select none)
Sometimes, I like to hide the selection border, make sure all is OK and perform the action I intended. Then I unselect-all and continue painting or whatever. The next time I want makea selection, nothing appears to happen, because the border is still hidden...
I think automatically unhinding the border when there's no selection anymore is the right thing to do.

dsizzle:
Fixes #490


humdingerb:
Very good!

PR #533, opened 2023-05-28
ColorSelectorTool: don't change color palette until mouse up

### LINKED ISSUE #253: Don't update Colors panel live when using colour picker tool
When using the colour picker tool, the Colors window gets updated live, the fg/bg colour wells of the main window does not. IMO, since the colour picker tool already has a live preview, the Colors window should also only get updated on mouse-up.

dsizzle:
- also change color palette on single click

Fixes #253


humdingerb:
Thanks!

Issue #128, opened 2022-04-11
Tools with a "size" setting should draw a circle showing the size


dsizzle:
...especially since there's already a setting for "crosshair cursor" vs "tool cursor", any tool with a size setting should show what that size looks like on the canvas for the "tool cursor" option.

dsizzle:
Just putting a note here in case I forget: 
Can we leverage Selection::SimplifySelection() to generate a polygon representation of any arbitrary brush shape? That’d be much nicer than just a circle or square showing the size.

humdingerb:
There's are outline artifacts when testing #536, best seen when disabling "Use current brush" and a size of around 50:

![ezgif-5-ccf49ecd2b](https://github.com/HaikuArchives/ArtPaint/assets/6107158/7826c0e4-f6b2-45af-ba6b-3f7eb88fdb5a)

dsizzle:
@humdingerb I think this ticket can be closed since we have #545 for the Airbrush spray - I didn't think there was any other work for this ticket?

humdingerb:
Indeed!

PR #535, opened 2023-05-29
ImageView: draw ellipse indicating rough size of brush

dsizzle:
This is a first version for #128 -  I think it works pretty well.  My concern was performance since it's drawing on every single mouse movement, but it seems to be fine. 

Works for Freeline,  Airbrush, Eraser, Blur, and Transparency.  I didn't do Hairy brush because the "size" of it is dependent on how quickly you move the brush. 

Eventually I want to try using the new selection algorithm to see if I can draw the actual shape of a brush rather than just an ellipse.


dsizzle:
can't decide if this should have the option to turn off.  Maybe.  But it's so dang useful to actually know what you're going to paint...

humdingerb:
Very cool!

A few observations:
* It's a bit flickery and the outline disappears if you stop the mouse-movement, but keep the mouse-button pressed. Try moving very slowly, pixel-by-pixel to see the outline dis/re-appearing.
* Using a largish brush, on first click - without moving the mouse - you see one-pixel lines at the top, bottom, left, right of the outline ellipse.
* The straight line tool could show such a preview-outline if it's not in "Adjustable width" mode.
* If there should be a setting for this, the Misc tab of the settings still has room for a checkbox. I agree, though, it's very useful and can't come up with a situation you wouldn't want the brush-outline...

dsizzle:
I think the flickering and your second bullet are because it’s drawing when the mouse button is not pressed, and when the button is pressed the actual painting takes over which may fight with it. 

Not sure how I feel about the straight line one. Might make the most sense if “use current brush” is selected though.

in terms of the setting I was thinking like the cursor mode, there could be a third option for “show tool size”?

humdingerb:
> I think the flickering and your second bullet are because it’s drawing when the mouse button is not pressed, and when the button is pressed the actual painting takes over which may fight with it.

But once you are actually drawing, there's no need for the brush size outline. Maybe only draw it when there's no mouse-down?

> there could be a third option for “show tool size”?
Yep. The two existing radio buttons plus the new checkbox "Show tool size".

dsizzle:
ok, now it doesn't draw if there's a mouse button pressed.  

I tried adding this to StraightLine Tool and it felt a little weird. 

haven't done the option yet.

dsizzle:
...and now there's a setting in the Misc section of the app settings.

dsizzle:
Ok, I went for the grand prize and now it draws the actual shape of the brush!!!

humdingerb:
> Ok, I went for the grand prize and now it draws the actual shape of the brush!!!

Fantastic!
I guess using polygons is a throw forward to supporting arbitrarily shaped brushes. An angled rectangular brush appears a bit scraggy. Ellipses are fine. I suppose it's not trivial to optimize rectangular brush sizes to use a regular 4 point rectangle? It's no biggie though.

One thing: The brush size outline doesn't reappear after mouse-up, but only once you've moved the mouse.I think it should.

humdingerb:
Alright, I'm convinced. Ready to merge.

> How often is the mouse still?

Only after a very long day at  the rat race... :)

dsizzle:
Speaking of rat races I thought about this whole driving to the office, and I wonder if your comment actually nailed it:

> The brush size outline doesn't reappear after mouse-up

maybe if I add code to draw in MouseUp it would feel more correct?

humdingerb:
> maybe if I add code to draw in MouseUp it would feel more correct?

I'd say so. There are situations where you don't move the mouse, like you have the Blur or Transparency tool with a largish brush and park the mouse over a part of the image and just click the mouse button a few times until you're satisfied with the look. In that case you wouldn't see the brush size after the first click. Not a big issue, but if the brush size could easily be drawn at MouseUp as well, that'd be nice.

PR #536, opened 2023-06-01
ImageView: draw brush cursor on MouseUp

dsizzle:
per discussion in #535, draw the brush cursor on mouseup.

I think it does look better - lmk

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -1143,6 +1143,26 @@ ImageView::MouseDown(BPoint view_point)
 
 
 void
+ImageView::MouseUp(BPoint where)
+{
+	if (fManipulator == NULL) {
+		where.x = floor(where.x / getMagScale());
+		where.y = floor(where.y / getMagScale());
+
+		int32 mode = B_CONTROL_ON;
+		if (SettingsServer* server = SettingsServer::Instance()) {
+			BMessage settings;
+			server->GetApplicationSettings(&settings);
+			settings.FindInt32(skDrawBrushSizeMode, &mode);
+		}
+
+		if (mode == B_CONTROL_ON)
+			DrawBrush(where);
+	}
+}
+
+
+void
 ImageView::MouseMoved(BPoint where, uint32 transit, const BMessage* message)
 {
 	// here we will display the coordinates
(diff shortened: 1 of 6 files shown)


humdingerb:
Yes, very good. It doesn't work for Airbrush and Transparency though.
And there's still the outline artifact. Best seen when disabling "Use current brush" and a size of around 50:

ezgif-5-ccf49ecd2b


Myabe some rounding error? New ticket?

dsizzle:
Yeah, probably a rounding error.  You don’t need a new ticket; just add it to #128 which is still open.

odd that this pr doesn’t work for transparency and airbrush… the mouseup should be mouse up for anything??

humdingerb:
FWIW, the MouseUp() and the call to DrawBrush() is also triggered for these tools. I checked with printf()...

dsizzle:
another problem I see is that I think the "mouseup" cursor is offset incorrectly when zoom isn't 100%?

dsizzle:
ok, fixed the offset part

but yeah, for airbrush and transparency, you can see it 'flash' sometimes but seems to get erased immediately. odd. :eyes:

dsizzle:
this one is a real head-scratcher.  It seems to really be something particular to those tools... I did notice that those two tools seem to use a slightly different way of compositing/drawing/updating the view, so I might try making them behave more like the other drawing tools to see if it solves it.  :exploding_head:

dsizzle:
Fixed Transparency tool.  Not sure why my changes fixed it, but it seems to be fixed now.

Discovered another issue; if you change tools (to, say, the Selection tool) the brush outline stays drawn for a short bit. Maybe until you mouse over that area again.

humdingerb:
> Discovered another issue; if you change tools (to, say, the Selection tool) the brush outline stays drawn for a short bit. Maybe until you mouse over that area again.

Yes. Not only for a short bit. It stays until you use the tool and select the area with the overlayed brush outline. Or your mouse pointer leaves the canvas area.
I switch between tools via shortcut keys, "F" and "S".

dsizzle:
slow progress, this "simple" issue is a challenge.  "Airbrush" mode should work now, but "Spray" mode is just artifacty and weird. 

haven't tried to address the "brush-outline-that-won't-go-away" yet.

humdingerb:
> this "simple" issue is a challenge

It's always the on-a-quick glance simple issues that turn out frustratingly annoying.
Well, not always. Most of the time you know it'll be a bugger before starting it...:)

dsizzle:
well, I think I solved the brush outline not going away issue, but Spray tool is just... hating me.

humdingerb:
Very good. If you need a change of coding scenery, we can merge as is and you can decide to get back to it at a later time. It's not a showstopper or annoying...

dsizzle:
It’s mildly frustrating because it seems to be some kind of race condition between when the tool stops drawing and the brush outline draws - but it’s not clear why the brush wins the race in every situation except spray… I would expect it to be more unpredictable??

With that said, if everything else works ok I’m inclined to merge it.

humdingerb:
Alright. Let's merge while we wait on your sudden brain wave. :)
I created #545 where you can add further findings.

PR #537, opened 2023-06-01
Image/Layers: use zlib to compress image data for each layer

### LINKED ISSUE #481: Project file compression
As noted by humdingerb, the project files for ArtPaint can be rather large. This ticket is to investigate compressing the file format.

dsizzle:
This was easier than I thought because the file reading/writing already had a compression format (set to 0) written into the project files, and it was passed to the layer read/write.  

Anyway this uses zlib for the compression - would that need to be added to the recipe(?) as a dependency for ArtPaint? 

Seems to work fine with both compressed and uncompressed files, and the compressed file sizes are much nicer - equivalent to zipping the whole project. 

Fixes #481


humdingerb:
Cool! Reduced my 16 layers file from 17 MiB to 150 KiB. Mostly empty space in those though... :)

Does the file type need some sort of versioning? For future forward/backward/whatever compatibility? A compressed project file does crash older ArtPaints. Not a problem, I suppose, because who wants to not run the latest ad greatest release.

dsizzle:
Hmm maybe needs some versioning. Good point.

dsizzle:
ok, I moved the uncompressed size field down to the "extra" section of the file, and so now old versions should just skip over it instead of crashing.  The layers will be garbage but the file will open.  

NOTE: if you saved any files with the previous version of the PR, those will crash with this version (and old versions of ArtPaint).  You need to re-save them with this version because the file format changed.

humdingerb:
Alright. Let's merge this then!

I was just wondering if in the future, ArtPaint files may contain additional stuff (masking layers?) that older versions won't be able to deal with. ArtPaint could check a file's version and if it's newer  announce "_This is a filetype of a newer version of ArtPaint. Can't open it. Consider updating ArtPaint._"

The old ArtPaint format could be 1.0, this compressed version 2.0. If there's a new feature that still opens in older ArtPaint versions, increase to 2.1. The next version that would break an older ArtPaint would be 3.0 etc.

I leave it to you, if adding explicit versioning is useful.

dsizzle:
yeah, i do think there might come a time when nothing else will fit into this file format, and versioning might be needed.  For example, saved selections might not.  Layer masks probably will be fine, but there will come a day when something doesn't work. 

It's a little bit of a tricky problem because obviously there is no way to make older versions of ArtPaint detect the change; we need to say "from now on, ArtPaint will detect changed formats" but prior to that it needs to not break.  I'll add a ticket so at least it gets some mind-share.  

The clever thing about the current format is that, at least for layers, there's an "extra data" section.  it has a "start" and "end" marker, and a size of the data section.  A version of ArtPaint reads as much of the data that it can, and then skips to the end marker, ignoring any data it doesn't know about - so I was able to add stuff like layer names, blend modes, compression info, etc. and old versions just ignore it.  

the downside is that older versions will truly ignore it, so in the case of compression data it doesn't tell the user why the file looks like garbage when you open it.  So, maybe this is a good point to update the format.

PR #539, opened 2023-06-03
Add tool shortcut to ToolButton tool tip

humdingerb:
It's nice to have a hint to the shortcuts in the GUI, and not just the documentation.


humdingerb:
Would that be OK, @dsizzle ?

dsizzle:
yes, I love it.  I think this was a sorely-needed change, thank you @humdingerb !!

humdingerb:
Dunno if it were "sorely" needed, but it's nice to have. :)

PR #542, opened 2023-06-03
ManipulatorWindow: change window feel

humdingerb:
With B_FLOATING_APP_WINDOW_FEEL the window floats over all open project windows. So, if you for example, "Rotate" on Project1 and switch to Project2, the "Rotate" window still floats above it and may confuse the user.

B_FLOATING_SUBSET_WINDOW_FEEL will only float over the invoking project window and disappear when switching projects.


dsizzle:
Nice change! Looks good to me; definitely a better user experience.

humdingerb:
Alright. Originally, I wanted to have the text tool's text view accepting drag&dropped text, but it of course disappeared whenever I changed focus to the e.g. StyledEdit window. I found out that this behaviour did make sense for the other ManipulatorWindow uses. No matter, copy&paste is almost as fast.
Maybe you'll put the text tool through the code wringer at some point and it'll use a window more similar to the other tools' ToolSetup window, i.e. none-floating.

PR #550, opened 2023-06-24
BrushEditor: fixed math for brush ratio to properly reverse from stored value

### LINKED ISSUE #546: Brush aspect oddness when storing a brush
Configure a brush, e.g.:
Size: 50
Aspect: 5.00
Angle: 40
Hardness: 45

Store brush. See the Aspect change to -5.00

Now switch to some other brush, then back again to the just stored one. The aspect is now -5.00, the preview pane shows the same shape as when you've stored the brush (i.e. aspect 5.00).
Now move any slider a tiny amount. -> The preview jumps to show the actual aspect -5.00, i.e. 90° rotated from the 5.00 setting.

dsizzle:
fixes #546

as far as I can tell. :P


humdingerb:
Fixed!

PR #551, opened 2023-06-26
NumberSliderControl: for proportional mode, re-add the subtracted 1

### LINKED ISSUE #547: Brush size oddness when entering a value
If you enter a size value in the text box and hit ENTER, the size is applied correctly.
~~If you enter it and use TAB to move the focus to the slider, the value decreases by one.~~
If you enter it, press ENTER, and use TAB to move the focus to the slider, the value decreases by one.

That's unique to the Size slider, all others behave correctly.

dsizzle:
- in the exponential value calculation, 1 is subtracted from the numerator and denominator, and never added back in, so the slider is 1 less than it should be.

Fixes #547


humdingerb:
Indeed it does. Thanks.

Issue #520, opened 2023-05-09
New brush types


dsizzle:
This has been mentioned before but it'd be nice if ArtPaint could take a selection and turn it into a brush. 

Two types:
- grayscale bitmap - turns selection into a monochrome brush that can be used with any color like any other brush. 
e.g.:
![00brush](https://user-images.githubusercontent.com/10878750/237007677-a2a66745-b927-4703-b41c-72fa562caf7a.png)

- multicolor - uses selection *exactly* as-is, and cannot paint with any other colors, just whatever colors are in the brush.
e.g.:
![01brush](https://user-images.githubusercontent.com/10878750/237007777-c0ba33ca-dc38-402c-8a85-607a9fae3ed6.png)

PR #553, opened 2023-07-01
Brush: replace uint32 array with BBitmap to store brush

dsizzle:
- this should make it easier to store different types of brushes and removes the need for "BrushToBitmap" function.
aka prep for #520 

I think I tested everywhere brushes are used.


humdingerb:
Yep. Still seems to be working with all tools.

PR #558, opened 2023-07-02
Crop: removed all the +1s for bounds

### LINKED ISSUE #552: Crop: width and height 1 pixel off
As an example, get this image of the Deskbar: 
screenshot1
* Zoom into the top right corner
* Open "Crop..."
* Enter Top = -1 and Width =188 - which should add a pixel to the top and left
--> Works as expected for the top, but adds 2 pixels to the right
Opening the Crop dialog again, truthfully shows a width of 189.

Similarly when adding to the height:
* Enter Left = -1 and Height = 268 - which should add a pixel to the left and bottom
--> Works as expected for the left, but adds 2 pixels to the bottom
Opening the Crop dialog again, truthfully shows a height of 269.

An off-by-one somewhere.

dsizzle:
Seems to fix #552 

...but could it be that easy?  

I was trying to remember how it used to crash but I couldn't get it to crash even trying all sorts of weird sizes.

### DIFF
--- a/artpaint/viewmanipulators/CropManipulator.cpp
+++ b/artpaint/viewmanipulators/CropManipulator.cpp
@@ -59,9 +59,9 @@ CropManipulator::CropManipulator(BBitmap* bm)
 	if (bm != NULL) {
 		BRect rect = bm->Bounds();
 		settings->left = min_x = rect.left;
-		settings->right = max_x = rect.right + 1;
+		settings->right = max_x = rect.right;
 		settings->top = min_y = rect.top;
-		settings->bottom = max_y = rect.bottom + 1;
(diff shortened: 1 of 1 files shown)


humdingerb:
> ...but could it be that easy?

Of course not. :)
Height and Width are now reporting one pixel short. The Deskbar test image from #552 is 187 x 267, but ArtPaint thinks it's 186 x 266. While the procedure described in #552 works correctly now, cropping to 1 x 1 actually gives a 2 x 2...

I have a dim memory that this was an issue you fixed before, probably introucing this very same issue you're solving now. :)

dsizzle:
closing because this didn't fix the issue.

PR #560, opened 2023-07-04
PixelOperations: moved blend mode strings to LayerWindow.h

dsizzle:
this should fix the catkeys for addons


humdingerb:
It does! Thanks!

PR #563, opened 2023-07-04
Brush: lock brush bitmap bits before trying to draw preview brush

### LINKED ISSUE #562: Race condition(?) on startup causes crash
Crash is in Brush::PreviewBrush()

dsizzle:
Fixes #562


humdingerb:
I don't think I've seen that crash, but I feel much safer with that lock in place...

PR #566, opened 2023-07-07
FillTool: add ability to snap angle with SHIFT to 22.5 degree increments

dsizzle:
- also fixed a couple of points where the gradient would flip incorrectly at perfectly horizontal or vertical angles

... note that I didn't update the help string because I don't want to annoy the translators...


humdingerb:
Maybe I can fudge something by adding the translated "SHIFT locks 22,5°" from other tools before we release...

However, it would only be perfect, if the help strings were dependend on the selected tool option: Here, the 22,5° hint doesn't apply for non-gradient fills.
Probably not worth the added complication, I bet you agree. :)

PR #569, opened 2023-07-09
FillTool: properly account for scale when snapping graident angles

### LINKED ISSUE #568: Gradient angle snapping doesn't work with zoom other than 100%
I swear I tested this because it happens every time... but trying to paint and it was all haywire outside of 100%.

dsizzle:
Fixes #568


humdingerb:
Fixed! Obviously, I didn't test the change very well either...

PR #577, opened 2023-07-15
ImageView: pressing DEL while in a manip executes manip and then deletes

### LINKED ISSUE #570: Pressing DEL during selection transfomation gives unexpected results
Steps to reproduce:

1. Select something
2. Choose Selection->Translate  and move the selection, but do NOT accept the move
3. Press DELETE

Expected results: 
- nothing happens, or
- DELETE cancels transformation, or
- DELETE causes transform to be applied and then the deletion occurs

Actual results:
- Deletion occurs in previous position of selection, and transformation is still active:
selbug

dsizzle:
Fixes #570 

...but it works with ANY manipulator.  I think that's ok because I'm not sure why a user would do, say, a normal translate and then hit DEL.  But both actions are separately undoable.  I.e. if you translate and hit DEL before clicking OK, the translate happens, then the delete, and you can then undo just the delete.  

I feel like if a user does do such an action, this result is likely what they intended.

### DIFF
--- a/artpaint/paintwindow/ImageView.cpp
+++ b/artpaint/paintwindow/ImageView.cpp
@@ -317,6 +317,11 @@ ImageView::KeyDown(const char* bytes, int32 numBytes)
 				delta = bitmap_rect.bottom - bounds.bottom;
 			ScrollBy(0, delta);
 		} else if (*bytes == B_DELETE) {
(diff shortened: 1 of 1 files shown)


humdingerb:
Intuitively right behaviour.

PR #580, opened 2023-07-18
ColorPalette: disambiguate "Colors" for translation

### LINKED ISSUE #579: "Colors" string improvement
In `ColorPalette.cpp`, there are two instances of the string `Colors`. Since they are the same, Pootle only shows one translation, however, the menu string needs to be different from the title one (at least for Turkish).

The menu item should also not use string concatenation, which looks like the source of the error in the first place.

humdingerb:
Fixes #579.

### DIFF
--- a/artpaint/controls/ColorPalette.cpp
+++ b/artpaint/controls/ColorPalette.cpp
@@ -837,10 +837,11 @@ ColorPaletteWindow::openMenuBar()
 
 	// in this loop we add possible palette sizes to menu
 	BMessage *msg;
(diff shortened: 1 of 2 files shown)


dsizzle:
This change looks good to me. Ship it!

humdingerb:
Thought you might want to test your new merging powers... 😃

dsizzle:
I considered it but didn’t really have time to deal with it if anything went wrong… 😄

humdingerb:
Alright. Your time will come. :)

PR #581, opened 2023-07-19
Use BNumberFormat for percentage values

bitigchi:
Also refactor the zoom sub-menu generation logic.

### DIFF
--- a/artpaint/paintwindow/MagnificationView.cpp
+++ b/artpaint/paintwindow/MagnificationView.cpp
@@ -88,15 +88,17 @@ MagnificationView::MagnificationView()
 	B_PLAIN_BORDER)
 {
 	BFont font;
+	double percentValue = 1600.0;
 
-	char string[256];
-	sprintf(string,"%s: %.0f%%", B_TRANSLATE("Zoom"), 1600.0);
+	if (fNumberFormat.FormatPercent(fPercentData, percentValue) != B_OK)
+		fPercentData.SetToFormat("%.0f%%", percentValue);
 
-	fMagStringView = new MagStringView("magStringView", string);
+	fPercentString.SetToFormat("%s: %s", B_TRANSLATE("Zoom"),
+		fPercentData.String());
 
+	fMagStringView = new MagStringView("magStringView", fPercentString);
 	fMinusButton = new BButton("minusButton", "-",
 		new BMessage(HS_ZOOM_IMAGE_OUT));
-
 	fPlusButton = new BButton("plusButton", "+",
 		new BMessage(HS_ZOOM_IMAGE_IN));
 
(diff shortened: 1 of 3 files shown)


dsizzle:
Thanks for this change!  

Unfortunately the help text at the bottom of the window incorrectly displays the help text: 

percent

Also a minor nitpick, but the "1600%" now has the locale-based thousands separator so it shows as "1,600%" which feels odd for a zoom %.

bitigchi:
Ah, it used to work okay, but apparently broken in the final revision. Will fix the broken render.

For the 1,600; it just uses the system settings for thousands separator. I'm not sure if it's possible to disable it just yet, I'll look into it.

bitigchi:
It looks like disabling the thousands separator is not currently exposed in Haiku's ICU implementation (feel free to correct me). Previously 1600 was not included, I included it because the bottom slider went all the way up to 1600. Shall I remove it?

humdingerb:
Just for my education, what's the problem this solves?

bitigchi:
> Just for my education, what's the problem this solves?

It formats the percentages according to the locale settings. For instance:

- %100 for Turkish and some other Turkic languages
- 100 % for French and some Scandinavian folks
- 100 ٪ for Arabic etc.

Detailed info here: https://en.wikipedia.org/wiki/Percent_sign

bitigchi:
Well, that's the whole point of this PR, otherwise I'd just mark them translatable and be done with it.

I'd rather go ahead with this PR and make it rather an incentive to people to see it and complete the implementation in Haiku. Otherwise it may go unnoticed for a very long time.

I see this kind of zoom value in other applications anyway, but if you think that's an issue, fair enough.

humdingerb:
> Otherwise it may go unnoticed for a very long time.

I doubt many people study the ArtPaint source. A ticket at Trac should be more effective.

bitigchi:
It looks like I've already opened a ticket in the past, but have forgotten about it.

They don't need to see the source, they could see the menu entry or the slider value and try to fix it if they get confused by it.

dsizzle:
I think this change is a good change and makes the code cleaner. The comma separator doesn’t bother me that much, I just thought it might be an option to the formatter.

humdingerb:
If the 1,000 separator offends, we could add this at the appropriate places:

```
	if (magLevel >= 10.0) {
		BString seperator(fNumberFormat.GetSeparator(B_GROUPING_SEPARATOR));
		fPercentData.RemoveFirst(seperator);
	}
```

bitigchi:
Neat! I'll try this and report back.

humdingerb:
It might be worth to cache the seperator string in a member variable. At least when using the zoom slider, the percentage calculation can be triggered rapidly. (Or is that only done once when releasing the mouse?)

bitigchi:
I couldn't get it to work on both places with the appropriate changes. It would be great if someone could try it at a convenient time...

humdingerb:
I committed the changes via "Edit" of the GitHub website, because I'm not sure how to push stuff into this PR...
We'll squash those additional changes into one commit.
Only waiting for @dsizzle's blessing. :)  (Feel free to squash&commit)

bitigchi:
For the sub menu, my change was identical, probably I forgot to clear the cache... LGTM!

dsizzle:
Looks good to me!

dsizzle:
squashed and merged.

bitigchi:
Hooray!

humdingerb:
Teamwork! 😃
@dsizzle : next time you squash,  you could edit the commit message to remove ugly or redundant stuff.

PR #584, opened 2023-07-22
ColorUtilities: move colorspace conversions to one file

dsizzle:
No functional changes, just code moving.


humdingerb:
Miraculously still builds and works. :)

PR #589, opened 2023-07-30
StraightLineTool: set updated rect so undo works

### LINKED ISSUE #585: Straight line tool: No undo
The straight line tool can't be undone... I think it used to work. No idea when it broke.

dsizzle:
Fixes #585


humdingerb:
Confirmed fixed. Merge!

PR #590, opened 2023-07-30
Adding some test images

humdingerb:
What do you think about adding those images for testing purposes?
Do the colors.png and ducky.png have a free to use license?


dsizzle:
I don’t know the licensing for the colors and ducky but I’m sure we could find free images that are similar.

dsizzle:
Yep, Walter is fine.  The ducky was just convenient when I first implemented the blend modes because there was a web page that showed the results and it used the ducky and rainbow so it was easier to compare.  It's not too much work to put the test images in another app to test, though.  

Thanks for recreating the rainbow in WB!

PR #591, opened 2023-08-01
PaintWindow: make size window non-user-closeable.

dsizzle:
- Cmd+W was allowed to close size window and then ArtPaint crashed on exit. This change disallows closing the size window, which should've been done before because it would put the app in a bad state.


humdingerb:
Easy and effective. Merge!

PR #593, opened 2023-08-01
ColorFloatSlider: set fMinRange and fMaxRange in SetMinMax

### LINKED ISSUE #592: Hue slider in HSV color space reverts to 255
Steps to reproduce:

1. In Color window, move Hue slider past 255 to a value such as 337
2. Change color model to RGB
3. Change color model back to HSV

Expected results:
H slider stays at 337

Actual results:
H slider moves to 255, though Hue doesn't actually change to 255 (blue) until another value is changed

dsizzle:
Fixes #592


humdingerb:
Good bug find and fix.

PR #595, opened 2023-08-04
ColorSlider: Remove redundant/unused class methods

dsizzle:
Only code cleanup, no functional change


humdingerb:
Still builds and runs, code got leaner. Merge!

PR #597, opened 2023-08-13
Selection: respect alpha values for selections.

### LINKED ISSUE #363: Respect alpha in selections

dsizzle:
This is needed for features like blurred/feathered selection edges, and painted selections.

Fixes #363

This also allows you to do cool effects.  For example, draw a stroke or two with the Airbrush, then choose "Select all non-transparent" and you get something like so:

alpha-sel

...and then you can just do a gradient fill and it says within the exact shape of the selection because the alpha of the selection is preserved:

sel-alpha02


humdingerb:
That's really cool!
I tried to replicate your example, but for some reason it doesn't quite work. When I try to apply the gradient fill, the anty selection outline disappears (no CPU spike) and the fill tools gradient preview line stays on. Until I move the mouse pointer outside the canvas window. Then the outline re-apears and the preview-line is gone. No gradient though...

dsizzle:
Hmm, I’ll try it some more tomorrow. I forgot one step I did in the middle, in case it makes a difference: I created a new layer and drew the gradient on the new layer.

humdingerb:
Indeed, adding a new layer and doing the grdient fill there makes all the difference.

dsizzle:
It actually also works for me without the extra layer, as long as my Threshold is not 0.  

But, the behavior you described seems to be a bug even without this change.   Filed #598 for it.

One note about this change is that it only works for "Select all non-transparent" selections... magic wand selections should also care about alpha, maybe?  Or maybe it should just be blurred or painted selections when we have those, and the current behavior is correct?

dsizzle:
hmm, I realize I need to change more stuff - selection data is saved as polygons, not the bitmaps - so alpha data is lost on undo/redo.

dsizzle:
Yep. I knew this was too easy.  

Filed #621 for a needed feature for this.  This is the gift that keeps on giving...

dsizzle:
ok, I think this is FINALLY done. @humdingerb if you have a chance, please test this - I believe I successfully tested all of these and they work:

1. Edit->Translate->Active layer
2. Edit->Translate->All layers
3. Selection->Translate
4. Edit->Rotate->Active layer
5. Edit->Rotate->All layers
6. Selection->Rotate
7. Edit->Scale->Active layer
8. Edit->Scale->All layers
9. Selection->Scale
10. Edit->Flip horizontally
11. Edit->Flip vertically

... as well as undoing and redoing.  please let me know if you find any problems. They should all "just work" as before.

humdingerb:
Looks good! \o/
Took you long enough... :) :) :)

dsizzle:
Thanks! Merged!

> Took you long enough... :) :) :)

You have no idea how sick of this change I am... :D

PR #599, opened 2023-08-14
ImageView: start selection drawing when select-all modes are used

### LINKED ISSUE #598: "Select all non-transparent" doesn't properly trigger selection outline drawing
When "Select all non-transparent" is used, the selection outline draws as dashed black-white "ants" but it doesn't animate.  This causes drawing issues.  

For example, if a user tries to draw a gradient fill inside a selection with a non-animating border, the selection border disappears until the mouse pointer is moved outside the canvas.

dsizzle:
this is for both "Select all" and "Select all non-transparent"

Fixes #598


humdingerb:
Looks good!

PR #600, opened 2023-08-26
ImageView/Image: Moved layer data copying to where layer structure is rebuilt

dsizzle:
This is a better fix for #240 I believe.  It fixes the undo-redo-undo-redo of deleted layers that didn't quite work before:

1. create a layer
2. rename it
3. delete it
4. undo delete layer, layer reappears with correct name
5. redo delete layer
6. undo delete layer, layer reappears with correct name <- this didn't work before this fix.  Now you can undo and redo and the layer properties come back 

Still missing from #240 is layer re-ordering, but at least I think this works better.


humdingerb:
Works indeed!

PR #606, opened 2023-09-04
ScaleUtilities: actually *use* y offset for vertical scaling.

### LINKED ISSUE #605: Scale shifting bug
Somehow this escaped the notice of the eagle-eyed ArtPaint QA team; I went back in history before 2.5, before transforming selections was a feature, and still saw this bug:

1. Load an image
2. Select all non-transparent
3. Open Scale manipulator and un-check "Lock"
4. Scale by /2 in Y direction
5. Preview looks fine, so hit Ok
6. Selection boundary is correct, but pixels shift down by a number of pixels when scale is applied and thus gets clipped by the selection boundary. 

walthappened

:crying_cat_face: 

Note: the issue is just the mismatch of the pixels and the selection boundary. The small version of the Walter logo is supposed o be there, that's not part of the bug.

dsizzle:
Fixes #605 

yeah, the code was passing in a y-offset but not using it for vertical scaling. The horizontal scaling code was correct.  I think it's been broken since I added different scaling filters. oops.

### DIFF
--- a/artpaint/Utilities/ScaleUtilities.cpp
+++ b/artpaint/Utilities/ScaleUtilities.cpp
@@ -90,12 +90,12 @@ ScaleUtilities::ScaleVertically(float width, float height, BPoint offset,
 	uint32* source_bits = (uint32*)source->Bits();
 	int32 source_bpr = source->BytesPerRow() / 4;
 
-	for (int32 y = 0; y <= height; y++) {
+	for (int32 y = 0; y <= (int32)height; y++) {
 		int32 low = floor(ratio * y);
 		int32 high = ceil(ratio * y);
-		float weight = ratio * y - low;
-		uint32* src_bits_low = source_bits + (int32)offset.x + low * source_bpr;
-		uint32* src_bits_high = source_bits + (int32)offset.x + high * source_bpr;
+		float weight = (ratio * y) - low;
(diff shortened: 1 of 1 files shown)


humdingerb:
Much better! There's still a little offset of about 1 pixel, it seems. Also, after the scaling, the selection border isn't completely flush anymore. There are also some pixels left from the original sized object. Best seen zoomed in, when doing what's described at #605.

walterYsized

dsizzle:
| There's still a little offset of about 1 pixel, it seems. Also, after the scaling, the selection border isn't completely flush anymore.

Yeah, I'm not sure what to do about that...I think it's more like "accumulated error"... if you look at the top it's perfectly flush, and then by the time it's at the bottom it's about 1 pixel off.   

|  There are also some pixels left from the original sized object.

I think this also happens without scaling (i.e. also with Translation and Rotation).  I wonder if the scaling I did to make the outline tighter should be undone.

dsizzle:
It seems like just rounding the offset up fixes the offset issue.  

The "left-behinds" is a general selection issue as I said - you can see that with Translation as well - unrelated to this change.

humdingerb:
Alright. The slight movement when invoking and confirming the Scale window seems to be from show/hiding the bounding box. Time to merge then...

PR #609, opened 2023-09-05
Selection: store selection map for undo/redo instead of polygonal data

### LINKED ISSUE #608: Scaling selections leaves behind artifacts
1. Load image such as Walter_the_OS.png
2. Select all non-transparent
3. Scale selection

scale-artifact

dsizzle:
Another in the chain of selection changes... this change is what I needed for #597, but I ran into the other issues along the way so I had to fix them first.

Anyway, what this change does is store the selection *bitmap* for undo and redo operations.  Previously the selection outlines were stored, not the bitmaps.  Using the bitmaps means we can store the grayscale (aka alpha) values for selections instead of just the shapes.  When blurred or painted selections are available, this will be critical. 

It'll use more memory but it's needed for more interesting functionality. 

I was also secretly hoping this might fix #608 and it seems to?

### DIFF
--- a/artpaint/application/Selection.cpp
+++ b/artpaint/application/Selection.cpp
@@ -301,6 +301,50 @@ Selection::AddSelection(BBitmap* bitmap, bool add_to_selection)
 
 
 void
+Selection::ReplaceSelection(BBitmap* bitmap)
+{
+	if (!bitmap)
+		return;
+
+	acquire_sem(selection_mutex);
+
+	if (original_selections) {
+		for (int32 i = 0; i < selection_data->SelectionCount(); ++i)
+			delete original_selections[i];
+		delete [] original_selections;
+		original_selections = NULL;
+	}
+
+	selection_bounds = BRect();
+
+	delete selection_map;
+	selection_map = new BBitmap(image_bounds, B_GRAY8, true);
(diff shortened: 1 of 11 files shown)


dsizzle:
Oh I should mention this is another of those changes that should have no visible effect, just need to make sure all selections and manipulations and undo/redo still work properly

dsizzle:
@korli I'm not sure. I’ve only seen that if I rollback git history and don’t do a clean build. I assume the auto build always starts clean, no?

korli:
> @korli I'm not sure. I’ve only seen that if I rollback git history and don’t do a clean build. I assume the auto build always starts clean, no?

Right. It seems it started with 5c75fafbba5b037ac41991b18a9071e58e5dfdc7

humdingerb:
> I was also secretly hoping this might fix https://github.com/HaikuArchives/ArtPaint/issues/608 and it seems to?

Looks like it. No obvious selection/undo regression AFAICS.

dsizzle:
> > I was also secretly hoping this might fix https://github.com/HaikuArchives/ArtPaint/issues/608 and it seems to?

> Looks like it. No obvious selection/undo regression AFAICS.

Excellent, thanks for testing - then I shall merge

dsizzle:
@korli Thanks for fixing the missing include paths for all the addons!

PR #610, opened 2023-09-06
ImageView: make sure to restart selection drawing after undo/redo

dsizzle:
Missed a spot with #598 / #599

to reproduce the problem:

1. load image
2. Select all non-transparent
3. Selection -> None (Alt + D)
4. Undo 
- note the ants don't crawl


humdingerb:
And with this change, the ants are on the march again. March!^W Merge!

PR #612, opened 2023-09-07
Big style clean-up

humdingerb:


dsizzle (changes requested):
Added several comments that apply in multiple places, like include order, deleting chunks of comments, deleting empty destructors.  

I'm only partway thru - this is a big diff ! :smile: 

It's upsetting to see how much duplicated code there is in the add-ons...

PR #613, opened 2023-09-07
Big style cleanup

humdingerb:
Had a bit of trouble resolving merge conflicts before, but after doing that at least 3 times, it seems to stick...

PR #615, opened 2023-09-11
WindowGUIManipulator: restore MessageReceived virtual function

dsizzle:
need to re-add this function after #613


dsizzle:
merged without review; needed to fix missing runtime symbol

humdingerb:
Sorry about that. I thought if the MessageReceived() just calls the parent class' (is that what you call it?), MessageReceived() function, it could just as well be removed...

dsizzle:
parent or base.  I can take the blame for this; I remember making a comment along the lines of being able to delete these types of functions.  Without going into detail, in this case it's the inheritance chain - the derived classes (like the AddOns) need that virtual function.

humdingerb:
Ah, sort of an inheritence tax we have to pay for the greedy add-on heirs...

PR #616, opened 2023-09-13
EllipseTool: Added rotate-able ellipses

### LINKED ISSUE #175: Ellipse tool: Add "Enable rotation" checkmark

dsizzle:
Fixes #175 

finally!

caveats:
- didn't add the "specify rotation via numerical value" mentioned in the ticket
- if you zoom *while* rotating, it's a little weird.  Rectangle has the same problem.  Solution: don't zoom while rotating. :laughing:


humdingerb:
Good change!

I'll let you merge your PRs from now on. Just in case you have some last minute (cosmetic) changes.

dsizzle:
Thanks! I actually did have a few; I was trying out CudaText instead of my usual Pe, and it seems to be using spaces instead of tabs. :/

humdingerb:
Koder user myself. But I didn't look for tab-vs-space...

Issue #618, opened 2023-09-14
Width oddity with freehand and line tool


humdingerb:
This is a zoomed in image of single freehand-tool dots with widths from 1 to 6 points (the red dots are just there to help count pixels):

![freehand_1-6](https://github.com/HaikuArchives/ArtPaint/assets/6107158/dd9a0dd7-b90c-473f-9227-aa2746c0b47d)

"Width" = 1: shows as 1 pixel width/height
"Width" = 2: shows as 1 pixel width/height
"Width" = 3: shows as 2 pixel width/height
"Width" = 4: shows as 3 pixel width/height
"Width" = 5: shows as 4 pixel width/height
"Width" = 6: shows as 5 pixel width/height

A one-off issue?

Same with the straight line tool:

![line_1-6pt](https://github.com/HaikuArchives/ArtPaint/assets/6107158/a94f86c2-a235-4d32-950a-9f05b6dd92cf)

"Width" = 1: shows as 1 pixel width/height
"Width" = 2: shows as 3 pixel width/height
"Width" = 3: shows as 3 pixel width/height
"Width" = 4: shows as 5 pixel width/height
"Width" = 5: shows as 5 pixel width/height
"Width" = 6: shows as 7 pixel width/height

A rounding issue?

dsizzle:
well, the code is completely different for those. For Freehand, it actually creates an Elliptical brush of (presumably) the correct size. I wonder if Brush sizes are just off by 1 pixel? 

For the Straight line tool it's supposed to be drawing a filled rectangle of the appropriate width. I'm guessing it's a floating-point rounding issue as you mention.

dsizzle:
Closing via #619 and #620

PR #619, opened 2023-09-16
Brush: add 1 to dimension of round brushes, which fixes Freehand tool widths

dsizzle:
- also use "ceil" instead of "floor" for rectangular brushes, and that seems to fix dimensions of square brushes.

addressed the "freehand" part of #618


humdingerb:
Yep, fixed!

dsizzle:
merged, thanks for both finding and testing!

PR #620, opened 2023-09-16
StraightLineTool: don't add 1 to even-numbered widths

dsizzle:
Addressed the "straight line" part of #618 

For some strange reason the width had 1 added to it if it was an even number?  No idea why but it seems to work fine without that.

### DIFF
--- a/artpaint/tools/BitmapDrawer.cpp
+++ b/artpaint/tools/BitmapDrawer.cpp
@@ -329,7 +329,7 @@ BitmapDrawer::_DrawShearedEllipse(BPoint center, float width, float height, uint
 			if (anti_alias == true) {
 				float error = x - fx;
 
-                color2.word = color;
+				color2.word = color;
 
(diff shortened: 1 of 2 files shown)


humdingerb:
Works worth width set in the straight line tool setup. If you "Use current brush" though you still get the same as described in #618::

"Width" = 1: shows as 1 pixel width/height
"Width" = 2: shows as 3 pixel width/height
"Width" = 3: shows as 3 pixel width/height
"Width" = 4: shows as 5 pixel width/height
"Width" = 5: shows as 5 pixel width/height
"Width" = 6: shows as 7 pixel width/height

The same happens with "Use current brush" for the freehand tool.

dsizzle:
You need to pull #619 to get the right behavior for the current brush

humdingerb:
Oops, sorry, I thought #619 was merged before you created this PR. All works as intended with #619 added.

dsizzle:
yay, merged.  Thanks!

PR #622, opened 2023-09-20
Clang fixes for 9/2023

dsizzle:
- fixed several memory leaks
- fixed bad deallocator
- fixed undefined values
- fixed NULL pointer accesses
- removed dead initialization

Related to #124


dsizzle:
merged directly without review since it's just static analysis fixes.

humdingerb:
FWIW, I did jus check if it still compiles and launches. The changes in the code seem to be mostly unused variables, duplication and NULL checks. All above the most trivial things, I wouldn't trust myself to judge... :)
Maybe I should just comment "Still builds" in the future. OTOH, I bet you already knew that...

PR #623, opened 2023-09-22
Add -gdwarf-3 to COMPILER_FLAGS for debug builds

humdingerb:
As recommended in https://dev.haiku-os.org/ticket/18590

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -127,7 +127,7 @@ case "$action" in
 		;;
 
 	debug)
-		debug="-e DEBUGGER=TRUE"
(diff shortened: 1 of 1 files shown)


humdingerb:
If there ever to appear compiler flags for regular builds, we need to add those in the build.sh line that's modified here. Or be more sophisticated and read the COMPILER_FLAGS from the makefile and add Thorin, Gimli and Balin (and possibly Oin and Gloin if dwarf4/5 is supported).

dsizzle:
merged. May the dwarf be with us.

PR #624, opened 2023-09-25
Selection: transform selections via bitmaps, not polygons

### LINKED ISSUE #621: View manipulators need to operate on Selection bitmaps
This is a superset of #525; perhaps I will close that one as a Duplicate but for now I will keep it.

Translation, Rotation, and Scale for Selections need to operate on the selection_maps as they do on standard bitmaps.  Currently they move the polygonal outlines, which they still should do for real-time updates, but when the manipulation is complete they need something like ManipulateBitmap that works on Selection bitmaps. Perhaps ManipulateBitmap can be made to handle both types of bitmaps, but it might be easier to duplicate the function as ManipulateSelection. 

This is yet another feature needed for #597, and future features like painted + blurred selections..

dsizzle:
Fixes #621 

Not sure how to test this one other than just making sure the translate, rotate, and scale selections work at least as well as before.

I tested with rectangle, ellipse, freehand, and "select all non-transparent pixels" selections, both on new, blank canvases and existing pngs like Walter_the_OS, with undoing and redoing and such.  

Filed #625 for the one issue I found that is also present in master before this change.

### DIFF
--- a/artpaint/Utilities/ScaleUtilities.cpp
+++ b/artpaint/Utilities/ScaleUtilities.cpp
@@ -160,6 +160,56 @@ ScaleUtilities::ScaleVertically(float width, float height, BPoint offset, BBitma
 
 
 void
+ScaleUtilities::ScaleHorizontallyGray(float width, float height, BPoint offset, BBitmap* source,
+	BBitmap* target, float ratio)
+{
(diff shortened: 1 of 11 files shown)


humdingerb:
Seems to work fine.

dsizzle:
merged! thanks!

PR #628, opened 2023-10-01
Tools: replace coordinate reading function with CoordinateReader class

### LINKED ISSUE #545: Brush size cursor doesn't work with Airbrush spray mode
Introduced with #536 drawing the brush size around the mouse cursor works for all tools perfectly except for Airbrush's spray mode. Here, once you stop drawing, the outline appears to be removed after MouseUp. dsizzle speculated in #536 

> ...it seems to be some kind of race condition between when the tool stops drawing and the brush outline draws - but it’s not clear why the brush wins the race in every situation except spray… I would expect it to be more unpredictable??

dsizzle:
... this makes all the drawing tools use one standard CoordinateReader instead of several having their own.  No user-visible change should occur; Airbrush, Freehand, Eraser, and Transparency should all behave as before.

I *think* this seems to fix #545 although I feel like I saw more brush-circle-where-drawing-starts artifacts.  Maybe that's ok for this change because it'll be more easily fixable with standardized code?

### DIFF
--- a/artpaint/tools/AirBrushTool.cpp
+++ b/artpaint/tools/AirBrushTool.cpp
@@ -74,24 +74,24 @@ AirBrushTool::UseTool(ImageView* view, uint32 buttons, BPoint point, BPoint)
 	while (LastUpdatedRect().IsValid())
 		snooze(50000);
 
-	coordinate_queue = new (std::nothrow) CoordinateQueue();
-	if (coordinate_queue == NULL)
-		return NULL;
-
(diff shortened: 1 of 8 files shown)


humdingerb:
I only see the brush-circle at the beginning of drawing with the Airbrush tool. But it id that before this change, too. Consolidating code is probably a good move in the right direction.

dsizzle:
yay, merged

Issue #174, opened 2022-05-31
Rectangle+Ellipse tool: Add outline width slider


humdingerb:
If you don't checkmark "Fill rectangle/ellipse", a 1 pixel rectangle outline is drawn. Similarly to the straight-line tool, we need an "Outline width" slider that is enabled if "Fill rectangle" is unchecked.
Also, a similar "Adjustable outline width" to interactively set the width.

dsizzle:
Idea, in case I forget: maybe try using compositing modes and "erase" the center hole using dst_out_fixed?

PR #629, opened 2023-10-02
Rectangle/Ellipse: Add "Outline width" slider

dsizzle:
Addresses #174 

... I'm not sure about the "Adjustable outline width" option.  Would that happen *after* rotation?  I didn't add that part for now. 

A few caveats; the preview drawing has some issues where it doesn't draw correctly, but the actual drawn shape is usually correct. For example: 

1. With large widths, a rotating rectangle gets the corners cut off, but when you click to accept it, the rectangle draws properly with no cutoff corners. 
2. With large widths and small rectangles, there's a diamond-shape drawn in the center - seems like maybe a bug with Haiku's PenSize, where if it's larger than the shape you're drawing there are artifacts?

(1) might also be weird Haiku view drawing behavior?

### DIFF
--- a/artpaint/tools/DrawingTool.cpp
+++ b/artpaint/tools/DrawingTool.cpp
@@ -171,6 +171,10 @@ DrawingTool::SetOption(int32 option, int32 value, BHandler* source)
 				if (BControl* control = dynamic_cast<BCheckBox*>(source))
 					fToolSettings.use_current_brush = control->Value();
 			} break;
+			case WIDTH_OPTION:
+			{
+				fToolSettings.width = value;
+			} break;
 			default:
 				break;
 		}
@@ -211,6 +215,8 @@ DrawingTool::GetCurrentValue(int32 option)
 				return fToolSettings.transparency;
 			case USE_BRUSH_OPTION:
 				return fToolSettings.use_current_brush;
+			case WIDTH_OPTION:
+				return fToolSettings.width;
 			default:
 				return 0;
 		}
--- a/artpaint/tools/EllipseTool.cpp
+++ b/artpaint/tools/EllipseTool.cpp
@@ -13,9 +13,12 @@
 #include "EllipseTool.h"
 
 #include "BitmapDrawer.h"
+#include "BitmapUtilities.h"
 #include "Cursors.h"
 #include "Image.h"
+#include "ImageUpdater.h"
 #include "ImageView.h"
+#include "NumberSliderControl.h"
 #include "PaintApplication.h"
 #include "Selection.h"
 #include "ToolScript.h"
@@ -24,6 +27,7 @@
 
 #include <Catalog.h>
 #include <CheckBox.h>
+#include <GridLayoutBuilder.h>
 #include <GroupLayoutBuilder.h>
 #include <RadioButton.h>
 #include <SeparatorView.h>
@@ -40,14 +44,15 @@ EllipseTool::EllipseTool()
 {
 
 	fOptions = FILL_ENABLED_OPTION | SIZE_OPTION | SHAPE_OPTION | ANTI_ALIASING_LEVEL_OPTION
-	               | ROTATION_ENABLED_OPTION;
-	fOptionsCount = 5;
+	               | ROTATION_ENABLED_OPTION | WIDTH_OPTION;
+	fOptionsCount = 6;
 
 	SetOption(FILL_ENABLED_OPTION, B_CONTROL_OFF);
 	SetOption(SIZE_OPTION, 1);
 	SetOption(SHAPE_OPTION, HS_CENTER_TO_CORNER);
 	SetOption(ANTI_ALIASING_LEVEL_OPTION, B_CONTROL_OFF);
 	SetOption(ROTATION_ENABLED_OPTION, B_CONTROL_OFF);
+	SetOption(WIDTH_OPTION, 1);
 }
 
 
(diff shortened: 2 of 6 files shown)


humdingerb:
Very appreciated change!
Can something be done against the flickering while rotating?

> I'm not sure about the "Adjustable outline width" option. Would that happen after rotation? I didn't add that part for now.

Personally, I'd rotate first, then adjust the width. But is doesn't really matter...

dsizzle:
> Can something be done against the flickering while rotating?

well...since it's code I'm sure there is *something* that can be done. :smile: But I would need to learn what that is.  The problem is just that we need to draw the shape and erase it and re-draw it and I'm not totally sure how to do that with a BView. I believe I've seen stuff with "offscreen" views or bitmaps in Haiku. This flickering occurs pretty much everywhere there's an interactive UI in ArtPaint but I think it's just a lot more noticeable with thick solid shapes.  The easiest "solution" in this case would be to not show the line thickness until after you are done rotating, but that seems less than ideal.

dsizzle:
> This flickering occurs pretty much everywhere there's an interactive UI in ArtPaint but I think it's just a lot more noticeable with thick solid shapes.

Hmm, I lied, the straight line tool with a Width of 100 doesn't flicker at all. I will look into it.

dsizzle:
I think I made it a tiny bit better, but to really fix it the way the drawing is done would need to be changed to the way StraightLineTool does it - which to render it the same way we render the final straight line, using ArtPaint's drawing methods. Rectangle and Ellipse use Haiku's drawing.  I may attempt this.

humdingerb:
Good luck on your anti-flicker mission!
BTW, would it be possible to have the cursor show the outline-width before clicking in the canvas to create the rect/ellipse? Like the brush-size-outline for the other tools.

dsizzle:
> Good luck on your anti-flicker mission!

It’s actually going pretty well. I might be done by tomorrow. The code is rather ugly though.

> BTW, would it be possible to have the cursor show the outline-width before clicking in the canvas to create the rect/ellipse?

I’m not sure how to do this so that it wouldn’t be confusing. Like just a circle of the appropriate size?

humdingerb:
Ugly but working code may be improved with time. :)

> I’m not sure how to do this so that it wouldn’t be confusing. Like just a circle of the appropriate size?

Yes, I think a circle or square of the outline-width would give an indication what to expect.

dsizzle:
> Yes, I think a circle or square of the outline-width would give an indication what to expect.

what about when “fill” is checked?  Don’t show it?

humdingerb:
Don't show it, I'd say, because a filled rect has no outline.

dsizzle:
I feel like that’s a little weird because then sometimes you see it and sometimes you don’t?

humdingerb:
Your call, ofc.  Not confusing to me though 😃
The cursor shows the outline width. Filled rect has no outline, therefore no cursor-size-preview.

dsizzle:
Ok - please give it a try when you have a moment - there are some artifacts when you rotate quickly or resize quickly but I think it works well overall.  I even think it works when you zoom while drawing/rotating.  

Haven't tried the draw-circle-to-indicate-size yet; this is just the flicker fix.

humdingerb:
Yep, appart from the artifacts, the heavy flicker is gone. Nice!

dsizzle:
How bad are the artifacts on your end?

humdingerb:
Quite bad actually, if I move the mouse quickly. See [MPG 2.3 MiB]: 

https://github.com/HaikuArchives/ArtPaint/assets/6107158/d79187e2-a41f-4745-b8d3-336d8d995034

(Hope this shows or can be downloaded.)

A still image:

rect_arti_still

(At the end I zoom in and out to show these artifacts are actually pixels on the canvas.

dsizzle:
I tried both in WebPositive and Firefox on Linux and the video doesn't show - but the image is enough.  It is bad that the artifacts actually paint on the image permanently.

dsizzle:
updated and I think it's better - I can only see slight corner artifacts on rotation of a rectangle, and I can't get them to stay.  But the artifacts when resizing should be gone now.

humdingerb:
> I tried both in WebPositive and Firefox on Linux and the video doesn't show

Sorry about that. I tried uploading it to my ussual https://0x0.st, but it kept timing out.
So I uploaded to github, hoping it were possible to download the clip from here, if playback in the browser didn't work...

Anyway, with your latest change it is much better. Resizing, as you said, doesn't leave artifacts anymore at all. But I still manage to produce some sticky artifacts when rotating. Most disappear when I keep rotating, but a few remain.

There's also quite some tearing, isn't there?
As I may have touch on a related topic recently in one of my famous [one-line contributions](https://cgit.haiku-os.org/haiku/commit/?id=493d3c69f3346eaa10a48d154d7451e8d26b6990). Is BScreen's WaitForRetrace() something that could help with flicker/tearing?

dsizzle:
Ok, I made the rotating a bit more efficient and shouldn't have any artifacts. Occasionally I see cut-off corners but no extra artifacts, and the corners always reappear when the rectangle is final.

I tried to address the tearing by slowing down the updates and removing the forced image updates and I think it's better. If you are resizing like a crazy person you can't expect a *perfect* view. :laughing:

humdingerb:
Much better!

And yes, the cut-offs seem to depend on rect size and possibly where it is on the canvas. Only apppear whilst rotating and are gone when actually drawing the result on canvas.

A video uploaded to my mailbox.org cloud. Let me know if you can view/download it.

[rect_cut-off.avi](https://share.mailbox.org/ajax/share/019cfa440fa02c4f1df01a5fa02c4a12bb9e6f8ee9d7afa2/1/8/NDk/NDkvNTM)

dsizzle:
Ok, I was able to download and watch the video - that's far worse than what I was referring to.  I have only seen tiny corners getting cut off, not like 1/3 of the rectangle. I'll see if I can repro and fix that before merging.

dsizzle:
ok, I think I fixed it.

dsizzle:
No, NOW I think it's fixed for real. :smile:

humdingerb:
Very good!
Found #633, unrelated to thiis change though.

dsizzle:
merged!

PR #631, opened 2023-10-06
BrushStoreWindow: fix broken check for duplicate brushes

dsizzle:
- break statement was erroneously moved during The Big Cleanup, and it allowed duplicate brushes to be added to Brush window.

### DIFF
--- a/artpaint/windows/BrushStoreWindow.cpp
+++ b/artpaint/windows/BrushStoreWindow.cpp
@@ -493,7 +493,8 @@ BrushStoreView::AddBrush(Brush* brush)
 		brush_info item = *(brush_info*)brush_data->ItemAt(i);
 		if (Brush::compare_brushes(item, new_brush_info) == true) {
 			index = i;
-		} break;
+			break;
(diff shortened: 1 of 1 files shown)


artpaint/windows/BrushStoreWindow.cpp
```diff
@@ -32,7 +32,7 @@
 #include <ScrollBar.h>
 #include <ScrollView.h>
 
-
+s
```
humdingerb, line 35: The old sloppy "ALT+S" leaving an "s"... I do that all the time. :)

humdingerb (commented):
Sorry about that. I literally did break; something...

artpaint/windows/BrushStoreWindow.cpp
dsizzle: ok, that annoys me, because that means I pushed without even building first - very sloppy and I didn't think I engaged in such behaviors.  Shameful.

dsizzle:
merged.

PR #632, opened 2023-10-06
ColorPalette: rearranged and added field for color set names

### LINKED ISSUE #255: Label color sets
It may be nice to have an indication what color set is currently shown in the Colors window.
How about a label "Set 1" etc. to the right of the arrow widgets that switch sets below the palette?
~~Or as a text control to have the user name their color set, if that's not too much hassle.~~

Edited: Since the arrow widgets have moved to the right of the palette, there's less space now, so maybe just the number, no "Set", no user edited labelling.

dsizzle:
Fixes #255 

Incidentally, all the functionality was already there to support color set labels, so basically it was a layout exercise and then wiring up what was there plus a few minor bits like setting the default name. 

lmk what you think about the layout.


humdingerb:
Very good. Comments:

* The name "palette 1" should be translated and capitalized and renamed to "Color set #1".
* It's a bit irritating that the palette name always is in selection/editing mode when switching sets.
* Its text control always has focus. It'd be nicer if it didn't and the cursor keys changed the selected colour (maybe even with pageup/down switching colour sets).
* There's quite some flickering going on when resizing the window.

dsizzle:
Fixed the focus and the name; will think about the arrow keys.

Yeah the flickering is really bad. I guess before you couldn't resize it so it wasn't a problem.

dsizzle:
I did a quick fix for the flicker; I think it's a lot better.

humdingerb:
Looks flicker-free to me. 👍 
BTW, by default the Colors window has minimum dimensions. The palette swatches are a bit small, even the default 16 colors. Could the default window size made a bit wider and taller, so the swatches are 2x the current height?

humdingerb:
One more thing: The color set name shouldn't allowed to be empty. How about reverting to the default "Color set #" in that case?

dsizzle:
- doubled the minimum height of the color container
- if the user tries to enter an empty name, it reverts to the previous name (this was the simplest change and programmers are lazy)

dsizzle:
... and added keyboard control to the color container view - it needs to have focus but then left/right/up/down and page up/page down now work as requested.

humdingerb:
Nice!
Would it work to have the focus be automatically on the color container view when switching to the colors window? Maybe by calling MakeFocus(true) in the view's WindowActivated()?

dsizzle:
> Would it work to have the focus be automatically on the color container view when switching to the colors window? Maybe by calling MakeFocus(true) in the view's WindowActivated()?

sheesh, so demanding. Your wish has been granted.

humdingerb:
Beautiful! Quick, merge before I have more brilliant ideas!

BTW, would that have worked too with using the color_container view's WindowActivated()? At least I interpret the BeBook's description as such:

> Implemented by derived classes to take whatever steps are necessary when the BView's window becomes the active window, or when the window gives up that status. If active is true, the window has become active. If active is false, it no longer is the active window.

dsizzle:
> Beautiful! Quick, merge before I have more brilliant ideas!

I appreciate all of your ideas and insight!

> BTW, would that have worked too with using the color_container view's WindowActivated()?

Perhaps it might. I was just thinking that controller the focus via the parent window makes more sense than controlling it from the container control. How does it resolve conflicts if two different controls on the same window try to get focus in their respective WindowActivated() methods?

humdingerb:
> I appreciate all of your ideas and insight!

"All" of them? Let's say "some" or even "most" if we want to be generous... :)

> ...controlling the focus via the parent window makes more sense...

Agreed.

PR #634, opened 2023-10-12
Clang static analysis: Fixed possible memory leaks and dead inits

dsizzle:
Fixes related to #124


humdingerb:
Nice. Fewer lines of code means less hiding holes for bugs.

Issue #197, opened 2022-06-05
Text only draws in black


dsizzle:
Text tool should respect the current colors, but instead it only renders black.

PeteCA:
On Sat, Jun 04, 2022 at 08:27:08PM -0700, Dale Cieslak wrote:
> Text tool should respect the current colors, but instead it only renders black.
> 
Unless things haver changed recently, this isn't exactly so.  However, I've just
spent about half an hour (with the old version) getting totally confused.
Then I finally went to the (old!) manual, and everything became clear!
It is definitely not 'intuitive'...
Basically -- in brief -- you need to have *both* the text-entry window *and*
the color window open.  Then any *change* in the color window will be
reflected in the text.  Not quite 'live' -- you have to release the mouse button
before it shows.  That color is retained for text only, unaffected by any later
color changes in foreground or background color.

As an example sequence:
	Select the text tool, and click on the canvas to make it active.
	Also open the color selection window.
	Make a desired *change* to the color window.
	The text should change to the selected color.
	If any characters of the entered text are selected, only they
	will change color.  Any insertions seem to assume the color
	of the surrounding characters, but only the last color selection
	is retained.

Hope that makes sense. I suspect it was a deliberate choice, so text color
is essentially independent of others, but it is definitely confusing to a naive
user!  I suspect we can do better, but I have no good ideas ATM.

humdingerb:
Excellent work, Pete Holmes! :)

But yeah, a bit weird. It doesn't make much sense having the colour apply to the slected part of the text, if font and size always apply to the whole text. (Rotation and shear won't make much sense if applied to parts of the text, I think).

I suggest, for now, to either
a) always make the whole text the foreground colour, and start with a foreground coloured text.
b) also optionally apply font and size to just a selection of text and apply a changing foreground colour live.

Maybe there should be a colour well added that shows the current foreground colour and openes the Colors window when clicked. That's suggest to change the colour in the Colors window, if you want the text colour changed.

dsizzle:
Yes, thanks for the investigation, Pete!!

While it does sound flexible and potentially interesting, it also sounds completely awful from an intuitivity(?) and discoverability perspective.  

I've always liked the "Principle of Least Surprise" when it comes to UI/UX so I think it should always use the current foreground color. It makes sense and is consistent with every other tool. 

[...]

PR #635, opened 2023-10-12
TextManipulator: make text use current fg color only

### LINKED ISSUE #229: Text tool typing box can be unreadable
The typing box uses document background colour but the text is in black so this is causing problems with dark colour schemes.
I see two solutions:
The first would be to use document background colour and document text colour there.
The second would be to use a fixed white colour background. Indeed, since (from what I understand from issue 197) ArtPaint only draws text in black, it makes sense to have the text in black in the typing box.
 
Artpaint text tool

dsizzle:
- previous functionality was flexible but non-intuitive

Fixes? #197 
Fixes #229


dsizzle:
reopened with better fix

PR #636, opened 2023-10-12
TextManipulator: make text use current fg color only

### LINKED ISSUE #197: Text only draws in black

### LINKED ISSUE #229: Text tool typing box can be unreadable

dsizzle:
- previous functionality was flexible but non-intuitive

Fixes #197 - depending on your point of view.   Removes the ability to multi-color text but IMO not a bad thing.
Fixes #229

### DIFF
--- a/artpaint/viewmanipulators/TextManipulator.cpp
+++ b/artpaint/viewmanipulators/TextManipulator.cpp
@@ -15,6 +15,7 @@
 #include "HSPolygon.h"
 #include "MessageConstants.h"
 #include "NumberSliderControl.h"
+#include "PaintApplication.h"
 #include "PaletteWindowClient.h"
 #include "Selection.h"
 #include "UtilityClasses.h"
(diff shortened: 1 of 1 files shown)


humdingerb:
I think it's OK to make the text tool more simple but intuitive, even if we lose a (hidden) feature. Maybe one day it'll become really powerful, with easy formatting and some layouting end stuff, possibly in a special text-layer that keeps the text editable etc.
Not super urgend though, I think.

dsizzle:
Yeah, I expect Text needs a redo in the longer term but I am thinking about painting/editing tools and functionalities before worrying about that. I don't expect to work on Text tools any time soon.

PR #637, opened 2023-10-13
EllipseTool: fix to remove initial rotation

### LINKED ISSUE #633: Ellipse rotating issue when != 100%
Do this:
* Set zoom to something other than 100%
* Set up the ellipse tool with rotation.
* Create an ellipse. When you mouse-up when you're satisfied with the shape and it comes to the rotation part, the ellipse gets rotated a bit even though you the mouse hasn't moved yet. The amount of initial rotation seems to depend on the zoom level. 

No problem at 100% zoom. The issue doesn't present with the rectangle tool.

dsizzle:
Fixes #633

### DIFF
--- a/artpaint/tools/EllipseTool.cpp
+++ b/artpaint/tools/EllipseTool.cpp
@@ -211,8 +211,8 @@ EllipseTool::UseTool(ImageView* view, uint32 buttons, BPoint point, BPoint)
 
 		float new_angle, prev_angle;
         if (GetCurrentValue(ROTATION_ENABLED_OPTION) == B_CONTROL_ON) {
-			BPoint centroid = new_rect.LeftTop() + new_rect.RightTop()
-				+ new_rect.RightBottom() + new_rect.LeftBottom();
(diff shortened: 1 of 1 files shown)


humdingerb:
Fixed indeed!

BTW, rotated ellipses appear a bit frazzled. More so than a rotated selection, it appears. Anything that could be transferred to unfrazzle the ellipazzle?

dsizzle:
> BTW, rotated ellipses appear a bit frazzled. 

Rotating ellipses is one of the harder things I've done in ArtPaint. I'm still not 100% happy with it either but for now I'm tired of it - I have spent a *lot* of time researching how to draw rotated ellipses. 

> More so than a rotated selection, it appears. Anything that could be transferred to unfrazzle the ellipazzle?

I'm going to say "no" for now but leave the door open for "maybe" :smile:

humdingerb:
Luckily there are always other features clamouring for attention to take your mind off the ones overstepping your frustration threshold... :)

PR #638, opened 2023-10-13
ToolSetupWindow: move Show() to *after* window feel is set.

### LINKED ISSUE #290: Button to close 'Tool settings' window is too small when the window is active
It's on nightly x64 hrev56511
Button size is good when the window is inactive (on left) but too small when the window is active (on right).
Screenshots are with Flat decorator but it also happens with the other decorators. (Haiku problem?)
It doesn't seem to happen in other windows but I only had a quick look...

inactive active

dsizzle:
Fixes #290 

- also did minor code move to remove extra 'if' statement

### DIFF
--- a/artpaint/tools/ToolSetupWindow.cpp
+++ b/artpaint/tools/ToolSetupWindow.cpp
@@ -52,8 +52,6 @@ ToolSetupWindow::ToolSetupWindow(BRect frame)
 		Unlock();
 	}
 
-	Show();
-
 	sfToolSetupWindow = this;
 	FloaterManager::AddFloater(this);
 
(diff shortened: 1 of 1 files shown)


humdingerb:
Fixed! Was the reason obvious to you, or a mystery that needed some digging? Just wondering, when you see the solution, it seems almost "duh"...

dsizzle:
> Was the reason obvious to you, or a mystery that needed some digging? 

Not at all obvious.  I have spent time looking at this on multiple occasions and trying to figure out what in the world is different about the Tool Setup window.  

> when you see the solution, it seems almost "duh"...

I wouldn't agree with that - I feel like this changed in the last year or so and wasn't an issue before. I think the "feel" should be able to be changed when a window is visible.

humdingerb:
> I wouldn't agree with that - I feel like this changed in the last year or so and wasn't an issue before.

Dunno when it first appeared. The "duh" part to me was that it was linked to the Feel in the first place. In retrospect, it's at least very probably that the Feel - i.e. the look of the window title bar and its close widget - should have to do something with it. Of course, that it was hinging on the Show() state of the window is not obvious.
Anyway, "death halloo" to the victorious bug hunter!

> I think the "feel" should be able to be changed when a window is visible.

Agreed.

dsizzle:
> The "duh" part to me was that it was linked to the Feel in the first place

ah, ok. Yes, THAT part was obvious to me - I knew it was somehow related to the Feel, but all the other windows do the same kind of changes for the "always on top" setting, so I was blocked on trying to figure out what in the world was different.

Issue #643, opened 2023-11-03
Hang in BitmapUtilities::RasterToPolygonsMoore


dsizzle:
The problem seems to be the function getting into a bad loop at the edges of the bitmap, or something similar. A way to trigger the issue is to move around at the low end of the Size slider in the Brush window. It seems like small sizes of a brush trigger the issue.  ArtPaint doesn't hang but the Brush editor does, and painting with a Brush no longer works.

PR #644, opened 2023-11-03
BrushEditor: size preview based on max dimension to avoid jumping

### LINKED ISSUE #523: Jumping brush size in preview panel when rotating
As mentioned in #522.
Create a brush that just fits diagonally into the preview pane:

screenshot1

Then slowly move the Angle slider: As old Pythagoras found out recently, applying his groundbreaking work on hypotenuses, the diagonal of a square often is longer than its sides. This forces the scale to adjust to make the brush fit, but also shrinks the brush preview:

screenshot2

dsizzle:
Fixes #523 

(but watch out for #643 when testing, which is unrelated to this change)

### DIFF
--- a/artpaint/paintwindow/StatusView.cpp
+++ b/artpaint/paintwindow/StatusView.cpp
@@ -395,8 +395,7 @@ CurrentBrushView::CurrentBrushView(BRect frame)
 
 	SetToolTip(B_TRANSLATE("Brush"));
 
-	frame.InsetBy(1.0, 1.0);
-	fBrushPreview = new BBitmap(BRect(0.0, 0.0, frame.Width(), frame.Height()), B_RGBA32);
+	fBrushPreview = new BBitmap(BRect(0.0, 0.0, frame.Width(), frame.Height()), B_RGBA32, true);
 	BitmapUtilities::ClearBitmap(fBrushPreview, 0xFFFFFFFF);
 
 	list_of_views.AddItem(this);
@@ -412,11 +411,7 @@ CurrentBrushView::~CurrentBrushView()
 
 void CurrentBrushView::Draw(BRect)
 {
-	DrawBitmap(fBrushPreview, BPoint(1.0, 1.0));
-
-	SetPenSize(1);
-	SetHighColor(0, 0, 0, 255);
-	StrokeRect(Bounds());
+	DrawBitmap(fBrushPreview, BPoint(0.0, 0.0));
 }
 
 
--- a/artpaint/tools/Brush.cpp
+++ b/artpaint/tools/Brush.cpp
@@ -379,20 +379,15 @@ Brush::delete_all_data()
 float
 Brush::PreviewBrush(BBitmap* preview_bitmap)
 {
-	float preview_width = width_;
-	float preview_height = height_;
+	float bmap_width = (preview_bitmap->Bounds().Width() + 1);
+	float bmap_height = (preview_bitmap->Bounds().Height() + 1);
 
-	float bmap_width = preview_bitmap->Bounds().Width() + 1;
-	float bmap_height = preview_bitmap->Bounds().Height() + 1;
+	float preview_width = 0.9 * bmap_width;
+	float preview_height = 0.9 * bmap_height;
 
-	while ((preview_width > bmap_width) || (preview_height > bmap_height)) {
-		preview_width /= 2.0;
-		preview_height /= 2.0;
-	}
-	int32 scale = (int32)(width_ / preview_width);
+	float max_dim = max_c(width_, height_);
 
-	int32 top = (int32)((bmap_height - preview_height) / 2.0);
-	int32 left = (int32)((bmap_width - preview_width) / 2.0);
+	float scale = max_c(1.0, max_dim / preview_width);
 
 	uint32* bits = (uint32*)preview_bitmap->Bits();
 	int32 bpr = preview_bitmap->BytesPerRow() / 4;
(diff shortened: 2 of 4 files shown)


humdingerb:
Works in the "Brushes" window, but there are sometimes artifacts in the canvas window status bar preview, esp. while rotating.
* create a rect brush and angle it diagonally
* increase its size very slowly until the scale jumps and it appears smaller
* now decrease its size by one tiny nudge
-> The preview in the Brushes window shows a big brush with the now aain larger scale, and the preview in the canvas status bar still shows the brush in the smaller scale, plus an artifactal line:

Hmm, cannot upload a file. Not possible in a PR comment? Anyway, uploaded as https://0x0.st/Ht-B.png

----

There's also a cut-off in the Brushes preview sometimes:
* create a rect brush and angle it diagonally
* increase its size until it almost touches the preview pane frame
* move the apect slider
-> the right side is cut of and it appears a bit of the left side pokes in at the rigt side: https://0x0.st/Ht-1.png

----

I also had to remove the ArtPaint settings or it crashed on launch: https://0x0.st/Ht-u.report

dsizzle:
Did I mention that I knew this would be a pain in the backside to fix?  :smile:

dsizzle:
Ok, please give it another try - it's slightly different than before in that the brush won't keep growing and shrinking; once it reaches "maximum" size, only the red bar changes, but I think that kind of makes sense rather than continually growing and shrinking. 

I do want to put a number indicating the numerical size someday, rather than just that bar.

humdingerb:
I like it.
Let me put this heretic thought to you: let's scrap the red line. There, I said it. :)
It was kinda neat when you had fixed scales - 1:1, 1:2, 1:3 - and the brush size in the preview changed. But now, with the brush size being constant and the scale incrementally in/decreasing, the usefulness of the line is rather diminished. To be frank, even before, the red line only gave you a rough sense of the actual brush size when it was 1:2 or 1:3. Above that it was hard to tell if it's 1:4 or 1:5...

Anyhow, here's a mockup of the size as number in a corner: https://0x0.st/HtXZ.png

Oh, and a large rect brush still gets cut-off at the right side of the preview.

dsizzle:
> Let me put this heretic thought to you: let's scrap the red line. There, I said it. :)

I never really understood the red line honestly. I don't think it conveys much information. I think a number would be so much better - but I was just going to write it on top of the brush like Photoshop does, not a fancy gray border or anything.

> Oh, and a large rect brush still gets cut-off at the right side of the preview.

WHAT?  Hmm, I didn't have that happen - I will look into it more.

humdingerb:
Another idea I just had. Hey, don't you roll your eyes at me, young man! :)

The thing is, even a number with the pixel size often isn't that helpful. It's OK as size indicator for the stored brushes pane. But for actually drawing with it on the canvas, it's just a number. The actual size of the drwn line depends on the dimensions of the canvas and the current zoom level. It would be nice, if while you're fiddling with the brush sliders, there'd be an overlay of the current setting on the canvas itself.

dsizzle:
Ok, I removed the red line and added a number for the size when it's larger than the preview.

>  It would be nice, if while you're fiddling with the brush sliders, there'd be an overlay of the current setting on the canvas itself.

:roll_eyes: 

Just use the Freehand tool with the "Use current brush" mode, and move the mouse over the canvas. Done.

humdingerb:
Sure, but then you have to move back and forth between the sliders and the canvas. Having the constantly updated outline overlay in the middle of the canvas while you're twiddling the sliders would be neat.
But, as I said, just an idea. Many more fishies to fry in the deep ArtPaint ocean...

humdingerb:
The brush not being in the center is a bit weird when resizing it and becomes a little wilder when rotating: https://0x0.st/HtAR.mpeg
Not necessaril bad, but unexpected... :)

To keep the GUI stable and also because the size info can also be handy for smaller sizes, I'd suggest showing it always.

I think we could get away with a smaller font size for the size info.

Keeping the brush filling the preview pane as before, but just slapping the size info on top in a corner isn't easily done, I assume. Probably some flickering/drawing views on top of each other/or a similar nuisance?

dsizzle:
> The brush not being in the center is a bit weird when resizing it and becomes a little wilder when rotating

Ok, I must really be having an off-day. I swear I fixed that so it's always centered.  It's not supposed to be like that at all. It's supposed to be centered.  So -

> Not necessaril bad, but unexpected... :)

- yes, it's bad AND unexpected. :(

> To keep the GUI stable and also because the size info can also be handy for smaller sizes, I'd suggest showing it always.

will do. 

> Keeping the brush filling the preview pane as before, but just slapping the size info on top in a corner isn't easily done, I assume. 

That's not hard, but I'd have to draw a white rectangle behind it or something so the text is legible - I don't think it'll flicker or anything.

humdingerb:
> That's not hard, but I'd have to draw a white rectangle behind it or something so the text is legible - I don't think it'll flicker or anything.

May be worth a try. Few brushes fill the preview all the way to its corners, so obscuring that part wouldn't be bad. And that way we wouldn't waste vertical space needlessly.
And if you want to have it look nice, use a panel-background grey and draw the black frame border at the top and right of the size box, as in mockup from above... :)
corner-size

dsizzle:
OK, how 'bout now?  I tried to do the fancy drawing but the black border is drawn by the parent view so I would have to change the code in the brush editor and the status view as well, and I'm too lazy to do that. :smile:

dsizzle:
Ok, FINE, I did the "fancy" border - let me know if you see any other issues.

humdingerb:
Beautiful! :)

Would it be possible to put the size numbers smack in the corner, without spacing  to the left and bottom? That would save some space, see:

Now:

now

Then:

then

dsizzle:
better?

humdingerb:
Perfect! Thank you.

dsizzle:
Excellent, merged

PR #645, opened 2023-11-05
AirBrushTool: attempting to get a smoother, better performing Airbrush

dsizzle:
- only calculate 1/4 of the brush and draw 4 pixels at once
- use LINEAR_INTERPOLATION in the CoordinateReader to get a smoother line
- adjusted linear interpolation to only get 1/10th of the coordinates
- remove snooze

@humdingerb can you give this a try to see if it's a better Airbrush experience?  It works a lot better on my end but I'm also not seeing what you're seeing, so this is just a more general attempt to improve the Airbrush feeling.

### DIFF
--- a/artpaint/tools/AirBrushTool.cpp
+++ b/artpaint/tools/AirBrushTool.cpp
@@ -76,7 +76,7 @@ AirBrushTool::UseTool(ImageView* view, uint32 buttons, BPoint point, BPoint)
 
 	image_view = view;
 	CoordinateReader* coordinate_reader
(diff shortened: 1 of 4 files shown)


humdingerb:
Nice. With this change it's again as fast as before!

dsizzle:
Ok, I cleaned up the change a bit - please re-test and if it still behaves well I can merge it.

humdingerb:
Still perfect. Good change!

dsizzle:
merged!

PR #646, opened 2023-11-06
Layer/ImageView: respect alpha value when cutting/copying/deleting

dsizzle:
Missed a spot with the "respecting alpha in selections" changes, and I expect there could be a few more spots hiding along the way - the cut/copy/delete now respects selection alpha. 

The easiest way to describe this is with some pictures - load up Walter, then make a layer with a gradient on it:
0walt

...then "Select all non-transparent" on that gradient layer and then hide that layer...
1walt

...go to the Walter layer and hit the Delete key. 

the `master` branch does this :disappointed: 
3walt

But with this change, it does this, as it should:
2walt

You can also cut/copy/paste with alphaed selections.

### DIFF
--- a/artpaint/layers/Layer.cpp
+++ b/artpaint/layers/Layer.cpp
@@ -155,8 +155,13 @@ Layer::Clear(rgb_color color)
 		int32 bpr = fLayerData->BytesPerRow() / 4;
 		for (int32 y = top; y <= bottom; ++y) {
 			for (int32 x = left; x <= right; ++x) {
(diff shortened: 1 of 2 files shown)


humdingerb:
Very cool. Proper alpha masking: check!

dsizzle:
...and this merge takes us another step closer to paintable/blurrable selections...yay!

humdingerb:
And a hard fight it was... kudos!

PR #647, opened 2023-11-07
BrushStoreWindow: delay drawing until all brushes are loaded

### LINKED ISSUE #602: Don't update Brush window until all brushes are loaded
When opening ArtPaint, if the Brush window is open, the brush preview flickers as each brush is loaded. It would be nicer if the window didn't display the brush preview until they were all done.

dsizzle:
Fixes #602

### DIFF
--- a/artpaint/windows/BrushStoreWindow.cpp
+++ b/artpaint/windows/BrushStoreWindow.cpp
@@ -240,6 +240,8 @@ BrushStoreWindow::brush_adder(void* data)
 {
 	BrushStoreWindow* this_pointer = (BrushStoreWindow*)data;
 
+	this_pointer->Lock();
+	this_pointer->BeginViewTransaction();
 	if (this_pointer != NULL) {
 		BList temp_list(*brush_data);
 		if (temp_list.CountItems() > 0) {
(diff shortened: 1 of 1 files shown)


humdingerb:
The effect was barely visible on my 9 year old machine... but it's gone with this change. It's still there in the brush preview in the canvas window's status bar though.

dsizzle:
Really?  I have never seen that happen in the status bar. You see the status bar preview cycle through all the brushes?

humdingerb:
I never noticed it before, but since this PR fixed the preview in the Brushes window, I had a look. It's identical to the what happened in the Brushes preview. Last only 1/10s or something. I had to double the number of my saved brushes to now 12. It helps to mix round and rect brushes to see the changes.

dsizzle:
ok, I had to try many times to get it to happen, so I'm not super worried about it.  I'm going to go ahead and merge this change for now.

dsizzle:
of course as soon as I did, now I see it every time... sigh...

humdingerb:
FWIW, I didn't find it annoying. It's like when you turn on the ignition and the speedo/rpm meter resets and their hands quickly move to max and back to zero. :)

PR #650, opened 2023-11-09
Brush: fix brush size shown in preview

dsizzle:
Introduced in #644, can't believe we didn't catch it!  For rectangular brushes, the size number in the brush preview changed with the Brush angle...ooops...  fixed now!


humdingerb:
oops indeed... :)
Fix works!

PR #651, opened 2023-11-09
AirbrushTool: simplify code for Spray; make it work with alpha selections

dsizzle:
The `master` version of Airbrush doesn't work well with the alpha areas of selections, but this version does, and the code is simpler.

I tried a lot of back and forth and I *think* it looks about the same as before - hard to tell.


humdingerb:
LGTM

dsizzle:
ok, then - merged

PR #657, opened 2024-12-27
93 threshold live

### LINKED ISSUE #93: Threshold addon: apply settings live
Moving the red line in the histogram widget to set the threshold only applies the effect after releasing the mouse button. The effect should be live, if possible.

HiCamino:
Hi, I have been looking at this project for a while and I am trying to fix some of these. This fix makes the Threshold add-on update when you drag the line in the histogram. I also fixed the drawing of the histogram. 

Fixes #93

### DIFF
--- a/addons/AddOns/Threshold/ThresholdView.cpp
+++ b/addons/AddOns/Threshold/ThresholdView.cpp
@@ -75,7 +75,7 @@ void ThresholdView::Draw(BRect)
 	FillRect(clearRect);
 
 	if (histogramBitmap != NULL)
-		DrawBitmap(histogramBitmap, histogramBitmap->Bounds(), histogramRect);
+		DrawBitmap(histogramBitmap, BPoint(histogramRect.left, histogramRect.top));
 
 	SetHighColor(255, 0, 0, 255);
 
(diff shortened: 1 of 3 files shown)


addons/AddOns/Threshold/ThresholdView.cpp
```diff
@@ -75,7 +75,7 @@ void ThresholdView::Draw(BRect)
 	FillRect(clearRect);
 
 	if (histogramBitmap != NULL)
-		DrawBitmap(histogramBitmap, histogramBitmap->Bounds(), histogramRect);
+		DrawBitmap(histogramBitmap, BPoint(histogramRect.left, histogramRect.top));
```
korli, line 78: histogramRect.LeftTop() would be better.

addons/AddOns/Threshold/ThresholdView.cpp
```diff
@@ -148,8 +148,16 @@ ThresholdView::MouseUp(BPoint point)
...
 	threshold = (int32)min_c(255, max_c(0, point.x - histogramRect.left));
-	Invoke();
+	SetValue(threshold);
+}
+
+
+void
+ThresholdView::SetValue(int32 value)
+{
+	BControl::SetValue(value);
```
korli, line 158: The new value should be checked against the old value before proceeding.

humdingerb:
@korli: satisfied with the changes?

PR #658, opened 2025-01-16
Added notification/messages to make Effect Strength slider work "live"

### LINKED ISSUE #92: Sharpness addon: make "Effect strength" live
While the "Sharpness" slider takes effect while you move the slider's knob, the "Effect strength" slider only takes effect after letting go of the knob. All settings should be live, if possible.

HiCamino:
This works but the gaussian blur could be faster it is "live" but it can be slow if you move the slider too fast. I found this page so maybe I can try to get this to work: https://blog.ivank.net/fastest-gaussian-blur.html but probably in a separate pull request

Fixes #92

### DIFF
--- a/addons/AddOns/Sharpness/Sharpness.cpp
+++ b/addons/AddOns/Sharpness/Sharpness.cpp
@@ -398,12 +398,15 @@ SharpnessManipulator::ChangeSettings(ManipulatorSettings* s)
 	new_settings = dynamic_cast<SharpnessManipulatorSettings*>(s);
 	if (new_settings != NULL) {
 		if (new_settings->blur_size != settings.blur_size) {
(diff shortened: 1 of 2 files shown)


korli:
There is a SetSnoozeAmount() on BSlider to snooze between modification messages on MouseDown(). Default is 20000.
You can surely make it greater when needed.

korli:
@HiCamino is it ready for merge?

korli:
@HiCamino OK, not yet tested from me. I'll try myself.   @humdingerb  I see you tested some PR for dsizzle, maybe you're interested with this one?

humdingerb:
I'm afraid that for really big images (8k), the Effect Strength slider is very unresponsive. As is, this shouldn't get merged IMO.

PR #660, opened 2025-03-11
Faster blur for sharpness

### LINKED ISSUE #92: Sharpness addon: make "Effect strength" live

HiCamino:
One last try, I hate to giveup so I moved code around in the thread and i use a fast blur from https://blog.ivank.net/fastest-gaussian-blur.html it seems better. 4k is pretty responsive and 8k is a little slow but the slider stays responsive. let me know if it is good enough

Fixes #92

### DIFF
--- a/addons/AddOns/Sharpness/Sharpness.cpp
+++ b/addons/AddOns/Sharpness/Sharpness.cpp
@@ -21,6 +21,9 @@
 #include "Selection.h"
 #include "Sharpness.h"
 
+
+#include <stdio.h>
+
 #undef B_TRANSLATION_CONTEXT
 #define B_TRANSLATION_CONTEXT "AddOns_Sharpness"
 
@@ -104,7 +107,7 @@ SharpnessManipulator::ManipulateBitmap(
 
 	delete blurred_image;
 	blurred_image = DuplicateBitmap(source_bitmap, 0);
-	ipLibrary->fast_gaussian_blur(blurred_image, BLUR_AMOUNT);
+	ipLibrary->fast_gaussian_blur(blurred_image, BLUR_AMOUNT, 1);
 //	CalculateLuminanceImage(source_bitmap);
 
 	start_threads();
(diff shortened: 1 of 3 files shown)


humdingerb:
The visual results seem to be quite bad. Take any image and slide the "Sharpness" slider between both extremes. Instead of blurred it gets blocky, and sharpening also introduces some blockiness. Moving the "Effect strength" slider introduces major artefacts.
I tried with relatively small images, as it appears to crash with really big ones...

Have you tested our changes?

humdingerb:
Some images showing the blockiness and artefacts:

Sharpness slider min:
sharpness-0

Sharpness slider max:
sharpness-max

Sharpness at ~80%, moving he effect strength slider produces artifacts:
effect-strength

The crash when moving the sharpness slider a few times with a large image:
[ArtPaint-17968-debug-12-03-2025-06-42-50.report.txt](https://github.com/user-attachments/files/19204657/ArtPaint-17968-debug-12-03-2025-06-42-50.report.txt)

All 100% reproducible here. Weir that it's working for you... I'm on Haiku Beta5, 64bit.

humdingerb:
I rebased from master, and checked out this PR 660 on top of it. I did a "build.sh clean all" to remove all old objects and built from scratch. I still see the issues described above.

Not sure how we should proceed here. Maybe we need to find a third person to break the tie... :)

humdingerb:
No harm done. I didn't spend too much time on it, at least. :)
It's still weird that I see these issues while you're not. Anyway, I suggest to keep this PR open in case somebody stumbles on it and wants to give it a try.

korli:
> maybe beta5 is behaveing different than nightly? haikuversion

Are you in a VM? if yes, how many cpus? i9-14900K is very powerful, that might also explain why it works "better".

humdingerb:
I assume @korli's question was to HiCamino, though I just want to point out that my issues are not with speed, but the artifacts, blockage and crash.

Issue #664, opened 2025-09-08
Scale is unusably slow on large images (4k, 8k)


HiCamino:
separated from #661 because translate and scale probably have different solutions.

Doing a selection on a large image and then trying to use edit->scale it is too slow to be used.

PR #665, opened 2025-09-09
Fix slow translate by removing extra copying

### LINKED ISSUE #661: Translate is unusably slow on large images (4k, 8k)
Doing a selecton on a large image and then trying to use edit->Translate it is too slow to be used.

HiCamino:
The Translate Manipulator code was copying the selection bitmap mutliple times every time you move the mouse when translating so I removed those copies. i think the author was trying to guard against the selection changing? but in translating it never changes and just moves around so no need to regenerate it every time. 
I also added some cleanup code to free memory used by bitmaps. 

i did some timing tests with an 8000px image and a selection of about 1400x1400
before this change it was about 1500ms every time i moved the mouse. after the change is <1ms
(even though my pc is fast it should be noticably faster on any machine without doing all the copies)

i tried the same thing with scale but it seems to be more complicated so far I couldn't fix it yet but I made a new bug for it #664

also let me know if you find bugs with this method bc it seems to work great on my machine

Fixes #661

### DIFF
--- a/artpaint/viewmanipulators/TranslationManipulator.cpp
+++ b/artpaint/viewmanipulators/TranslationManipulator.cpp
@@ -41,7 +41,8 @@ TranslationManipulator::TranslationManipulator(BBitmap* bm)
 	WindowGUIManipulator(),
 	copy_of_the_preview_bitmap(NULL),
 	selection(NULL),
-	transform_selection_only(false)
+	transform_selection_only(false),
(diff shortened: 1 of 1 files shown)


korli:
Code LGTM, I didn't test.

humdingerb:
After a bit of testing, I concur, it works nicely!
Thanks very much!

PR #671, opened 2025-10-07
Check x and y values to make sure they are valid before translating

### LINKED ISSUE #670: Moving a selected area way out of the canvas crashes
After 9f3b8eea4f4cf346adb443d0e51f776c4849df67 this crashes ArtPaint (whereas before, it'd mess up or clear the canvas):

* Open a largish image.
* Zoom far out, so the canvas is quite small in the window.
* Select an area.
* Translate (ALT+T) to move the selected area way outside the canvas.
* Confirm the translate-dialog with OK.
--> Crash

When starting ArtPaint with heap protection (`LD_PRELOAD=libroot_debug.so MALLOC_DEBUG=grs25`) the Debugger puts the segfault in line https://github.com/HaikuArchives/ArtPaint/blob/master/artpaint/viewmanipulators/TranslationManipulator.cpp#L224

Attached the debug report: 

[ArtPaint_debug-4721-debug-03-10-2025-15-05-04.report.txt](https://github.com/user-attachments/files/22692092/ArtPaint_debug-4721-debug-03-10-2025-15-05-04.report.txt)

HiCamino:
this seems to fix the crashing with translating out of the image

Fixes #670

### DIFF
--- a/artpaint/viewmanipulators/TranslationManipulator.cpp
+++ b/artpaint/viewmanipulators/TranslationManipulator.cpp
@@ -214,11 +214,17 @@ TranslationManipulator::ManipulateBitmap(
 		top = (int32)selection_bounds.top;
 		bottom = (int32)selection_bounds.bottom;
 		for (int32 y = top; y <= bottom; y++) {
+			if (y < 0)
+				continue;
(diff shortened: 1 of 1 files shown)


korli:
LGTM

PR #673, opened 2025-10-11
fix bounds checks when translated selection is moved in -x or -y

### LINKED ISSUE #672: Translating selection in negative direction draws incorectly
caused by #671 (sorry!) but if you move a selection in negative x y or both then click ok it seems to be translated by double the amount and gets cut off by selection.

HiCamino:
translated selection bounds so they match the translated selection and then make sure no drawing is attempted outside of the edges of the image
- also moved/added some lock/unlock for bitmaps to hopefully make it safer

Fixes #672


korli:
LGTM

humdingerb:
Thanks for having a look!

PR #674, opened 2025-10-19
Add checkbox to color picker to use all layers

### LINKED ISSUE #487: Color picker: pick from all visible layers
Currently only the active layer is used for picking a color.

I first thought there should be a radio option like "All layers" or "Active layer". But is that actually needed?
Normally you want to pick the colour under the mouse pointer and do not care on which layer it is. Otherwise, if you enter a transparent area on your layer, it just picks up alpha = 0 (as it does now).

HiCamino:
this seemed like a pretty easy fix and works in my testing

but one weird thing is if you choose 'sample all layers' then it can get the color of the checkerboard bakground but i don't think that is fixable because of how the image is rendered in Artpaint?

Fixes #487

### DIFF
--- a/artpaint/tools/ColorSelectorTool.cpp
+++ b/artpaint/tools/ColorSelectorTool.cpp
@@ -281,7 +281,10 @@ ColorSelectorTool::ColorSelectorTool()
 	fOptionsCount = 2;
 
 	SetOption(SIZE_OPTION, 1);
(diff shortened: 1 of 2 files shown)


HiCamino:
i have ideas to fix it but maybe it should be a separate pr

humdingerb:
Works perfectly now. Merge!

PR #678, opened 2025-11-24
Refactored TranslationManipulator::PreviewBitmap

HiCamino:
Refactored TranslationManipulator::PreviewBitmap to match how ScaleManipulator::PreviewBitmap works to make it work properly when ImageView is fixed. Basicaly makes updated region simpler.

i am reworking the checker bg to work properly with Haiku invalidate/draw/etc and in my branch TranslationManipulator doesn't work properly but with this fix it does. it should work exactly as before (but maybe better performance?) this isnt a feature change just a refactor.

I sometimes see little artifacts with a large selection and large image but i can't get it to happen regularly to debug it. it is only temporary and goes away when you click ok or cancel. let me know if it is bad or can figure out how to make it happen regularly but i have spent a while trying to fix it but not even sure why/when it happens. seems somehow related to moving off the edge of the image but its pretty rare?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,18 @@
+v2.7.2:
+
+* Big speed-up for scaling and moving, esp. when working with large images
+* Fixed crash on resizing
+* Added commands to flip a selection horizontally/vertically
+* Treshold add-on:
+  - support live preview
+  - fixed histogram rendering
+* Updated documentation
+* Updated translations
+* New translations:
+  - Ukrainian (uk) (Alexey Lugin, stotish)
+  - Czech (cs) (theokeist)
+
+
 v2.7.1:
 
 * Quick fix for 1-pixel width freehand tool freezing the app.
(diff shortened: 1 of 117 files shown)


HiCamino:
no thats not what I was referring to. it was a single pixel wide line that flashed when moving. 

but i dont see what you are seeing :( :( :(

walter01

HiCamino:
I pushed a change that might fix what you see. it is soemthing I changed to try to fix my artifact but might cause yours.

HiCamino:
ok I still see random artifacts but I think its a lot better and more rare. can you please try the latest?

HiCamino:
thanks for testing and i dont think that crash is related to this change

humdingerb:
Can't say I see a change in performance. Maybe some more improvements can be found in the future...
I'll leave the PR open a few more days to give others the weekend to chime in.

BTW, is the commit message detailed enough to give future devs an idea when they have a look at the change down the road? No real programmer myself, so I can't judge that. :)

HiCamino:
> is the commit message detailed enough to give future devs an idea when they have a look at the change down the road? 

I dont know maybe i can add some comments to the code. next i have to rework selection so that might be scarier and then the rework of imageview for the checker bg might be scary too
so this one is not bad considering

overall artpaint code can be difficult in places!

PR #685, opened 2025-12-29
Refactor ImageView drawing to remove checker background from render

HiCamino:
- Set default for "bg" parameter in Image::Render to false so it doesn't draw the checker bg
- Remove code to render the checker bg in Image::DoRenderPreview
- In ImageView add checker bg bitmap. it is the size of two checker squares and is drawn as a tiled bitmap first thing in the ImageView::Draw function. the bitmap is drawn in the ImageView::MakeBackground() function and this is also called when the settings change to redraw the background bitmap
- in ImageView::Draw the background is drawn first and the brush drawing was also moved here. this is because in Haiku drawing should happen only in the Draw() function so other Draw commands were removed unless in functions called by Draw().
- ImageView::BlitImage() uses alpha drawing mode to draw the changed bitmap parts
- ImageView::UpdateImage() uses Invalidate() to force redraw but doesn't use BlitImage() anymore
- ImageView::MouseDown() invalidates window region when scrolling view to force redraw
- ImageView::MouseUp() code is not needed because DrawBrush happens in Draw()
- ImageView::MouseMoved() invalidates area where brush is drawn but doesn't draw brush because it happens in Draw() function. Also the Draw() when the mouse leaves the view is changed to Invalidate()
- ImageView::DrawBrush() clear code is not needed because Draw erases the bg before calling DrawBrush. Also calls to Draw are replaced by Invalidate()
- A few tools (FreeLineTool, EraserTool, HairyBrushTool) make sure updated_rect is within the image bounds to prevent crashes

NOTE: with this commit, Selection and Color Selector have artifacts. These will be fixed in separete commits
Color Selector fix is #686 
Selection tool fix is #687

please either test it all together, or if this is tested separate then ignore color selector or selection tool artifacts

### DIFF
--- a/artpaint/paintwindow/Image.cpp
+++ b/artpaint/paintwindow/Image.cpp
@@ -1490,28 +1490,11 @@ Image::DoRenderPreview(BRect area, int32 resolution)
 		int32 right = (int32)area.right;
 		int32 bottom = (int32)area.bottom;
 
-		int32 gridSize;
-		uint32 color1;
-		uint32 color2;
+		union color_conversion white_bg;
+		white_bg.word = 0xFFFFFFFF;
+		white_bg.bytes[3] = 0x00;
 
-		gridSize = 20;
-		rgb_color rgb1, rgb2;
-		rgb1.red = rgb1.green = rgb1.blue = 0xBB;
-		rgb2.red = rgb2.green = rgb2.blue = 0x99;
-		rgb1.alpha = rgb2.alpha = 0xFF;
-		color1 = RGBColorToBGRA(rgb1);
-		color2 = RGBColorToBGRA(rgb2);
-
-		if (SettingsServer* server = SettingsServer::Instance()) {
-			BMessage settings;
-			server->GetApplicationSettings(&settings);
-
-			gridSize = settings.GetInt32(skBgGridSize, gridSize);
-			color1 = settings.GetUInt32(skBgColor1, color1);
-			color2 = settings.GetUInt32(skBgColor2, color2);
-		}
-
-		BitmapUtilities::CheckerBitmap(rendered_image, color1, color2, gridSize, &area);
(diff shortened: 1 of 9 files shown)


HiCamino:
this was allot of work and a headache and if it all works as it should you should not see any difference at all :( 

but it makes the code better and more consistent with haiku and also removes the checker bg from the rendered image so #674 will work

humdingerb:
Appears to work nicely! Very well done! Leaving those PRs open for a bit to allow more feedback.

> and if it all works as it should you should not see any difference at all :(

There's one difference: The checker-bg doesn't appear fixed when zooming in anymore, as the checker-size is fixed, where it used to increase with the zoom level. At first I was a bit distracted, because the checker-bg appears to be moving in relation to the actual pixels of the image. This is only really noticable when using the mouse-wheel.
I suppose it can't be helped and we'll soon get used to it.

HiCamino:
ok i made the bg static like before

HiCamino:
i found an issue that needs to be fixed. the crop rectangle doesnt draw on top of the background anymore

humdingerb:
LGTM, after playing around a bit. Hope we've tested enough, the app is quite complex... :)

When rebasing #686 I had to resolve some conflicts. Maybe have a look before the rebasing starts. (Though I admit I'm not too good juggling branches...)

humdingerb:
Seeing as their commit message title is very similar, it'd be a good idea to sqash them all into one, if you don't mind. I suggest we merge this weekend, if there are no objections from anyone.

HiCamino:
ok you should only need to merge #687 bc it should have all of the changes and then we can close 685 and 686

humdingerb:
I actually just meant to squash #686 and #687 into one... but I guess that's now too late. :)

humdingerb:
Superceded by #687

Issue #701, opened 2026-09-14
Recent change in Bitmap.h causes compile error


haydentech:
The return type of BBitmap::BitsLength() recently changed (Haiku commit 9661eca), and that causes a compile error when that function is used in Layer::readLayerOldStyle()

This was my quick fix to get it compiling again:

```
	// align the file pointer to four-byte boundary.
	size_t alignment_offset;
	alignment_offset = (4 - (file.Position() % 4)) % 4;

	if (file.Read(bits, alignment_offset) != static_cast<ssize_t>(alignment_offset)) {
		delete layer;
		return NULL;
	}
	bits += alignment_offset;

	size_t remaining_length = layer->Bitmap()->BitsLength() - alignment_offset;
	if (file.Read(bits, remaining_length) != static_cast<ssize_t>(remaining_length)) {
		delete layer;
		return NULL;
	}
```

Haiku commit: https://github.com/haiku/haiku/commit/9661ecabc2b8b461ddb0bf38b0cfd07c7e08c4e7

humdingerb:
@waddlesplash : Is that kind of breakage expected?

I can confirm that this still builds on Haiku beta6. Wanna create a PR?

waddlesplash:
The breakage is expected, yes.

waddlesplash:
(It's an API but not an ABI break, so source compatibility is broken but binary compatibility isn't.)

humdingerb:
👍

korli:
When fixing this, please make sure it still builds on r1beta6.

PR #702, opened 2026-09-15
Fix background flicker during resize

haydentech:
- Also fixes image being drawn at top-left instead of centered
- Solves issues https://github.com/HaikuArchives/ArtPaint/issues/697


artpaint/paintwindow/ImageView.cpp
```diff
@@ -240,8 +240,9 @@ ImageView::Draw(BRect updateRect)
 	Window()->BeginViewTransaction();
 	SetDrawingMode(B_OP_COPY);
 
-	// draw checker background
-	DrawBitmapAsync(background, convertViewRectToBitmap(Bounds()), Bounds());
+	// draw checkered background behind image
+	BRect imageBounds = the_image->ReturnRenderedImage()->Bounds();
+	DrawBitmapAsync(background, imageBounds, convertBitmapRectToView(imageBounds));
```
HiCamino: do you need this part? i tried just adding the dolayout and it seemde to fix the issue at least on my side

artpaint/layers/Layer.cpp
```diff
@@ -578,17 +578,17 @@ Layer::readLayerOldStyle(BFile& file, ImageView* imageView, int32 new_id)
...
+	if (file.Read(bits, alignment_offset) != static_cast<ssize_t>(alignment_offset)) {
 		delete layer;
 		return NULL;
 	}
 	bits += alignment_offset;
 
-	if (file.Read(bits, layer->Bitmap()->BitsLength() - alignment_offset)
-		!= (layer->Bitmap()->BitsLength() - alignment_offset)) {
+	size_t remaining_length = layer->Bitmap()->BitsLength() - alignment_offset;
+	if (file.Read(bits, remaining_length) != static_cast<ssize_t>(remaining_length)) {
```
HiCamino: these changes in layer.cpp should be in a seperate pr for issue #701

artpaint/paintwindow/ImageView.cpp
haydentech: I'm seeing that as well.  I'll resubmit the PR.

artpaint/layers/Layer.cpp
haydentech: I agree, will resubmit this separately.

artpaint/paintwindow/ImageView.cpp
haydentech: The PR should be the correct, minimal fix now.

artpaint/layers/Layer.cpp
haydentech: Done -- this has been resubmitted separately.

HiCamino (approved):
looks good to me

only thing i wonder is does dolayout get called when the frame is resized so could this same code be removed from frameresized()? or change frameresized to call InvalidateLayout()?

